PR #17, opened 2015-01-27
Add a simple README.md

The6P4C:
The landing page for this project would benefit from a README. A simple one has been included in this commit.

This could be expanded to include the Road Map or extended description on the About web page.


jpd002:
Yup, I totally agree with that. I also need to add info about the build setup to make it easier for everyone to build the emulator on their machines.

PR #21, opened 2015-03-09
Get Torus by Lukasz running

MoochMcGee:
http://www.pouet.net/prod.php?which=52237

This doesn't get it fully working, but it gets it running. The video output is a bit jittery, and the torus is glitchy.


jpd002:
Thanks for the changes! I've merged them in the master, fixing a compilation error and adding reflection entries for proper disassembly :)

jpd002:
Torus is glitchy because of FPU rounding problems. 

f9b59cfe5942f0dbc2fe994b7e345053c168ff54 fixes the problem, but breaks TriAce games.

Issue #22, opened 2015-03-09
Budokai 1 crashes Play!


MoochMcGee:
It crashes in CMipsExecutor::PartitionFunction() at line 319.

MoochMcGee:
Apparently, the this pointer points to invalid memory.

jpd002:
Crash was caused by stack corruption when interrupt handler was called on IOP. 3ae11dc849a17a5f55bbade7ee11b19328b2f303 fixes the issue.

The game works now, but it's kinda screwed up probably due to timing issues and the emulator triggering vblanks too often.

![dbzbk1_01](https://cloud.githubusercontent.com/assets/6760433/7060861/e7522f26-de56-11e4-8da7-d944429bd008.jpg)
![dbzbk1_02](https://cloud.githubusercontent.com/assets/6760433/7060862/e8d5e61c-de56-11e4-8443-9678bdf65691.jpg)

MoochMcGee:
It doesn't crash anymore, but for me, I'm just getting a black screen.

jpd002:
Are you using SLUS20591?

MoochMcGee:
Yes.

PR #25, opened 2015-03-18
Fixes for Budokai 1 to boot further

MoochMcGee:
This implements more features of AddDmacHandler so that Budokai 1 boots further.


jpd002:
Thanks for the patch, I'm going to get the game and test the changes as soon as I can.

jpd002:
After further inspection, the patch might not work properly because calling AddDmacHandler with 'next = -1' with will overwrite any handler that was previously registered to be executed last. I think the proper behavior should be to link the handler at the end of handler chain.

Thanks nonetheless! As mentioned in #22 , the game now runs with a fix in the IOP BIOS implementation.

PR #49, opened 2015-05-19
Android multitouch support

SeannyM:
Adds multitouch support to the Android on-screen controls. Buttons are assigned pointer IDs so when the pointer ID assigned to a button is released, that button is released too.


jpd002:
Awesome work! Thanks!

PR #51, opened 2015-05-22
Android: Some changes to navigation bar functions

SeannyM:
Just some small changes to how the navigation buttons act in the Android UI:
-On devices with on-screen navigation buttons, the navigation buttons will be hidden during emulation allowing for the full screen to be used
-In the file browser, pressing the back button will bring the user to the parent directory instead of just closing the app


jpd002:
Immersive mode is definitively a must, it was on my TODO list for the Android port. 

I think your change to the back button's behavior is good, but I'm not quite sure how it should be handled to be as most "Android-friendly" as possible. Would it be better if we kept an history of directories and quit the application once we've walked back to the top of the history? The official documentation isn't really clear about the way the back button should behave in the case we have for the emulator's file browser. Nevertheless, I think your change improves the emulator's experience, so that's why I'm merging it in.

Thanks!

PR #56, opened 2015-06-01
Android: Add images for on-screen buttons

SeannyM:
I think it's about time we add some images to the on-screen controls. I'm not a master artist, but I think the controls I made look alright. If you'd like any adjustments, such as more transparency, let me know. In the future I'll probably add adjustable transparency, along with other button options such as adjustable scale.

Here's how they look in action:
buttons


zxcvbad:
Looks good and clean to me, will go well with future analogue sticks

jpd002:
That's awesome! It looks very sleek, I love it! 

I got some questions:
* Do you think you could add some visual feedback when the user is pressing the button? I think it's important for the user to see that the button he's pressed reacts even if the game doesn't react instantly. Maybe making the button brighter would do the job (there's a LightingColorFilter in the Android SDK that might work for that).
* Is there any source files to the images you've created (SVG or other)? If so, I think it would be nice to have them in the repository in case we'd want to make different resolutions later on.

SeannyM:
@jpd002 I used LightingColorFilter to darken the buttons when pressed, which IMO makes the most sense for pressed buttons. If you'd prefer light buttons when pressed, I can change it. About the images, I have .psd (Photoshop) files of the images, but they're the same resolution. If you want me to add them somewhere, just tell me where the best place to put them would be. I wasn't really thinking about resolution when I designed the buttons, oops. If we ever need higher resolution images, I could just recreate them.

jpd002:
Oh yeah, darkening the buttons does indeed make more sense. Works great on my side! 

If the source images have the same resolution as the ones you've added in your commit, then, I guess there's not much value adding them as PSD files. If you want to push PSDs to the repo later on, I think it would be best to add them in an "images" directory in the root. If we do that, it would probably make sense to move everything that's in "icons" in there too.

Thanks!

PR #59, opened 2015-06-14
Android: Add a setting to enable or disable FPS

SeannyM:
By default the option is disabled. I didn't think one option was enough to create an entire settings menu yet, so I just added a checkable item to the menu in the file browser, under About.


jpd002:
Thanks for implementing that feature. It will be in today's weekly build.

I'm working on a settings menu using Android's PreferenceActivity, so I will most likely migrate that setting over there once it is ready.

Issue #60, opened 2015-06-15
Pluggable VPU implementation 


bigianb:
Hi JP, I thought I would post this here rather than emailing. It's great to see the project gaining traction - I'm still very busy with real life, but watching with interest from the sidelines.
I've been tinkering with some very high level emulation for BGDA - basically I want to hook the VPU and GS implementations. So, when a game is loaded, I look to see what the ID is and if there are any specialised VPU / GS handlers available. If so, install those rather than the default ones.
This is pretty easy to do with the GS Handler but things are more tricky for the VPU. The EE currently binds the VIF and VPU constructors at compile time, so they cannot easily be hot swapped later.

I can of course hack something in my branch to do what I want but before doing that was wondering if you had a plan to do this already. I'm trying to keep my branch as close to yours to give myself an easy life when merging :)

MoochMcGee:
this is a horrible horrible idea and should be banned from this emulator

jpd002:
@bigianb While I wasn't planning to do that, I wouldn't have any objection to change the way the VPUs are constructed inside Ee::CSubsystem. Not quite sure about the pluggable part though. Do you want to enhance the game's visuals using the hooks? Are you looking for a way to get rid of that dreaded "DELAYED_MSCAL" thing?

About that, I was planning to write a test in ps2autotests to look into MSCAL's behavior when it's used in the same way BGDA (and other Snowblind games) uses it and hopefully find a proper fix for it.

Good to hear from you again btw :smile:

bigianb:
@jpd002 The delayed MSCAL works fine - in fact the emulation status of BGDA and the other snowblind engine games is now pretty spot on. What I'm looking to do is to use the core play engine to build an engine which gives an enhanced experience for the BGDA games. For stage one, I am aiming to get the display running in progressive scan at native resolution. Stage 2 would then be to allow modding, texture swapping etc (which involves hooking the EE CPU which I already do in my branch).
How to do this - well I know what all the VU calls do, so if I can intercept the MSCAL then I have the vertex information and if I intercept the GIF then I have all the texture information. I don't want to duplicate the VIF decoding, so I want to hook in at the VPU level.

Of course I am also hoping that stage 1 will give me a big speed boost as it will eliminate most of the per frame texture swizzling.
[...]

PR #61, opened 2015-06-17
Injectable vpu

bigianb:
See issue #60 
This PR makes the vpu objects heap allocated rather than stack allocated so that they can be injected at runtime rather than having their type fixed at compile time. This enables a user of the core library to provide specialised implementations (to enable enhanced features for specific games as an example).

I've tried to keep the changes in this PR as minimal as possible. I'm also open to other implementations that would provide similar configurability.

A 'better' or at least a more conventional solution would be to specify the m_vpux objects using a pure abstract interface rather than the heavyweight CVpu class. Currently any alternative implementation needs to extend CVpu - which is quite heavyweight. The downside with that approach of course is that all calls over the interface incur a vtable lookup.

Ian

### DIFF
--- a/Source/ee/Ee_SubSystem.h
+++ b/Source/ee/Ee_SubSystem.h
@@ -37,6 +37,9 @@ namespace Ee
 		void						SaveState(Framework::CZipArchiveWriter&);
 		void						LoadState(Framework::CZipArchiveReader&);
 
+		void						SetVpu0(CVpu*);
+		void						SetVpu1(CVpu*);
+
 		uint8*						m_ram = nullptr;
 		uint8*						m_bios = nullptr;
 		uint8*						m_spr = nullptr;
@@ -52,8 +55,8 @@ namespace Ee
 		CDMAC						m_dmac;
 		CGIF						m_gif;
 		CSIF						m_sif;
-		CVpu						m_vpu0;
-		CVpu						m_vpu1;
+		CVpu*						m_vpu0;
+		CVpu*						m_vpu1;
 		CINTC						m_intc;
 		CIPU						m_ipu;
 		CTimer						m_timer;
--- a/Source/ee/Ee_SubSystem.h
+++ b/Source/ee/Ee_SubSystem.h
@@ -37,8 +37,8 @@ namespace Ee
 		void						SaveState(Framework::CZipArchiveWriter&);
 		void						LoadState(Framework::CZipArchiveReader&);
 
-		void						SetVpu0(CVpu*);
-		void						SetVpu1(CVpu*);
+		void						SetVpu0(std::shared_ptr<CVpu>);
+		void						SetVpu1(std::shared_ptr<CVpu>);
 
 		uint8*						m_ram = nullptr;
 		uint8*						m_bios = nullptr;
@@ -55,8 +55,8 @@ namespace Ee
 		CDMAC						m_dmac;
 		CGIF						m_gif;
 		CSIF						m_sif;
-		CVpu*						m_vpu0;
-		CVpu*						m_vpu1;
+		std::shared_ptr<CVpu>		m_vpu0;
+		std::shared_ptr<CVpu>		m_vpu1;
 		CINTC						m_intc;
 		CIPU						m_ipu;
 		CTimer						m_timer;
--- a/Source/ee/Vpu.h
+++ b/Source/ee/Vpu.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "AlignedAlloc.h"
 #include "Types.h"
 #include "../MIPS.h"
 #include "../Profiler.h"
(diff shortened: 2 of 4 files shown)


MoochMcGee:
PLEASE BAN THIS PR FROM EVERYTHING EVER

Source/ee/Ee_SubSystem.h
```diff
@@ -52,8 +55,8 @@ namespace Ee
 		CDMAC						m_dmac;
 		CGIF						m_gif;
 		CSIF						m_sif;
-		CVpu						m_vpu0;
-		CVpu						m_vpu1;
+		CVpu*						m_vpu0;
```
jpd002: It would probably be better to use a shared_ptr or unique_ptr for m_vpu0 and m_vpu1. I'd rather avoid any explicit memory management because I think it's error prone. You can use a typedef to make the type name shorter, something like `typedef std::shared_ptr<CVpu*> VpuPtr`.

yuriks: Considering the previous lifetime of this object (tied to the parent as an inline member) `unique_ptr` would be the direct 1:1 semantic replacement.

bigianb:
Thanks for taking the time to think about this and comment.

I agree about the explicit memory management. I was trying to keep the change small and make it consistent with m_os which is a raw pointer.
@yuriks - I don't think it is a unique pointer because the object can be injected (that's the reason for the change) and there is no reason that the EE subsystem should own the lifetime of the injected object. Another wrinkle is that the vif holds a back-reference to the vpu object (the vpu passes *this). So to do this cleanly a shared_pointer which is injected, held by the vpu and held by the vif would IMHO be the cleanest way to do it. Using a unique_ptr would require passing the raw pointer to the vif still (via *this) and also passing in a raw pointer to the setVpux routines.
One thing to consider is the circular reference between the vif and the vpu - the vif should probably be holding a weak pointer to the vpu rather than a shared one.

yuriks:
@bigianb Mutually circular references like those usually indicate that having `shared_ptr` on both ends isn't actually the best solution: It can usually be determined that only one of the ends in this chain in fact owns the other, and that the lifetime is subsumed into the lifetime of its owner. In that case there's nothing wrong with using a raw pointer to hold a back-reference. As someone with admittedly only superficial knowledge about the PS2 architecture, it seems natural to me for the VPU to own it's corresponding VIF.

Likewise for injection of the VPUs, that would be done at start up, and the ownership would then be passed from the injecting code to the EE which will hold the owning pointers to the VPUs. The `Ee::CSubSystem` class currently owns all the other associated pieces of context for the emulator core. Unless a completely independent piece of code must use the VPUs past the emulator's shutdown (which I don't see any obvious use-case for) keeping the VPU there while emulation runs also looks fine.

I'm not very acquainted with the codebase, so sorry if I missed any relevant context here, just trying to leave my 2 cents regarding objects ownership and lifetimes.

bigianb:
> Mutually circular references like those usually indicate that having shared_ptr on both ends isn't actually the best solution
Agreed, the problem of course is that the object never gets destructed because it always has a lifetime - hence my comment that the vif could hold a weak reference rather than a (strong) shared pointer.

Semantically, the vif holding a weak reference and it holding a raw pointer as it does, is the same. The weak reference is sometimes preferable because the lifetime and ownership is clear - with a raw pointer there is always a danger someone may delete it in the destructor. weak_ptr is quite new in the c++ standard though and I'm not sure all the compilers that we use support it. Using a weak_ptr to break a cycle is a bit of a kludge though and it adds an overhead for every use which is unpleasant as this is a high bandwidth interface.

The existing vif code avoids the accidental destruction issue by holding a reference rather than a pointer - which is neat and also safe given that the vpu object owns the vif object. So on balance, I'd probably not change the vif implementation at this time.

bigianb:
I changed the code to use shared pointers. I thought about unique pointer as discussed above but using it does not add any extra safety, it just adds a restriction in that the IOC setter must move the object into the subsystem. I'm not adverse to using a unique_ptr here, I just don't see the benefit over a shared pointer.

Source/ee/Vpu.h
```diff
@@ -39,7 +40,17 @@ class CVpu
 							CVpu(unsigned int, const VPUINIT&, CGIF&, uint8*, uint8*);
 	virtual					~CVpu();
 
-	void					Execute(bool);
+	void* operator new(size_t allocSize)
+	{
+		return framework_aligned_alloc(allocSize, 0x10);
```
jpd002: Is aligned allocation necessary here? In Ee::CSubsystem, it's needed because the MIPS contexts that are stored there are referenced by jitted code (which uses aligned SIMD reads), while nothing from CVpu is referenced from it. 

I did a quick test on my side and aligned allocs didn't seem to be necessary.

jpd002:
I tested the branch on my side and it works good. I'll merge after we decide on the fate of the aligned allocations.

Source/ee/Vpu.h
bigianb: I added it because without it visual studio warns that the object on the heap will not be 16 byte aligned. There was no reason that I would see - I searched for __aligned and the relevant declspec but could not see any. There must be some object in the class structure though that is specify this alignment though or the compiler would not give the warning.

jpd002: I'm getting the same build log with and without aligned allocation. There are other places where the compiler says objects might not be 16-byte aligned, but CVpu doesn't look like one of them. I'll remove the aligned allocation on my side.

PR #92, opened 2015-07-11
Updated Manifest to remove space and updated banner.

LinkofHyrule89:
This is the Android Manifest from LoungeKatt I just removed a space from
it. Also I updated the banner I prefer my banner over his of course :P


Thunder07:
got to say, his is more round *day dreams* and smooth *dreams some more*, so I'd go with his but i definitely think "Play" should also be included.

LinkofHyrule89:
I think my biggest problem with his is that it has rounded corners and from what I've seen Android TV banners are suppose to be square. But it also does look weird since it doesn't contain any text like most banners do.

Thunder07:
ah fair enough, I agree then, I did a quick google search, most icons also dont have boarders.
I think you should horizontally rotate the "!" so that the shadow is facing the other way,remove the boarders, and apply a slight softening/smooth touch.

LinkofHyrule89:
Here is a screenshot of mine https://plus.google.com/+MatthewGarbett/posts/eMzEB9aFQco installed.

LoungeKatt:
If you want to remove a space from another pull request, it is best to sync the pull request and submit an additional commit or wait for it to be merged. By submitting a separate pull with the same content, either pull request will cause a conflict with the other and you remove the author from the original code.

It also helps to @ tag someone when critiquing their code to someone else, especially if you think they submitted an icon that was actually meant to be replaced.

The icon was the default icon extended to fit in the correct banner size. If you are still designing the icon, though, it may be good to smooth it out. A phone may be 1080p, but the Shield console is 4K. Rough looking graphics will look ten times as rough on a 52 in. screen.

LinkofHyrule89:
Thanks for the info I'm still learning how all this works sorry about that.

LoungeKatt:
Just make sure not to auto-format after removing the space. The typical Android IDE will add it back in because that is "common practice"

LinkofHyrule89:
That makes no sense to me it looks dumb with a space lol.

LoungeKatt:
It separates the categories from the actions.

LinkofHyrule89:
I suppose that's true if you had a lot it could be confusing. I'm trying to figure out this gui editor in Android Studio to see what I can't come up with for the project.

LoungeKatt:
Even if you have a lot, each category gets it's own intent-filter.

LinkofHyrule89:
LoungeKatt do you have Hangouts?

LoungeKatt:
I haven't touched a messaging app outside of texting since AOL was still popular.

LinkofHyrule89:
Oh ok well if you have a Google account you should message me on that maybe you can help me with this. https://www.google.com/+MatthewGarbett just press the little message on Hangouts bubble under my picture.

LoungeKatt:
I know how it works, I just don't use it.

I haven't used Android Studio. When they switched from Eclipse, I switched to Textpad.

PR #94, opened 2015-07-11
Android case detection for .elf files

Flav900:
Makes the .elf files detectable with any case by forcing the file name to lower case for check.


jpd002:
Thanks for the fix!

Flav900:
No problem!

PR #95, opened 2015-07-12
Ubuntu: Fixed Case-sensitive Build Error

Thunder07:
there are other things which need to be changed such as "$ANDROID_SDK_ROOT/tools/android.bat to $ANDROID_SDK_ROOT/tools/android" but that would break builds on windows.

however, with that said, I've successfully built "Play!" and installed it on my phone using Ubuntu.

### DIFF
--- a/Source/PS2VM.h
+++ b/Source/PS2VM.h
@@ -6,7 +6,7 @@
 #include "MIPS.h"
 #include "MailBox.h"
 #include "PadHandler.h"
-#include "iso9660/ISO9660.h"
+#include "ISO9660/ISO9660.h"
 #include "VirtualMachine.h"
 #include "ee/Ee_SubSystem.h"
 #include "iop/Iop_SubSystem.h"
--- a/Source/ee/MA_VU_Upper.cpp
+++ b/Source/ee/MA_VU_Upper.cpp
@@ -1,7 +1,7 @@
 #include "MA_VU.h"
 #include "../MIPS.h"
 #include "VUShared.h"
-#include "VIF.h"
+#include "Vif.h"
 
 using namespace std;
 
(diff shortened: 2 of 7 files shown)


jpd002:
Thanks for the fixes! I work in Windows and OSX so I never see those problems, so, sorry about that.

I still need to figure out what's the best way to build and develop the Android version, but if we decide to keep those scripts (build_debug.sh and build_release.sh), they need to be improved to support building on other OSes.

Thunder07:
They weren't a problem at all, I was actually surprised to see that you only missed the few. 

I was also thinking about this, but I was too lazy to implement it, since manually fixing it to work with Ubuntu was fast. 
If you don't get around to it yourself I'll get on it, but the concept is simple os identification through  the script with if statements. 

But what do you mean how to build the android version, using native code?

jpd002:
I meant, the development tools. I'm using cygwin and those shell scripts to do all of the Android development, but people might want to use other tools such as Eclipse or Android Studio.

Fadi5555:
@jpd002 you are really working hard for giving more improvement to your project.

the Best part is when come here some developers and people who have knowledge about that.
All that will help to see the progress faster.
It's nice to working together

Best of luck to you all.

Thunder07:
For that, by default android studio projects use gradle which can also be ran independently of android studio using CLI, for one that would give you the advantage of being able to script the build, but you can also use that project with android studio... 

And since I never used eclipse, i can't comment on that but Google recently dropped support of eclipse ADT plugin, so there is that... And you really can't please everyone.

jpd002:
Yeah, I think going towards Gradle/Android Studio would be the best bet, but I still need to try it and see what it's worth.

PR #98, opened 2015-07-13
Android: MultiOS Build Support

Thunder07:
I don’t run Windows or Mac so I couldn’t verify if this works, but I did download the SDKs/NDK so the paths are correct.
just need you to verify that Darwin for Mac & CYGWIN for windows work.

Edit: I've also sent pull requests to the other projects for the same fixes.

### DIFF
--- a/build_android/build_debug.sh
+++ b/build_android/build_debug.sh
@@ -1,5 +1,15 @@
 #!/bin/bash
 set -e
-$ANDROID_SDK_ROOT/tools/android.bat update project -p .
-$ANDROID_NDK_ROOT/ndk-build NDK_DEBUG=1
-$ANT_HOME/bin/ant debug
+
+case "`uname`" in
+	CYGWIN* | MINGW*)
+		$$ANDROID_SDK_ROOT/tools/android.bat update project -p .
+		$ANDROID_NDK_ROOT/ndk-build NDK_DEBUG=1
+		$ANT_HOME/bin/ant debug
+    ;;
+  Darwin* | Linux*)
+		$$ANDROID_SDK_ROOT/tools/android update project -p .
+		$ANDROID_NDK_ROOT/ndk-build NDK_DEBUG=1
+		$ANT_HOME/bin/ant debug
+    ;;
+esac
--- a/build_android/build_debug.sh
+++ b/build_android/build_debug.sh
@@ -3,12 +3,12 @@ set -e
 
 case "`uname`" in
 	CYGWIN* | MINGW*)
-		$$ANDROID_SDK_ROOT/tools/android.bat update project -p .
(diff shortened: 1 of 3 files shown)


LoungeKatt:
You could probably eliminate ANDROID_SDK_ROOT and ANDROID_NDK_ROOT since they should be part of PATH.

Thunder07:
Probably, would also need to append the build instructions to reflect that.

Thunder07:
I guess this is obsolete with the new arrangement.

LoungeKatt:
You could repurpose it for a master build script.

Thunder07:
you already have that covered.
unless you (@jpd002) still plan on keeping each build separate.

LoungeKatt:
I meant you could combine it into one script for the master project. The one I was messing around with somehow gets a ../ off when it builds everything at once, even though each project builds separately.

Thunder07:
I put this together https://github.com/Thunder07/Play-Build/tree/scripts

jpd002:
Nice! I think it the scripts should go inside a `build_android` folder though, because we might have scripts for other platforms at some point.

Thunder07:
Ya i was thinking about this.
I was thinking taking the script a step further, ./build.sh *OS* *Release_Type*
this will allow it to support multiple OS using one scipt and it can remain in the root directory of the project.

Thunder07:
just pushed this https://github.com/Thunder07/Play-Build/commit/38a818b9354ba659054bb6a31af2e2a2202e8f90
a single script to build debug, release & clean but as it stands it only supports android, not sure how you're getting the other platforms build, so I've used placeholders for now.

Fadi5555:
I have a question.
What is multiOS build? Is that a process for multi platforms?
Also is this thing will include an iOS port? Because it's named (multiOS)?

Thunder07:
no, this will just allow the project to be build on different OS(operating systems), such a windows, mac & linux.

Fadi5555:
@Thunder07 thanks for answer.
Well, that's a good idea. But the OS like win, Mac, Linux all these already have a ps2 emu (PCSX2) so in my opinion we should focus on mobile OS like Android / iOS. Because the didn't have a ps2 emu yet.

Anyway the creator @jpd002 doing his best to makes his project has more compatibility.
Also if there a good developers like you and @LoungeKatt who have more knowledge about these stuff.
The project will be more than playable and stable as well.
Working as a team will be useful and the reason for successes of any project.

Thunder07:
This is not what this is for, this allows you to build Play! for anything (android, ios, windows, macosx) but allows you to do it on windows, macosx and linux.

which devices or games the dev chooses to fix/support first is of his own accord.

Fadi5555:
Oh I got your point now
Thanks for explaining that for me.
Yeah absolutely this is a good idea indeed for fixing the issues on play!

PR #201, opened 2015-07-16
Added slight delay to IDEC command.

itssupermachotime:
Fixes issues with games that changes DMA3 parameters right after issuing a IDEC command.


itsdeltamachotime:
No, you falsely accused me and for that you are all fucked.

Just the fact that you admit to not respecting me or saying I don't deserve it means I will trouble more.

PR #212, opened 2015-07-16
Immaturity

itsomegamachotime:
No, you falsely accused me and for that you are all fucked.

Just the fact that you admit to not respecting me or saying I don't deserve it means I will trouble more.


itsomegamachotime:
@LoungeKatt Stop being immature. I don't care who started it.

PR #241, opened 2015-07-20
WIP: Cover Saving, Serial Searching, On Demand Description

Thunder07:
This is still work in Progress, The code is fully functional except for getSerial() which throws a 'random' serial to test with but the code need to be cleaned even more.... and some changes shouldn't even be here.(you can ignore  GSH_OpenGL.cpp, didn't see that there :p )

Progress so far:
1) Covers are saved (much faster loading on the next run and needs no connection)
2) using the serial no, compared against db, I can find name and id (thegamedb ID)
using that ID I can request the cover from thegamedb, yielding more accurate results. (when serial is not matched, thus ID is missing, we'll still attempt to get cover using name)
3) save game description too.(this is done the 1st time the image is loaded, or preview opened)**(Done)**
4) parse thegamedb to sql and cross match for serials.
5)Clean the code a bit more, there is alot of redundant functions, since I've used/recycled some of @LoungeKatt and also made my own, the re-adapted each to work together, I just pushed this to show progress **(Done)**

TODO:
1) waiting on the serial function from @jpd002 (the longer it takes the better since I need to concentrate on my study :/ )
2) add cover clearing option.
3) maybe allow user to choose own cover.
5) fuzzy search (games with no matching serial numbers, find closest name with levenshtein distance)

before I spend more time on this, @jpd002 do you think I should keep working in this direction? (aka would this be used eventually when its complete)
@LoungeKatt Sorry I butchered your code like this, I started working on reading the data from sql.db and worked backward to get images from your code.

finally, if you want to test these commit you'd need to download this [sqlite database](https://www.dropbox.com/s/o1mghnw5upqeoz3/skynet.db?dl=0) rename to thegame.db and pushed to /data/data/com.virtualapplications.play/databases/ then set the correct permission.
NOTE: there is no function to getserial of game, so I'm using a set list of serial to demo purposes. (aka the covers that you'll see are not those of your games).


LoungeKatt:
I'm glad you butchered the code if you intend to include the GameDB database in the application. I would rather not be held responsible for that.

Thunder07:
@LoungeKatt  I'm not planning on that... plus I'll tweet them or something...
one of the things that I forgot to mention is I do plan is parsing there db from the phone... which I haven't done yet which is why I linked to the db here, just for demo purposes.

LoungeKatt:
@Thunder07 I hope that works out, since it's @jpd002 that will get stuck in the situation if they consider it a licensing violation. I'm not trying to oppose the idea, just making sure it's done without risk.

As far as the code goes, though, it isn't a competition so there is no real need to justify changing something. If it is more efficient or more reliable, that's all that matters.

Thunder07:
@LoungeKatt I do get that, I recall apps such as Yaste being removed from the play store because google 'thought' Yaste's owner was in violation, without even asking him.

LoungeKatt:
BTW removing the AsyncTask is going to cause a StrictMode violation.

Thunder07:
@LoungeKatt how do you mean, for http requests?
well, all of that is already running within an async, so I only removed it from your class, but the functions are being called from another Async.

LoungeKatt:
As long as they are still inside an AsyncTask, it should be fine. They started making it a violation to run them on the main thread. There is a hacky way around it, though.

	StrictMode.ThreadPolicy policy = new StrictMode.ThreadPolicy.Builder().permitAll().build();
	StrictMode.setThreadPolicy(policy);

Thunder07:
@LoungeKatt I've mentioned it the list will be downloaded for offline use.
but I've also sent another msg just to be sure, emphasising the "I intend on Downloading the ps2 game list"

also for such thing, http or any heavy duty task, Its better be done using Async or a separate thread, to avoid the freezing the UI even if temporarily.

LoungeKatt:
That should be enough to CYA. You might want to add that screenshot to a folder in the project somewhere so they don't change their minds after they implement keys.

LinkofHyrule89:
That's good you got permission for that probably saves a lot of time. Great job on this I think it'll help a lot hopefully make loading the app a bit faster. I still think it wouldn't be too bad of an idea to have a loading screen in the form of an app logo.

Thunder07:
@LinkofHyrule89 you're right but functionality 1st, then aesthetics.
I'll probably be part of this, but no ETAs.

LinkofHyrule89:
Dang devs and their "no ETAs" rule :P jk Anyways yeah I understand there is a ton of behind the scenes stuff that needs to take place I've made a few posts in the hopes I can gather up one or two more developers to help out with the project.

Thunder07:
@LoungeKatt I see you getting the front & back, is there much need for this?
I'm considering dropping them since we're now saving the covers, so they take space...
and they're really a tiny icon... unless you plan on making it clickable/viewable as well?
````
						Element images = (Element) root.getElementsByTagName("Images").item(0);
						Element sleave = (Element) images.getElementsByTagName("boxart").item(0);
						//TODO: to keep or not to keep
						String cover = "http://thegamesdb.net/banners/" + getElementValue(sleave);
						BitmapDrawable game_icon = new BitmapDrawable(decodeBitmapIcon(cover));*/
						Element boxart = (Element) images.getElementsByTagName("boxart").item(1);
						String image = "http://thegamesdb.net/banners/" + getElementValue(boxart);
````

LoungeKatt:
The plan was to make it the description background and then write the description over it with a translucent filter. A lot of it came down to finishing by Sunday. I can submit the pull to your pull when it's all done.

Thunder07:
Great News, TheGamesDB are ok with us storing the database locally, not only that I also asked them if they'd like to get the serials merged into their DB and they do....
so once that's done, we wont need to cross match on the phone speeding up initial run.

screenshot_2015-07-21_13-30-16

LinkofHyrule89:
That's awesome! It's helpful to everyone I think 😊 Glad to hear they are supportive of the project.

Thunder07:
@jpd002 I'm happy enough with this pull, just let me know once you get the getserial function done, preferably it take File or a string path...
once that's done, I'll close this pull, and send a new one without GSH_OpenGL.cpp, Application.mk and project properties.

Ill work on the rest of the TODO on 2 weeks time.

LoungeKatt:
I have all the TODO done when this gets merged. ;)

Thunder07:
@LoungeKatt haha I can't stop you :)
just for future reference, GG.GameDetails.getBackCover(); of type getGameDetails should return back cover image, if not available it will return front cover.

I was also thinking of looking into this after my exams, but since you'd be more free for the time being you can have a look maybe https://github.com/moondroid/CoverFlow/

LinkofHyrule89:
That looks kind of like what they have on WiiFlow it would be a cool thing to have as an option.

LoungeKatt:
@Thunder07 I rewrote a lot of the functions to pull right from the database using a database helper class and used GameInfo for all the general functions (ie. saveImage, getImage, etc.). It's down to around 3 classes and still does everything.

Thunder07:
cool :)

LoungeKatt:
I'm not going to ask where you got the serial numbers and assume you made that database ;)

I took that one and modified it to include the fields from GamesDB. I'm merging in the data so you don't have to worry about that.

LinkofHyrule89:
I'm glad what I found earlier was able to help you guys it makes me feel useful even though I can't really provide actual code for the most part.

Thunder07:
I though it would be best not to include it within the app, to allow users to get updates (should there ever be any,but what are the chances)(also also with the whole just incase). but this way you do buy some sweet sweet time.

LinkofHyrule89:
Well once GameDB adds it to their database you won't need to include it any more right?

Thunder07:
@LinkofHyrule89 yes and no, depends... if they do include it, it would still need to be parsed into db, which takes few seconds. but including it saves time, since it can be pre-parsed and packaged.

LinkofHyrule89:
Very true my line of thinking is that eventually all the small tasks that are needed to display the games and their box art can take place while a logo is being displayed so it doesn't confuse the user that it's taking so long. But I imagine once everything is optimized and it's using the game id instead of names it'll be a lot faster especially after the first load since it'll just cache the information for later unless they add new games.

LoungeKatt:
I was looking at their updates and the timestamps and such, and including a base will save a lot of time, but using the db helper, you can update entries similar to automatic updates for software.

LinkofHyrule89:
So basically include the base and when it gets updated online it'll pull it?

LoungeKatt:
There are a few ways that could work. One is using their own lastupdate entry to compare and pull, another would be blindly updating the entries for the user's games, and the last would be having it push a new file once a month or so.

LinkofHyrule89:
It seems like on boot it should match everything it can and then update with the online database if there is one it can't identify maybe?

LoungeKatt:
That would clock in at just a little slower than getting it on demand (how it is before this pull)

LinkofHyrule89:
**Please wait as your games are cached...** Guess we'll see how the new system compares to the old one once it's pushed.

jpd002:
The whole thing looks promising, glad TheGamesDB granted permission to do store information locally.

GetDiskId function WIP: https://github.com/jpd002/Play-/tree/diskid. Still needs some work to expose the function to Java (and probably doesn't compile on Android yet).

LinkofHyrule89:
Good job jpd002 the progress being made build to build is phenomenal .

Thunder07:
@LoungeKatt since you'd be commiting the db after me, can you run through the db and remove the "." and replace "_" with "-" for the serials, since @jpd002 function returns serials like this SLPM-67010

LoungeKatt:
I just saw your other pull, but ironically I had changed the function on the native side instead. Rather than changing 4000 entries, changing one line of code made more sense.

Thunder07:
Hehe ya that works, dunno why I was considering fixing it from the java side...  Though its much easier in native side as you said :p

Thunder07:
@LoungeKatt if I push another local branch to this remote branch, will that work to reset this pull request? Or do I need to start a new one?

Thunder07:
this has been moved to https://github.com/jpd002/Play-/pull/254

PR #246, opened 2015-07-22
Fix:DisksUtils & Added to NativeInterop

Thunder07:
Not sure how you wanna handle these, but this should save you free minutes trying to figure out whats missing.

### DIFF
--- a/Source/DiskUtils.cpp
+++ b/Source/DiskUtils.cpp
@@ -1,4 +1,5 @@
 #include <boost/algorithm/string.hpp>
+#include "make_unique.h"
 #include "Utils.h"
 #include "stricmp.h"
 #include "DiskUtils.h"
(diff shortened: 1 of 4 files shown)


jpd002:
Well, I wanted to add a JNI function named `GetDiskId` that would call `TryGetDiskId` and throw an exception if it returns `false`. I don't know if that's a good pattern or not, I don't have much experience with Java.

Thunder07:
That works, I just wanted to make sure everything else is working fine. 

But either or really, it doesn't matter how you implement it on one side, we can easily adjust for it on the other side. 

Unfortunately in java you can't use reference values, I would have preferred the try function would return a boolean while the it's reference value returned the serial (haven't tried static variable, that might still hold true) but that would only make the code look nicer, I doubt it would more or less effective.

jpd002:
I've merged the diskid branch in master. I've also made the GetDiskId function return `null`. Thanks for the feedback!

PR #252, opened 2015-07-22
Game Information Database (see comments)

LoungeKatt:
This pull is meant to be rebased on https://github.com/jpd002/Play-/pull/241 and merged with https://github.com/jpd002/Play-/pull/246

This is being submitted purely for testing purposes and to track progress of the combined result.

The workflow:

Attempt to get serial from disk image

if serial is found:
  retrieve database information
  use game id to fill in any missing info
  store information to game id
  retrieve image from cache or web

if serial is not found:
  retrieve game id by search
  use game id to fill in any missing info
  store retrieved values by game id
  retrieve image from cache or web


Thunder07:
why did you Revert "Fix DiskUtils, Add function to return disk serial", your function had the advantage or returning a null when when serial not found, mine returns an empty string... I'll say it shouldn't matter once we verify an image is valid before adding it into display.... which also brings me back to using TryGetDiskId this function as boolean to confirm that.

LoungeKatt:
I reverted it to make sure the functions did the same thing, but either works honestly. The null check is implied the second the serial string is used as a database search parameter.

Thunder07:
haven't tried searching null, but attempting to search an empty string results in a %% search (wildcards) so it returns the 1st game in the database (devil my cry), not entirely sure about null, probably the same, but its probably better to do if (serial != null) over if (serial != "")

Thunder07:
not sure if you get a notification when I @ you in an edit, so ^^

LoungeKatt:
You still need to retrieve two copies of the same description because it isn't throwing all the files and all the details in piles. It searches each file individually. It should be noted that regions can have different content, editions, or other (albeit minor) details that will some day be important. Might as well get it over with.

The GameID column is empty because I have been working on the code first (using the handful of games that are completely filled out) and then going back while it compiles and adding more. There are still a few changes not included, but even if it were done, it's still waiting on the other pulls anyway.

Thunder07:
ok makes more sense, but since you said you're including the db, I imagined it'll be a full list.
but what missing detail? from the ones I checked, each request contains the description, the 2 sided cover, name and ID.

The crossed-match db does not match special editions to normal ones(e.g ffxii vs ffxii zodiac system), since it's doing an exact name comparison (ignoring case), since they don’t match in names, thus they'll be kept separate, so this isn't a concern....
if and its very unlikely, if different regions had different description, that wouldn't apply to any game in TheGamesDB list, since they're all mashed into 1 id anyway. (ffxii zodiac, i think, was only release in japan, this is an exception and its reflect in the name, almost like a special edition)

LoungeKatt:
There is also the possibility of labeling the files without it being directly from the API, and more commonly getting the regional cover. I had to add a cover entry because a few had revisions. It would be a quick shift of the hard-coded prefix into the database entries and then loading the whole url from the database to add images from other sources.

LoungeKatt:
The most recent commit is where maintaining a boolean would be useful. I don't know that it would be any faster than a null check, though.

LinkofHyrule89:
What about homebrew demo .elf files? Does the system just ignore those? I guess it's not really a big deal.

Thunder07:
@LinkofHyrule89 well, unless you know of a super secret db that has covers for homebrew, there isn't much we can do.

LoungeKatt:
This is better submitted after the pulls it extended are finalized.

PR #266, opened 2015-07-30
Android:Moved UI operation until after creating.

Thunder07:
Not sure, why, but every now and then this causes problems in my build.
Well, the problem is, the ui it's trying to access wasn't created yet, but every know and then it works.

Just so, you should probably give the UI time to startup before attempting to access it.

you can also, use Startup() which is called after OnCreate but before OnPostCreate, but since non of these methods are time-critical, this is just as good.


LinkofHyrule89:
Yeah it's probably a good idea to not try and use something that doesn't exist lol.

jpd002:
Humm, is `onPostCreate` the best place to do this? People are suggesting that `onCreateView` should be used instead: http://stackoverflow.com/questions/23653778/nullpointerexception-accessing-views-in-oncreate

Thunder07:
OnCreateView is not part of Activities but part of fragments. 

I was thinking if it's worth while adding them in, but as it stands you have one "page" in the app, which really doesn't justify them.

jpd002:
Oh, makes sense to use `onPostCreate` then.

jpd002:
Thanks!

PR #267, opened 2015-07-30
Android:Updated DB with the initial set of TheGamesDB IDs

Thunder07:


jpd002:
Working good! Thanks!

PR #270, opened 2015-07-31
Android: Update nav drawer, add appcompat

Thunder07:
I could describe it, but here are few screen shots.
2 thing though, there is no way to use appcompat_v7 support library as a jar, as jar libraries don't contain any resources, so I had to add all the resources along with it, the way I did,
I've also switched build to java 1.7 to allow diamond operations. needed for NavigationDrawerFragment.java, and why not?

[screenshot_2015-07-31-23-14-47](https://cloud.githubusercontent.com/assets/5013823/9018788/d537f536-37da-11e5-98f4-503ab632538b.png)
[screenshot_2015-07-31-23-14-54](https://cloud.githubusercontent.com/assets/5013823/9018787/d5366efa-37da-11e5-895f-99dc73427102.png)

with that said, android studio/gradle are slowly adapting native coding support, so once they do and we switch to grade, all them resources, libraries can be replaced with a single line "compile 'com.android.support:appcompat-v7:21.0.3'"

also, you may or may not need to call "android update lib-project --path appcompat" in build_android to setup the library.

Edit: I'm also using android studio for all the editing, so it give itself permission to fix some spaces.


LoungeKatt:
Any advantage to this other than looking really hideous... err... material?

LinkofHyrule89:
I think it'll look nice but if we're using that shade of blue might want to tweak the background color to match better.

LoungeKatt:
The other method was compatible all the way back to 2.1 as well. So the question is still if there is any advantage other than the appearance.

LinkofHyrule89:
I would say it has a lot of extra resources you can utilize and it's the way to stay up to date with what Google has for the newest versions of Android. AppCompat also includes animations and other stuff like that as well. I'd say in general it's mainly visual but why are you against it lol.

LoungeKatt:
Who said I was against it? I take it you never questioned anything your parents or teachers said because they are parents and teachers...

LinkofHyrule89:
No I question stuff all the time :P I guess it's good to know all the advantages and disadvantages and such. I'm sure there is more information in the documentation or something.

LoungeKatt:
That's why I asked. The documentation is mostly going about it the previous way and this appears to be based on a third-party extension (https://github.com/ikimuhendis/LDrawer).

LinkofHyrule89:
Oh I don't don't know anything about that I thought he was just using the regular implementation maybe it's just easier to use that. It looks like it's pretty much the normal stuff just already packaged.

LoungeKatt:
If I followed all of the documentation correctly, this version of the drawer more or less wraps the previous version of the drawer. It seems like you can still manipulate the drawer the same way, but have to go through the wrapper to do it. Usually the third-party versions extend the drawer to add features like customizing the animation style or supporting non-standard views.

LinkofHyrule89:
Yeah there isn't a reason to do that if it doesn't give extra features I would think.

LinkofHyrule89:
I made a new gradient I'm not sure if you have to do something to it to make it the proper format or anything. http://i.imgur.com/P67aOvm.png

LinkofHyrule89:
I was able to pull it and compile a bit broken since it's not finished but I added my background too it looks awesome. http://imgur.com/a/OjCSD

Thunder07:
@LoungeKatt rotation was disabled so it made no different, but i've added it any how for future reference.
onOptionsItemSelected is not needed, the navigation drawer is already closing by itself when you select an option.

As for your 209 comment, https://github.com/Thunder07/Play-/commit/03adee1fd7193692db5142d94cb54af046c62028#diff-f11373a886bca74336c18e3b08ec9058R165

LoungeKatt:
Setting the orientation was a hack for the number of columns being hardcoded. The configuration override is where it should actually be getting adjusted.

The drawer starts open and overlaps the first row of games. Is that intentional?

Thunder07:
yes, so that you know its there, but once you open/close it it will remember that.

LoungeKatt:
It was hard to read the entire change with all the whitespace changes so I don't know how much I am missing. It doesn't seem like overlapping the main view would be "desired" functionality.

Thunder07:
if you're getting anything other than what's in the screenshot, then no it's not, so can you post one?

LoungeKatt:
https://github.com/Thunder07/Play-/blob/03adee1fd7193692db5142d94cb54af046c62028/build_android/res/layout/main.xml#L27 you have it below the other view, but fitting the system window. That is most likely the cause.

screenshot_2015-08-01-00-57-45

https://github.com/Thunder07/Play-/blob/03adee1fd7193692db5142d94cb54af046c62028/build_android/res/layout/main.xml#L21

The xmlns tags should also be removed. Normally an IDE will warn about them being nested.

Thunder07:
@LoungeKatt I'm not getting that, but you probably have a custom dpi ya?

LoungeKatt:
Play! is set to stock so it is seen how it was meant to be seen. Regardless, I only use modified DPIs that are "standard" (ie. 320 -> 160, 640 -> 320) so that wouldn't offset the view.

Thunder07:
@LoungeKatt Play! might be set to stock, but not your status bar.
I'll double check anyway.

LoungeKatt:
The status bar shouldn't cause the pseudo actionbar to suddenly overlap the main view. If the view isn't flexible, that is a fault in the design of the view. You can't really expect to slap a disclaimer on the app that "your device must use a stock DPI" because that wasn't covered in the template.

Thunder07:
tell that to swiftkeyboard.

and it can be fixed, fitsSystemWindows moves the entire app, all the way to above the status bar, thus allowing you to colour it, then I need to move all the way back, I do so by setting the padding according to the statusbar standard dp... but since yours is custom, it breaks loose a bit.

to fix it, i need to measure the statusbar height and adjust accordingly.

LoungeKatt:
If you are hardcoding a padding, that would be the issue. 

public int getStatusBarHeight() { 
      int result = 0;
      int resourceId = getResources().getIdentifier("status_bar_height", "dimen", "android");
      if (resourceId > 0) {
          result = getResources().getDimensionPixelSize(resourceId);
      } 
      return result;
}

Thunder07:
@LoungeKatt you're running kitkat ya?

LoungeKatt:
On one of them.

Thunder07:
then do you get the overlap on both?

LoungeKatt:
Can't turn the Shield on right now. Nvidia says it'll explode.

Either way, http://stackoverflow.com/questions/22192291/how-to-change-the-status-bar-color-in-android is suggesting the same thing when using appcompat. It may also be good to look into their statusbar color. Ironically, they are talking about the same blue theme.

Thunder07:
I already have a method for changing colour, there are few colours included but this is set to be default since its blue, if you wanna test the rest just edit manifest's theme with something from style.xml

as for the issue, can you try this, obviously i can't tell if it works for kitkat/ your device https://gist.github.com/Thunder07/3c4dfa75da581217054a

Thunder07:
@jpd002 if you've missed it, what do you think?

@LoungeKatt  just seen this "I'm not sure if this was a WIP, but did you intentionally add back stuff like the exit menu item?", no its not and what's wrong with having exit option?

LoungeKatt:
Toolbar toolbar = getSupportToolbar();
    toolbar.bringToFront();

    private Toolbar getSupportToolbar() {
    		Toolbar toolbar = (Toolbar) findViewById(R.id.my_awesome_toolbar);
    		ViewGroup.MarginLayoutParams mlp = (ViewGroup.MarginLayoutParams) toolbar.getLayoutParams();
    		mlp.bottomMargin = -statusBarHeight;
    		toolbar.setLayoutParams(mlp);
    		setSupportActionBar(toolbar);
    		return toolbar;
    	}

    @Override
    public void onBackPressed() {
    		if (NavigationDrawerFragment.mDrawerLayout != null && mNavigationDrawerFragment.isDrawerOpen()) {
			    NavigationDrawerFragment.mDrawerLayout.closeDrawer(NavigationDrawerFragment.mFragmentContainerView);
    		} else {
    			super.onBackPressed();
    			finish();
    			return;
    		}
    	}

    public boolean onKeyDown(int keyCode, KeyEvent event) {
    		if (keyCode == KeyEvent.KEYCODE_MENU) {
    			if (mNavigationDrawerFragment.isDrawerOpen()) {
				    mNavigationDrawerFragment.mDrawerLayout.closeDrawer(NavigationDrawerFragment.mFragmentContainerView);
    			} else {
				    mNavigationDrawerFragment.mDrawerLayout.openDrawer(NavigationDrawerFragment.mFragmentContainerView);
    			}
    			return true;
    		}
    		return super.onKeyDown(keyCode, event);
    	}

Just seems like you make a lot of extra work for yourself to do things a certain way, such as having back control the menu.

Thunder07:
Well.... The onbackpress does what it says on the tin..... 
Onkeydown... well, that's was part of the gist and not the pull, and it's referring to the menu button, not the back. 

And I still like to know what's wrong with the exit button.

LoungeKatt:
It's extra work because you have back open and close the menu then look for a double tap to close the app, which in turn needs an exit button because it's horribly counter intuitive to exit.

I suggest trying other apps with the same navigation style. Gmail, YouTube, etc. The above is how they accomplish it without being as complicated.

Thunder07:
well, at this stage is merely personal preference, just because it's done one way doesn't mean it can't be done another.
as for the backbutton, it severs 2 functions at once at the same time, isOpen =! isOpen while checking for the double tap, I'm sure you got that already, but no harm in making sure.

if you prefer it another way, you can submit a pull, should this be merged and @jpd002 can decide which one he'd rather have it.

anyways, you never said, if the last commit fixed the actionbar overlaying the covers.
That probably also applies to the settings menu in your case as well ya? (at which case haven't been fixed)

LoungeKatt:
Personal preference would be a valid point discussing a logo or a view layout. Navigation and hardware keys are expected to follow a standard workflow across the entire platform. Luckily this is Android and Google won't reject the app for it like Apple.

This may be your first jump into app development, so elaborate is to be expected. As I said, it will be easier to let the feedback make the point. It stings a little more that way, but it's effective.

The settings menu worked fine across the board. The main menu still looks as it did before, but as you said, I can submit a pull for that later.

Thunder07:
Technically this still does but with slight addition.

"stings", because I care about difference in personal preference?, I'm not saying I don't care and I will take things as such into consideration in the future, should the feedback favours one way or another.

and that's not exactly the context I used that in, if there is a mistake in my code, I'm willing to look into it and fix it, albeit, in this case, I don't have that issue and it will be easier for you to fix then I.
but I referred to it in the context of difference of opinion.

LoungeKatt:
Stings because while I say it is "non-standard", the typical user will say "broken" or "sucks"

"When the previously viewed screen is also the hierarchical parent of the current screen, pressing the Back key will have the same result as pressing an Up button -- this is a common occurrence. However, unlike the Up button, which ensures the user remains within your app, the Back key can return the user to the Home screen, or even to a different application."

http://ux.stackexchange.com/a/16198

It seems like this is being taken as a personal attack and honestly isn't worth the effort. I have nothing more to say on the matter. Do with that information what you please.

Thunder07:
No, you're being insistent, as such so am I. 
The reference you used, assume multiple hierarchy which we don't have, even at which I have a solution for, where it won't show the menu or attempts to exit until you're back in the main  screen.

Thunder07:
if you create a commit of the changes, ill add it to this pull.

Thunder07:
@LoungeKatt just picked the changed for the drawerlayout fix and created a commit from it.

LinkofHyrule89:
Change the line in project.properties to android.library.reference.1=../../appcompat and store appcompat in the root projects folder with the other repos and boost. My setup is Repositories-->appcompat,boost-,codegen,dependencies,framework, and play. This allows you to not have to include appcompat on the repo. However that being said @jpd002 says he's probably going to upgrade the project to gradle so after that it'll just pull it from the Android SDK like @Thunder07 talked about in the OP.

Thunder07:
@LinkofHyrule89 sound id, I've rebased and open a new pull https://github.com/jpd002/Play-/pull/274

PR #274, opened 2015-08-02
Android: Update nav drawer, add appcompat

Thunder07:
A rebase of pull https://github.com/jpd002/Play-/pull/270
having also allowed the white spaces, as not be commited into this, for easier viewing,
I've also taken @LinkofHyrule89 advise and moved appcompat folder out, as to not be included as well.

you can find the appcompat in $android_sdk/extras/android/support/v7/appcompat/ (if it's not there, you'd need to run the sdk manager and install support library, just as well, i've uploaded it here for ease of access www.madnation.net/dl/appcompat.zip)
i've tried to use absolute path but that doesn't work.
any how, after copying appcompat to root directory, you'd need to run "android update lib-project --path appcompat" from root dictory of Play-Build, then build as normal

List of changes:
1) update to appcompat v3
2) overall material design
3) transparent nav drawer
4) transparent navbar
5) programmically adjusted padding (for navbar and status bar height, includes Main & Settings Page)
6) programmically created gradient background (this will become important later when theme selection is added)
7) moved UI operation to onPostCreate, to avoid future issues

accidentally pushed out purple color theme, fixed in last commit.
if you have a color preference you can either change AndroidManifest.xml to select one of the predefined styles (in style.xml) or you can add one there to you liking there as well, I'm sure you can follow.


jpd002:
Thanks for the changes. Sorry I didn't reply to this sooner, but I will take a look at this PR soon.

LinkofHyrule89:
I have more changes to merge in that I'll be posting soon.

LoungeKatt:
https://github.com/Thunder07/Play-/commit/2493f9c752772e8a9b525267f0e3fd022c032fa6 <- 90% unnecessary.

LinkofHyrule89:
I posted my changes here 96a1f5e6879e6de07f94bc412ecb5795cc794f54 I guess he can probably just add them to his pull request or I can wait to get them merged.

LoungeKatt:
@LinkofHyrule89 Did you base that on and test from this repo?

LinkofHyrule89:
It's based on Thunder07 repo nav3 branch so it should merge directly with this pull request.

LinkofHyrule89:
I think we should use the purple as the default theme to try and keep a consistent branding. I really love the blue and bright purple and think they should both be included as optional themes. Here is a super quick mockup of my idea.
screenshot_2015-08-02-01-46-40

Thunder07:
@LoungeKatt
Xml doesn't change with theme, nor can it take user preference into account. 
The rest is to fix the padding issues caused by using translucent nav bar, in landscape mode the nav bar overlays the right cover, in settings again the right side is under the nav bar. 
In portrait, it fitsystemtowindow allows content to go behind the nav bar, as such padding to the bottom is needed, so scrolling to the bottom will have that extra padding to push the cover above the nav bar. 

Now, which part is useless?

LoungeKatt:
The entire section where you draw a gradient, setting fitssystemwindow to false instead of simply removing the line, and a bunch of other extravagant code that is more for line count than function.

Thunder07:
@LinkofHyrule89 purple is Apptheme, and can be enabled as such, (when it was purple, the background looked blue to me, when I made the theme blue, the background looked blue to me :p).

jpd002:
Ok, I've tried this PR and here's my thoughts:

- I really like the global look and feel of this, but I think we should keep the behavior of this as close as possible to existing Android applications that have a drawer. I think the Google Play Store application is a good reference for this.
- Pressing the back button when the drawer closed should close the application. When the drawer is opened, pressing back should close the drawer.
- I think we should not have an "Exit" button in the drawer.
- I'm not sure about the transparency, while it does look nice, but it might make things hard to read. We can leave it in for now, but we might want to revisit that later.

Thunder07:
@jpd002 commited as you liked.
for point on, is it refering to anything specific or point 2 and 3
as for point 4, i know, it kind of breaks out of material design even if by a bit, you can adjust transparency at https://github.com/jpd002/Play-/pull/274/files#diff-c4257fa5734cb0df77279a55f969c72aR92 remove the replace to go back to full solid or adjust to try different transparencies.

LinkofHyrule89:
Maybe we can make the transparency optional?

Thunder07:
@LinkofHyrule89 I can have that on my todo list.

jpd002:
I think it should be transparent or not transparent, I don't think an option to change that would be interesting.

LinkofHyrule89:
We could make it not transparent on the default dark purple theme and transparent on the other themes. I personally like the transparency but we should definitely explore our options.

Thunder07:
I just want to point out something that could 'potentially' become an issue.
the way, the nav bar is setup with 2 listviews, when used in landscape on a small screen the 2 list could overlap, collapsing one of them into a scroll mode, while not a huge deal, I thought I'd point it out.
but i'd imagine anything that's capable of running the emulator shouldn't really even be less than 720p, thus its shouldn't be an issue.
just so, I don't have a small screen to test with to see if that's the case.

LoungeKatt:
@Thunder07 I know a lot of things you don't, but the point is making the app work with your PR

Thunder07:
but it does, and I'll appropriate reasoning behind it.

LoungeKatt:
@Thunder07 It's a suggestion. If you previously omitted the "," before, you shouldn't need to justify why it is there instead of simply fixing it. Appropriate whatever you like, but don't be a PITA.

Thunder07:
@LoungeKatt, I really don't mind suggestions as long as they serve a purpose, as such I asked for your reasoning, and considering how anything could be a learning experience, but without reasoning it really won’t help me(or anyone following this) not make the same error again.
Now since this has been committed and it doesn't effect compilation nor performance, I see no need for it.
and you'd have to forgive me if I was/am PITA, as I don't know what I've done for that(today anyway).

LoungeKatt:
"else you know something I don't?" is typically sarcasm and hard to interpret as a real question.

The reasoning would be so that you could always return, since finish only fires onStop, onDestroy, etc and does not actually halt the app immediately.

Thunder07:
granted my use of 'else' in that question didn't help.
but I don't get your point, where does onStop/onDestroy come into this?
since finish is not gonna be called because of the return statement.

"you could always return" and isn't that what I'm doing, using return to avoid finish?

LoungeKatt:
I can't really understand what you posted, but onDestroy() is called by the finish() command.

I have no idea what statement you are talking about and if it was removed, it seems a little confrontational to drag it back into the conversation because you saw it in an email.

Thunder07:
I merely brought it up to explain my reasoning as you're trying to explain yours.
i assume you removed it because you didn't want to get into an argument, and since we're trying to be civilised, i thought no harm in bring it back in. (to justify my choice, and if you had an opinion on the topic)

	@Override
	public void onBackPressed() {
		if (mNavigationDrawerFragment.mDrawerLayout != null && mNavigationDrawerFragment.isDrawerOpen()) {
			mNavigationDrawerFragment.mDrawerLayout.closeDrawer(NavigationDrawerFragment.mFragmentContainerView);
			return;
		}
		super.onBackPressed();
		finish();
	}

we're talking about this no?
you wanted me to change it to 

	@Override
	public void onBackPressed() {
		if (mNavigationDrawerFragment.mDrawerLayout != null && mNavigationDrawerFragment.isDrawerOpen()) {
			mNavigationDrawerFragment.mDrawerLayout.closeDrawer(NavigationDrawerFragment.mFragmentContainerView);
		} else {
			super.onBackPressed();
			finish();
		}
	}

now, since there is no actually statement after the end of else{}, there isn't much point in enclosing it if I'm already ending the earlier if with a return.

LoungeKatt:
I honestly don't even remember the point of this conversation and since my attempt to avoid a pointless argument was cited and subsequently ignored, I don't intend to even try to remember.

LinkofHyrule89:
So hows about those PS2 games?

LoungeKatt:
@LinkofHyrule89 How's Tetris :facepunch:

LinkofHyrule89:
Oh you know it's never ending.

Source/ui_android/java/com/ikimuhendis/ldrawer/ActionBarDrawerToggle.java
```diff
@@ -0,0 +1,182 @@
...
+import android.support.v4.widget.DrawerLayout;
+import android.support.v7.app.ActionBar;
+import android.support.v7.app.ActionBarActivity;
+import android.util.Log;
+import android.view.MenuItem;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.ImageView;
+
+import com.virtualapplications.play.R;
```
jpd002: Humm, from what I understand, Ldrawer is an external library. Do we need to import stuff from the emulator in there?

LinkofHyrule89:
@jpd002 that just makes the cool menu button animation.

Source/ui_android/java/com/ikimuhendis/ldrawer/ActionBarDrawerToggle.java
Thunder07: no, as libraries are, you'd import from them and not to them.
however the particular line you reference gives us access to Resources from the main app, in this case few lines lower 'R.drawable.ic_drawer'
but even then you mention the emulator, I’m assuming you mean the main ui of the app and not the emulator activity?

Thunder07:
@LinkofHyrule89 he's asking the reason for importing stuff from the main app into a library, not it's function, you can see the explanation above.

LinkofHyrule89:
Yeah I realized that after you answered the question lol.

jpd002:
Sorry I couldn't continue my review of this last night. Here's some other questions and comments:
- If the background is now rendered using code is it necessary to keep that file? https://github.com/Thunder07/Play-/blob/nav3/build_android/res/drawable/gradient_grid.9.png
- I had at look at the original LDrawer project (https://github.com/ikimuhendis/LDrawer) and saw it was meant to be used as a library which would explain why you have those `import com.virtualapplications.play.R` in there. I guess you decided not to use it as a library because the project files are using Gradle, but would it make sense to change the way it's used once the emulator's project files are updated to use Gradle (something I'm currently investigating)?

Thunder07:
Once you swap to gradle, that folder with the library classes can be substituted with "compile 'com.ikimuhendis:ldrawer:0.1'". 
No that resource is not needed, I kept it as a placeholder, as it only makes sense (efficiency wise) to keep the coded method if you'd like to have a theme/colouring system later, which I plan on adding, assuming you wouldn't mind it.

jpd002:
Well, I think theming is an interesting idea and it's ok if we want to prepare for the inclusion of that feature. So, I think we should remove that gradient file for now.

LoungeKatt:
You could technically still load it as a library. I think @LinkofHyrule89 mentioned it somewhere on here. You would add ```android.library.reference.1=/path/to/appcompat``` in project.properties and then all you need inside the project is the NavigationDrawerFragment.java

Thunder07:
@jpd002 sure, if you're done reviewing and would like to merge now/soon I'll get on it now, else I'll probably do it later. 

@LoungeKatt the library is not provided directory as jar, as such I'd need to compile it to one 1st adding and additional useless step. (unless ant build can do that by itself)plus the whole reason for moving appcompat out was it's there was too much "useless" commits.

jpd002:
Yeah, just trying to avoid adding external files to the repository here. I also don't like the idea of having to add LDrawer specific resources into the emulator's resources folder, but I guess that can't be avoided for now.

@Thunder07 Yeah, I'm pretty much done reviewing. I don't want to put too much pressure on you, so merging can wait a bit if you need some time.

Thunder07:
@jpd002 not at all, I've waken up 1:30 ago and can't fall asleep again, so I'm contemplating getting out of bed, that's the only reason  I ask :p

jpd002:
@Thunder07 Acutally, I've just tested on Nvidia Shield TV and it seems the items on the navbar are not accessible using the gamepad/remote. Is there any way we could fix that?

Thunder07:
@jpd002 navbar or nav drawer... as are they not selectable or you cant open it.

I have my ps3 control connected, so i cant open it, but once i do with touch i can navigate an select, so im asking.

if tis a matter of opening, then i can probably map the menu button (i assume the shield has 1) to open the nav drawer.

jpd002:
@Thunder07 Well, the actual button on the navbar is accessible and the drawer can be opened, but you can't move the cursor to select "About" or "Settings".

Thunder07:
@jpd002 ummm... odd enough, i can access them just fine :/
ill have a look online to see if anything of sorts was mentioned.

LoungeKatt:
Adding
     android:clickable="true"
    android:focusable="true"
to the buttons should fix it and doesn't hurt if it isn't needed

Thunder07:
@LoungeKatt its a listview, so im sure its taking care of that... but maybe if i ass it to the listview itself?

LoungeKatt:
I forgot it was a listview. Give me a second, there was another workaround for that.

LoungeKatt:
android:choiceMode="none" -> android:choiceMode="single"
and add
android:focusable="true"

Thunder07:
@LoungeKatt i check it out... but it seems that the cursor is jumping from the top listview to the scroll view with the covers....

though im working on replacing that with gridview (which is why i originally said the items work), as it seems gridview is not stealing focus

LoungeKatt:
Originally it was a gridlayout, but that doesn't support automatic stretching. That's why it was converted to a gridview, but that requires an adapter, which is why it ended up a table.

the actual line should be android:choiceMode="singleChoice"

LoungeKatt:
The table is stealing focus because of the line <requestFocus />
All you really had to do was remove that line. Probably a little late, though.

Thunder07:
that only steals focus as its inflated.
anywho non of the suggestions worked.
I also rather like adapter, and as it stands their isn't any otherway to do that. (unless we stay with this)

@jpd002 ummm, ya from the looks of it, you can either merge sacrific them two options until i get gridview fully working or wait it out a bit longer, i can probably work on it and get it Saturday morning/evening (my time, so evening night your time)

LoungeKatt:
If you want I can restore the version I had in the next hour or so.

Thunder07:
thats jpd002 call.
but i do have something in the work https://github.com/Thunder07/Play-/pull/1 but that would be very rough compared(littery hacked to pieces just for testing) to yours if you had it done before

LoungeKatt:
I'm not sure it is the same thing, but that's up to the judges.

Thunder07:
Well as it stands it's a mere display case, no click function or even a proper size adjustment, so much I marked it poc rather then wip :p

LoungeKatt:
I used a custom gridview that supports the scrollview wrapper. It should be clickable and such. Testing a build now.

LinkofHyrule89:
Android TV usual should have it's own layout using the SDK for the side menu similar to what dolphin and other apps have done however making it work with the current ui is a step in the right direction.

jpd002:
Well, I'd rather not sacrifice those two options, so I'll wait for a fix. But if @LoungeKatt's test is successful, we can go with that and merge this PR.

LinkofHyrule89:
Everything in this pr works really good on mobile devices so definitely should be merged 😊

LoungeKatt:
Took an extra build because childView is not the same as childview.

Thunder07:
incase of emergency use this patch :p https://gist.github.com/Thunder07/6bcabef52f974d36c1d2
it'll move the settings/about into the top listview, just as a tmp fix.

LoungeKatt:
Holy merge fail, Batman!

PR #275, opened 2015-08-02
Graphic updates - 9patch to resource / cover retrieval

LoungeKatt:
This would also allow for themes to alter the entire background


Thunder07:
since now, we've have the covers saved and you'll allow rotation, why not just allow android to handle all of onConfigurationChanged, including redraw on rotation, since everything should be loaded locally, there should be little to no lag.

LoungeKatt:
http://stackoverflow.com/questions/5123407/losing-data-when-rotate-screen

Android likes to handle rotation with a pseudo restart, but this way the app at least gets a say in what is cleared and what is reloaded.

https://github.com/AbandonedCart/Play-/commit/423bc54789b3bd8f484dd57c7ce0f3d903d52d29
 will then bypass the "full" refresh by retaining the file list.

jpd002:
All good on my side. Is it ready to be merged?

LoungeKatt:
Why yes it is. Sorry for the delay. Messing with streaming PS4 haha

jpd002:
Haha, no problem! Thanks!

PR #276, opened 2015-08-02
Integrate (updated) iCade library

LoungeKatt:
Changed to just integrating iCade.


jpd002:
Lowering the deployment target is no problem for me, but I don't really feel like downgrading the project... As it is now, it's impossible to open it in Xcode 6.0?

LoungeKatt:
Xcode 6 only supports up to iOS 8. I was copying out of an existing project for "correct" values, so I can increase that again from inside Xcode. Without any change, it displays an error and won't open.

jpd002:
Well, I think it's better to keep the project as "up-to-date" as possible for now. Sorry, I don't really have a good reason for that, it's just a personal preference. If Apple decides to break something in a new Xcode version, then, I might consider sticking to a previous version.

LoungeKatt:
That's fine. Apple likes to force people to update, but I can't upgrade to Yosemite. I think the theory is that if they make all their stuff only work on the new one, every other software developer will update. Yeah, uh huh, sure...

PR #278, opened 2015-08-03
Various UI changes including Play Blue, and Dark Purple Themes.

LinkofHyrule89:
Added play blue theme, dark purple theme, update logos (made them bigger and more aligned on xhdpi and xxhdpi), update placeholder text to use strings, use logos instead of title and subtext, change titles to be "- Games" via the string but in the future it should change automatically to be a header so the user knows where they are at such as Recent, Everything, and Favorites. Removed pointless subtext and title for navigation drawer.

Screenshots here https://plus.google.com/+MatthewGarbett/posts/YwNkCHysVF5


LinkofHyrule89:
I'm just going to redo this after the other PR is merged.

PR #281, opened 2015-08-05
Convert game grid to GridView with scroll support

LoungeKatt:


Thunder07:
ummmm.... though as i mentioned my gridview is incomplete, i didn't need to use any hacks for hight stuff.

and surely you should have based this on my branch to avoid the whole rebase again.

LoungeKatt:
Extending an API isn't a hack. It makes the view scrollable but prevents it being cut off. It is a pretty big stretch to compare a working version with a partial test when complaining about custom views.

I can rebase if it's going to be merged but when have you said no to extra work?

Thunder07:
well, the view works perfectly in my partial test, what has been hacked is mostly your code... aka no click long click folder selection, which ive done to get the gridview to display stuff, and i hard coded cover size just because i went to go to bed at that stage... as such i can say with certainty it works, just not ready for prime time.

LoungeKatt:
So what is your problem exactly? Is it because there is a custom view? I can easily do away with that.

I think the only way you'll be happy is if you get to do it, so if you want until Saturday just to have your name on it, I can take this down and we can make it another last minute change.

Thunder07:
custom views aren't a problem, your using the grid view(which is a scrollview) inside another scroll view.

and im just trying to understand why your getting this cut off, since i didn't get it. (one of us is making a mistake, either you or my mistake fixed that problem.. which i'd like to think is unlikely)

LoungeKatt:
I could understand if you had some specific aim to save time, eliminate extra work, or improve the functionality. Nobody is making a mistake, I just happened to come across certain scenarios you may never have or never will face when a GridView goes rogue. It is more commonly seen when putting the GridView with another flexible view (ie. custom actionbar).

If you can go to bed knowing there is no custom view and the scrollview is no longer there, then...

PR #282, opened 2015-08-05
#281 rebased on #274

LoungeKatt:


Thunder07:
"Convert partial library into actual library project"
we solved that by keeping project directory, why bring it back in?

LoungeKatt:
It was not "brought back in" simply put into the project folder. Had you looked at the commit before commenting, you may have noticed the source you left in the main project was also moved into the library folder. @jpd002 said gradle was a consideration. At which time the move is one directory level upward. Hardly anything you can't handle.
 
If you had looked at the change (again), you would see that it was removing the code to darken the theme background. I can see how the color is retrieved, but I also see that rather than picking the correct shade to begin with, it is hacked together inside the activity.

Let me know when you have useful feedback, or it at least reflects more than the commit description.

LoungeKatt:
appcompat has been moved up to https://github.com/jpd002/Play-Build/pull/1

It was scheduled for possible deletion either way, so this should prevent any further trivial issues over the exact location of the folder.

LoungeKatt:
@jpd002 that's everything. I may not be around later but it should be split up enough to only take certain parts if needed.

The sort stuff can easily be discarded if you only want the grid. Homebrew only hasn't been tested since I don't have any. It doesn't do anything if none is found, though. Recent works by last modified and updates the game when it's launched.

jpd002:
@LoungeKatt Thanks for the PR. If we're going to commit appcompat, I think it should be committed in the Dependencies repository. Play-Build should only have submodules.

jpd002:
Could this also be missing a `res/values/dimens.xml` file? I'm trying to build this branch and it's saying that it can't find values like `tool_bar_top_padding`.

Thunder07:
"You are welcome to comment on the upstream source, but there is no need to behave like a stalker by going through a fork to leave your illegible comments." 

@LoungeKatt, LOL, you do know the comment I made was about the branch that you made this pull request from right..? 
But I'll repeat myself "upstream",  since you removed it.

"Remove unnecessary step, Prevent color recalibration"
The recalibration, that you thought would cause confusion is directly related to theme colour as such it can't and should not be a pre-set resource or hex value.

LoungeKatt:
@jpd002 It is all in the appcompat folder now. Until that switches to gradle, it requires the specific version we have, which is what was submitted in the PR to -Build, but would be in -Dependancies

@Thunder07 You made the comment on the fork, not on the PR. It is all the better it was removed because at least this time you make sense.

It is baffling that you aren't familiar with the code you submitted, so let me explain how it works. 
@style/Blue references a style. That's simple enough. 
That style has items, such as "colorPrimaryDark" which is set as "#1976d2"
"#1976d2" is a HEX value, which you are saying it should not be. It is. You added it.
This HEX value is then being altered to be darker by the line that was changed.
Rather than using code to make it darker, "#1976d2" should be updated.

Unless you are pulling colors out of a hat somewhere and the fact that the HEX values in the blue style are what is made darker is some magical coincidence, the basis for the theme colors is a HEX value.

Thunder07:
another lol... you're fully permitted to comment on [my branch](https://github.com/jpd002/Play-/pull/274#commitcomment-12491348), but im not on yours? (the difference, you made the comment on the line)(fyi, that comment would appear in the PR anyway)

and thats you assuming I want them (colorPrimaryDark and the background) to be the same shade, but that shade is too dark, as such I would have needed to make it brighter still.
finally, before commenting on how i do/don't know my work... you should really check you've got it (atleast that part of the code correct)....
and this is how to disable that, if done correctly.

		float[] hsv = new float[3];
		Color.colorToHSV(attributeResourceId, hsv);
		hsv[2] *= 1.0f;// make it darker
		colors[0] = Color.HSVToColor(hsv);

should be

		colors[0] = attributeResourceId;

LoungeKatt:
@Thunder07 Had you commented on the line, I would understand. The point is that you are in such a hurry, you don't even take the time to hit back and type out something that is, but if used then, that makes but has no sense. I would rather the comment be an hour late than sound like a drunk text.

You have 
```colors[0] = Color.parseColor("#" + Integer.toHexString(attributeResourceId)```
in a code comment, yet now "done correctly" it is
```colors[0] = attributeResourceId;```
So which one will make "that part of the code correct" or do I have to guess?

If you had a valid explanation for why the original code is right, post that first instead of droning on about how my fix was wrong. I dismiss you because your comments sound like bitching. Regardless, I do not intend to discuss any further changes with you, nor will I be responding to any more of your comments.

@jpd002 For the sake of your sanity and mine, this will be my last submission to the Android side.

Thunder07:
i couldnt be arsed to check the validity of the code for your sake.

LoungeKatt:
@Thunder07 I removed them as a courtesy. While I have no respect for you as a person, I do feel it is only fair. As you stated, the line comments appeared in your fork. Now they don't.

This reply was also a courtesy. Loose ends and such. The color change has been reverted, so there is no reason you should need to spend any more time on this PR. Off you go.

LoungeKatt:
@jpd002 If you decide down the road that you would like to expand the theme capability to support two different colors (but still support a single color in progressively darker shades) the changes are:
https://github.com/AbandonedCart/Play-/commit/9d63d7ed1d67907e2efa3b4b5a82840a90baba51
https://github.com/AbandonedCart/Play-Dependencies/commit/34473785dd6a1531133940f24c197d27098de5bc

jpd002:
Ok, I got this to work. There are some minor issues that will need to be fixed, but I think this is good enough for now. Merging this.

@LoungeKatt Thanks for the tip. I'm not a huge fan of modifying external libraries, but it could be handy later on.

LoungeKatt:
@jpd002  Fair enough. Glad it didn't self-destruct on merge, haha.

PR #290, opened 2015-08-10
Android: Seperate Activity For FilePickers/External Emulator Launcher

Thunder07:
This just moves the External Emulator Launcher to a separate activity, it mostly functions to keep the code uncluttered.
but it also does offer the possibility to customizing the luncher name to external.
however, a small notice, if the main app was launched and then closed (using the home button), then a game is launched through an external launcher, after that game is exited, the main up would show up.
however, I expect people using an external launcher wouldn't be a frequent users for the main ui.

### DIFF
--- a/Source/ui_android/java/com/virtualapplications/play/ExternalEmulatorLauncher.java
+++ b/Source/ui_android/java/com/virtualapplications/play/ExternalEmulatorLauncher.java
@@ -1,0 +1,70 @@
+package com.virtualapplications.play;
+
+import android.app.Activity;
+import android.app.AlertDialog;
+import android.content.DialogInterface;
+import android.content.Intent;
+import android.os.Bundle;
+import android.os.Environment;
+import android.util.Log;
+import android.view.ContextThemeWrapper;
+
+import java.io.File;
+
+public class ExternalEmulatorLauncher extends Activity {
+
+    public ExternalEmulatorLauncher() {
+    }
+
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        // TODO: This method is called when the BroadcastReceiver is receiving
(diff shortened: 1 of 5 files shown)


Thunder07:
hi, @jpd002 I know you initial said you'd be interested in seperating the External ISO launcher, but i haven't rebased this since i pushed it, and since none of us payed any attention to it, it got tracked back.

if you'd like me to rebase it, let me know, if you dont want it, you're free to close it.

jpd002:
I had a look at this a while back, but I wasn't too sure of the approach. 

I like the idea of adding a new activity that handles the launching of a file, but it's a bit of a shame that there's some duplicated code (especially in `onCreate` and in `launchDisk`). I thought it would be better to move the virtual machine's initialization logic and the function that attempts to load a disc image in a separate class that would be used by both `MainActivity` and `ExternalEmulatorLauncher`.

I'm still planning to do that change at some point, but it's low priority for me ATM. If you wish do that refactor yourself and update your PR, I'll be glad to merge it.

Thunder07:
I'd say I can to do that, I'll have a look later tonight.

Thunder07:
I've updated the commit, hopefully this is closer to what you're looking for.

jpd002:
Yup, that's more like it. 

Why don't you let the exceptions coming from `launchDisk` bubble up and be caught by its caller instead of returning a string? Also, right now, no message will be shown if there's a error launching the disc image when you go through the MainActivity.

Thunder07:
you're right, if I let the method (`launchDisk`) throw the exception, the IDE would have warned me about the exception in `MainActivity.launchGame` and prevented me from build this, without handling it there too.

I'll revise it, once i get the chance.

jpd002:
Thanks for this and sorry for the delay!

Thunder07:
Not at all :)

PR #291, opened 2015-08-10
Android: associate missing serials with a database entry

Thunder07:
while this solution isn't ideal, as it doesnt give the user the freedom to reset any mismatched title.
that isn't a huge issue at the moment, since they can't do so anyway, irrespective of this fix.

to test this, checkout a game such as Marvel vs Capcom 2.
the noticed difference is that its cover takes longer to load each launch, since it has no database entry associated with its serial, it make an online request for the cover everytime.


jpd002:
Thanks for the improvement!

PR #292, opened 2015-08-12
Update headers with plays logo and section header.

LinkofHyrule89:
Rebased with newest master and only included the commits to change the header to include the play logo and header text. Like explained previously the header text should eventually be programmatically updated to show how the games are being sorted such as unsorted, favorites, homebrew, and recents.


LinkofHyrule89:
Accidentally added an ant file to this branch lets try again.

PR #294, opened 2015-08-12
Update icons, banner, and temp box art to match default blue theme.

LinkofHyrule89:
Updated the resource files for the icons, banner, and temp box art to match the new default blue theme that was in the last merge from Thunder07. Thanks to LoungeKatt for helping with this earlier.


Thunder07:
@LinkofHyrule89 since you're updating the cover art, can you also fix the width... It seems to be slightly narrower then PS2 covers.

LinkofHyrule89:
I'll probably try to redo it do you know what the size of the box arts are?

Thunder07:
Probably no definitive size, but looking at the downloaded covers from gamedb, they seem to have a a ratio of 10:7 (height vs width) (so a 245 width would be 350 height, in this case you can also shorten the cover to get the same effect).

LoungeKatt:
PC, 360, PS2, DVD, etc.
3240 x 2175 (170 Spine)

http://vgboxart.com/forums/showthread.php?21025-Cover-Dimensions (first Google search result for "ps2 box art dimensions")

One side: 3240 - 170 / 2 or 1535

Scaled ~ 240 x 340

Thunder07:
@LoungeKatt are the covers showing up small on the tablet?

LoungeKatt:
They show up average size on the Android TV. I haven't tried a tablet.

LinkofHyrule89:
Do you think 250 would be ok for the width?

LoungeKatt:
I think the original (242) only left about 5-10 pixels on either side, so that may be cutting it close.

PR #297, opened 2015-08-13
Gridview

Thunder07:
These PR is meant to fix the gridview lag issue, caused by constant attempt of reading cover from storage.
The reason for that is adapters try to recycle views once they go off screen, as such with the current method, each time a view is recycled/created, the storage is polled for the cover image.

I've attempted to fix this by pre-processing all the data the gridview needs before passing them to it.
thus everything needed for display is always available to the adapter from the start.

to be able to pre-process the data, I needed to make sure every method (and async) always returned a value instead of it(the method/async) setting view when the data has been retrieved.
as such this explains the 1st 3 commits.

 Android: Added GameInfoStruct and used to replace appropriate array:
this was done in an attempt to have a unified structure which contained all the needed data, including cover display bitmap.

Android: Seperate GameAdapter class, changed Async class to normal class:
this commit is a bit less organised, due to the nature of how connected this functions are.
but it's mainly to separate GameAdapter into its own class, which just makes it easier to call.
and changed async classes to just function holder class, I removed async so that I can call those method from an async function (since an async function can't call another async)
Note:You've notice while i removed Async, I've kept the constructor for the class, this was mainly done, to avoid major changes to the structure...
e.g before```new GamesDbAPI(mContext, gameID, serial).execute(game).get()```
Now ```new GamesDbAPI(mContext, gameID, serial).get(name)```

Android: Fixed Cover view padding build up
Android: stretch cover, to make them all uniform.
some covers vary slightly in height an wide and standing side by side, and over each other, you can see that, this will make sure to stretch all of them to fit the view, making them all similar in height and width... note though they're getting stretched, its mere pixels (~10 to 20 maybe, in either width of height), since they all have roughly the same ratio, as such the effect of stretch(lose of quality) would not be noticeable.
Android:Fix passing dataID which is a null in this statement
just fixes, the 1st 2 already exist in master, while the 3rd was brought upon by the changes in this PR

Android: optimize imports:
While I was at it, I ran the project through Android Studio optimize imports option to remove unused imports.

Note:
I've also come up with this solution here.
https://github.com/Thunder07/Play-/commit/6a7bc5b914733ed156b6ad716ae0489ab5eed676
but I though this PR is the better solution. since this single commit would still permit for lag, the 1st time the cover is loaded, since it wasnt cached yet.
(there is an explanation to the actual changes in the comment section of that commit... since I've moved a function around, it's hard to see what changed within it)


LoungeKatt:
https://github.com/nostra13/Android-Universal-Image-Loader

PR #298, opened 2015-08-14
Add an LruCache to resolve scrolling redraw

LoungeKatt:
Alternative to #297 that overcomes a few issues created while solving scroll lag:

Does not require an excess Singleton class (Uses the existing database / cache)
Does not delay the initial load until the image is available (Initialize the view, fill as loaded)
Supports using Async task (Parallel loading and thread safe)

This has not been fully tested against a large list of games for this emulator (since I don't have a lot of games), but it has been tested against a large collection of images for other lists and grids.

### DIFF
--- a/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
+++ b/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
@@ -499,7 +499,7 @@ public class MainActivity extends ActionBarActivity implements NavigationDrawerF
 		
 		protected void onPreExecute() {
 			progDialog = ProgressDialog.show(MainActivity.this,
-				getString(R.string.search_games),
+				getString(isConfigured ? R.string.updating_db : R.string.search_games),
 				getString(R.string.search_games_msg), true);
 		}
 		
--- a/Source/ui_android/java/com/virtualapplications/play/database/GameInfo.java
+++ b/Source/ui_android/java/com/virtualapplications/play/database/GameInfo.java
@@ -20,6 +20,7 @@ import android.graphics.BitmapFactory;
 import android.os.AsyncTask;
 import android.os.Build;
 import android.os.StrictMode;
+import android.support.v4.util.LruCache;
 import android.util.Log;
 import android.util.SparseArray;
 import android.view.View;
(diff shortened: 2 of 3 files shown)


LoungeKatt:
Based on the discussion at https://github.com/nostra13/Android-Universal-Image-Loader/issues/675

Thunder07:
this is probably a better solution than mine.
but I imagine the lag would still exist, atleast during the 1st scroll of each run?

you should use [android.util.LruCache](http://developer.android.com/reference/android/util/LruCache.html) instead of [android.support.v4.util.LruCache](http://developer.android.com/reference/android/support/v4/util/LruCache.html)

LoungeKatt:
Whether or not there is lag in the initial scroll depends on how soon you start scrolling. The images are cached to memory as soon as they load. It is much like the minimal delay you experience when opening Twitter and immediately scrolling up for more items. The lag during the first scroll would still be proportionally less than the delay for loading everything into memory before allowing a scroll and would provide the same result.

Only images are redrawn as a result of Android's subpar garbage collection. The view is not reloaded on scroll, so creating a second array to store the values would only be beneficial if they were then being used in a completely different view later. There isn't really a middle ground when the entire purpose of this is to provide an alternative that will avoid needing to gather all of the data into memory first.

Thunder07:
"Only images are redrawn" true, but I think you're miss-understanding the process (or I am)... it is redrawn since the view was recycled and everything in it was reset with different values(including the images) as such we are using a different view everytime(well, its one of the 4, but it doesnt contain the data we want, as such we request them again and again and again). "The view is not reloaded on scroll", I guess I just explained that too.

"initial scroll depends on how soon you start scrolling", again, one of us has thing wrong, but I added breakpoints to see how it works and this is my experience...
only the onscreen views are created.... when you scroll, the minute the very edge of the next view is onscreen, the getview method is called, thus the createitemview is called... the childview that is passed already has cover, description and click set(not always, only when another view has gone off screen), but we reset them with different data from getGameInfo everytime... which brings the image issue... the issue isn't entirely redrawing the images, but actually having to access them from storage to be loaded into the image view... so your solution works, since they're cached.. so they're the only thing that doesnt need to be re-read from another source.
my point is "so creating a second array to store the values would only be beneficial if they were then being used in a completely different view later", this is what is happening.

LoungeKatt:
TL; DR it's justification for wasting resources.

PR #299, opened 2015-08-15
Gridview fix: padding and cover size uniformity 

Thunder07:
I though these are better in their own PR since they're small fixes and easy to see the effect.

they'll probably conflict with the other PR later, but when and should this get merge and when and should that PR get approved it can be fixed.

Android: Fixed Cover view padding build up:
due to the reusing of the same view by the adapter, any padded cover could potentially keep getting padded and also will not lose it's padding when used in non padding position (if that same cover was set anything by last position, it would be set with the padded values), as such this fixes that, at that it will always reset the padding on any cover and always padd the original padding values as to prevent buildup.
Android: stretch cover, to make them all uniform.
simple enough, some covers are narrower, taller (albeit by few pixels by height/width) than others, as such you can see that when they're stacked side by side or on top of each other.
this is cause all of them to stretch to fit their views, giving them all the same width vs height.

### DIFF
--- a/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
+++ b/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
@@ -594,7 +594,7 @@ public class MainActivity extends ActionBarActivity implements NavigationDrawerF
 			} else {
 				ImageView preview = (ImageView) childview.findViewById(R.id.game_icon);
 				preview.setImageResource(R.drawable.boxart);
-				preview.setScaleType(ImageView.ScaleType.CENTER_INSIDE);
+				preview.setScaleType(ImageView.ScaleType.FIT_XY);
 				((TextView) childview.findViewById(R.id.game_text)).setVisibility(View.VISIBLE);
 			}
 			
--- a/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
+++ b/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
@@ -642,7 +642,8 @@ public class MainActivity extends ActionBarActivity implements NavigationDrawerF
 		private final int layoutid;
 		private final int padding;
 		private List<File> games;
-		
+		private int original_bottom_pad;
+
 		public GamesAdapter(Context context, int ResourceId, List<File> images, int padding) {
 			super(context, ResourceId, images);
 			this.games = images;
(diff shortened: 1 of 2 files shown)


LoungeKatt:
I think you also need
```
if (original_bottom_pad == 0){
    original_bottom_pad = v.getPaddingBottom();
}
```
from https://github.com/Thunder07/Play-/commit/29851dc8c75cd9da6869924c3e7811e51ef9c708#diff-c372cf262e74ea218f6a1574f9d460daR138

Thunder07:
hehe thanks for the catch.

LoungeKatt:
And there is one more FIT_CENTER at https://github.com/Thunder07/Play-/blob/5971e9e433305fad69bb5526146ffc2e3f84145b/Source/ui_android/java/com/virtualapplications/play/database/GameInfo.java#L181

Thunder07:
fixed thanks

jpd002:
Thanks for the submission!

For 1db530ad97705b05fcd09c15b250bbd87862d5ad, I was thinking of getting rid of the status bar and navigation bar transparency in the themes (see 060b0ab8861876ec1a9bde4644848331257d8076). I think this is better for two reasons:
- Currently, the status bar overlaps the tool bar, which doesn't look great.
- Adding padding to the list view and navigation drawer requires lots of code (and you're adding more in that commit), only for the possibility of seeing disk covers scroll under the navigation bar.

So, unless there's a good reason for having a transparent status bar and navigation bar, I won't merge that commit.

The other commit looks good, but it could have been in a different PR since this is a different bugfix. I might end up cherry picking that commit, if that's ok with you (still gotta test it though).

LoungeKatt:
The expandable grid may not be so bad now that I can say "you've seen the issue I was trying to explain"

Thunder07:
@jpd002, it's really your call. 

personally I have no navigation bar, as such I've added that on request.

But I wanna say this if the issue is to do with code using alot of code, you can fix that by adding the padding to the xml using standard values, the reason I've gone for the coded method is to avoid it breaking for those with custom sizes. 
I don't have the overlapping issue, again, could that be to nonstandard size? 

Again I've no real preference when it comes to these options anyway, I believe you can still colour the status bar on lollipop without any of this anyway, kitkat would lose out, but a small price for all the coding it required.

PR #300, opened 2015-08-15
Android: Indexing

Thunder07:
Android: added GameInfoStruct unified structural class for data handling
Android: Replace Array with GameInfoStruct
I've added GameInfoStruct, since its much easier to handle data across multiple methods if we're using a unified structure which holds everything... also few commits down, I've also linked GameInfoStruct to updating the IndexDB.

Android: two classes in charge of setting up the IndexDB
I used SQLiteOpenHelper without ContenProvider, because according to google documentation, it's only needed if I need to expose this data to other application, which at the time of writing I don't see the point of.. also having SQLiteOpenHelper always me to predefine few methods such as getAllIndexGameInfoStruct() etc, purely making them more convenient to call.

Android: update IndexDB when game is started:
instead of editing the files modified date, I'm now using the database.

Android: moved createListItem() into Adapter, pass array position for update:
as the name suggests, createListItem() was moved so it will have access to games variable inside the adapter, so it can update it with any changes.

Android:Clean up
self explanatory, but I'm removing all the redundant functions, since they're now no longer needed.

Overall effect:
No longer need file picker, since the entire phone is indexed on 1st run (only valid images and their associated folders).
on the 2nd run, it will scan any folder which was previously detected with a  valid image.
Data from games.db will be copied into the IndexDB as required with associated image downloaded, this will come in handy later, when I add an option which allows user to set their own cover and description, since I don't want the user to have access to games.db (as this will be wiped every time we update the database).
This file scan method seems to be a bit faster as well, I've added timers into the code and this method could scan the entire content of my phone in 1 second (1000~ millisec) vs 4 seconds(results would obviously vary based on storage size, folder numbers etc), it should be due to implementation, but could be due to the checks, such as skipping ./Android ./Data, either-way it comes in handy.

TODO later: (Done)
that will be in its own pull some time in the next few days, as this pull is getting large enough.
-- allow user to modify Title, Description and Cover Image. (Done)
-- Clear Index List Option (Done)
-- Verify The Image still exist on attempted launch, since this code now relying on index for the display list, the user could have moved the game, maybe disconnected the storage, thus we should prompt him to that. (Done)

Edit:
should have mentioned this, though it was implied in my explanation, since we're indexing the devices, games from different folders will all be able to appear without having to clear folder.

### DIFF
--- a/Source/ui_android/java/com/virtualapplications/play/GameInfoStruct.java
+++ b/Source/ui_android/java/com/virtualapplications/play/GameInfoStruct.java
@@ -1,0 +1,86 @@
+package com.virtualapplications.play;
+
+
+import android.graphics.Bitmap;
+
+import java.io.File;
+
+public class GameInfoStruct {
+    private String m_TitleName;
+    private String m_ID;
+    //private final String m_Serial;
+    private String m_FrontLink;
+    private Bitmap m_Thumb;
+    private Bitmap m_Thumb2;
+    private String m_Overview;
+    private File m_File;
+
+    public GameInfoStruct(String m_id, String m_titlename, String overview, String m_frontLink) {
+        m_TitleName = m_titlename;
+        m_ID = m_id;
+        m_Overview = overview;
+        m_FrontLink = m_frontLink;
+    }
+
+    public GameInfoStruct(File file) {
+        m_File = file;
+        m_TitleName = file.getName();
+
+    }
+
+    public String getTitleName(){
+        return m_TitleName;
+    }
+    public String getID(){
+        return m_ID;
+    }
+    /*public String getSerial(){
+        return m_Serial;
+    }*/
+
+    public String getFrontLink() {
+        return m_FrontLink;
+    }
+
+    public void setFrontCover(Bitmap thumb) {
+        m_Thumb = thumb;
+    }
+    public void setBackCover(Bitmap thumb) { m_Thumb2 = thumb; }
+
+    public Bitmap getFrontCover() {
+        return m_Thumb;
+    }
+
+    public Bitmap getBackCover() {
+        if (m_Thumb2 ==  null) {
+            if (m_Thumb != null) {
+                return m_Thumb;
+            } else {
+                return null;
+            }
+
+        } else {
+            return m_Thumb2;
+        }
+    }
+
+    public void setFile(File file) {
+        m_File = file;
+    }
+
+    public File getFile() {
+        return m_File;
+    }
+
+    public void setDescription(String overview) {
+        m_Overview = overview;
+    }
+
+    public String getDescription() {
+        if (m_Overview == null) {
+            return "No info";
+        } else {
+            return m_Overview;
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


LoungeKatt:
It seems like a lot of this is fixing things broken by the commits before it (ie. [Android: check for elf and don't do an online check for them](https://github.com/Thunder07/Play-/commit/6ec109149b518633053bc32573067f84cfbdf60c) which restores [Android:Fix ELF using other games cover](https://github.com/jpd002/Play-/commit/8f9f281be6a2a470363586c55200e5da9d35d2dc)), so is there more to it than removing the provider and adding a file index (read: database table)? 

I am ignoring the new attempt to implement GameInfoStruct (see #297), since it isn't really relevant to this, albeit a better reason than the last one.

Thunder07:
not exactly, "Android:Fix ELF using other games cover" was caused by a null gameInfo.getGameInfo(game, childview); irrespective of ELF or not, but it manifest to me through elf, since by deault, we shouldnt be getting a correct match to the online database.

Android: check for elf and don't do an online check for them:
will stop ELF from making it that far, but as I said the earlier one is more than just for ELF, but mere misfortune naming due to how it manifested.

and GameInfoStruct, does worked better here, since it linked in to everything else very nicely.


"so is there more to it than removing the provider and adding a file index (read: database table)"
I didn't remove the provider, that is there still serving the same purpose.
I've kept the database seperate, since its alot easier to define methods instead of having to relay on contentprovider, it will also be easier to backup, instead of having to import the entire database... or have to read and write it to get just that table... and that also applies to restoring it.
does it need to do more?

LoungeKatt:
At 17 commits, I figured I was missing something. GameInfoStruct is more appropriate here, but still not entirely necessary. I'm not your addiction counselor, though, so if it makes you feel better it is less useless than it has been so far. That's something.

Thunder07:
I could collapse them all into one if that would help you feel better.

LoungeKatt:
I don't really care. It is probably easier this way. The next PR may not even be able to use a Singleton.

Thunder07:
but we're not, not yet anyway...

but saying this faster the fact.... the main reason was to have the extra methods I've added... but looking back at it, and it's still do-able... I could have had those same methods wrapped around the content provider.

LoungeKatt:
I wasn't saying this doesn't work, but that may help for live editing. They hid it behind a misleading overview. Such is Google.

Thunder07:
@jpd002 I know you're probably busy with everyday life, work, family and working on the emulator, approving PR etc..

but for the moment I like your opinion on the concept of this PR, even if you still don't have time to check this out.
just so I know if i should keep rebasing every now and then, or if I should stop(though a bit late now) basing my work on this branch,
as I already have,
game title, overview, cover editing done,
recent to only show games that have been ran before instead of only of showing all games but sorted on last modified, suggested by @adriandennis17, and small tweaks in between.

I could push those as a separate PR but it will actually contain all of this plus that, since they're all based around the index database. if you'd like to have a look at those as well.

jpd002:
@Thunder07 Yeah, I think this is a good concept. Sorry I haven't had much time to look at it yet, I will try to do so soon.

jpd002:
That's basically been merged in #317, closing.

PR #302, opened 2015-08-17
Fix GridView item selection on Android TV

LoungeKatt:


jpd002:
Selection was indeed broken on Android TV. Thanks for the fix!

LoungeKatt:
No problem. It looks silly, but it seems something about the hardware (read: controller selection) doesn't like when you reroute the focus to a subview of a "selection" view (list, grid, etc)

PR #304, opened 2015-08-17
Android: Theme statusbar fix

Thunder07:
with transparency removed, we're no longer colouring the statusbar with the actionbar (since it was transparent, the actionbar colour showed behind the statusbar).

I also want to point out, that we can simplify ChangeTheme() a bit more by using this
```TypedArray a = mContext.getTheme().obtainStyledAttributes(new int[]{R.attr.colorPrimary});
int attributeResourceId = a.getColor(0, 0);
a.recycle();```

however, the same logic, can be used backwards, and instead of calling R.attr.colorPrimaryDark, we can define it as a colour and call it through R.color.action_bar_Blue_Dark. initial the reason for me setting it as a resource that it was just easier to call R.color.mycolor instead of the above code

Edit:
if there is a commit you don't need/want, let me know and I'll remove it, or you can cherry-pick around it.

### DIFF
--- a/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
+++ b/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
@@ -28,7 +28,6 @@ import org.apache.commons.io.comparator.LastModifiedFileComparator;
 import org.apache.commons.io.comparator.SizeFileComparator;
 import org.apache.commons.lang3.StringUtils;
 import com.android.util.FileUtils;
-import android.graphics.Point;
 
 import com.virtualapplications.play.database.GameInfo;
 import com.virtualapplications.play.database.SqliteHelper.Games;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by LoungeKatt** [comment]
I doubt you still need
```
private Toolbar getSupportToolbar() {
	View toolbar = findViewById(R.id.my_awesome_toolbar);
	final LinearLayout content = (LinearLayout) findViewById(R.id.content_frame);
	setUIcolor(content);
	return (Toolbar) toolbar;
}
```
```Toolbar toolbar = (Toolbar) findViewById(R.id.my_awesome_toolbar);``` could replace 
https://github.com/Thunder07/Play-/blob/35ed84085ae5961541b95b648b3401b0ac577293/Source/ui_android/java/com/virtualapplications/play/MainActivity.java#L92

```setUIcolor((LinearLayout) findViewById(R.id.content_frame));``` could be added to 
https://github.com/Thunder07/Play-/blob/35ed84085ae5961541b95b648b3401b0ac577293/Source/ui_android/java/com/virtualapplications/play/MainActivity.java#L95

and could replace the ```getSupportToolbar();``` at
https://github.com/Thunder07/Play-/blob/35ed84085ae5961541b95b648b3401b0ac577293/Source/ui_android/java/com/virtualapplications/play/MainActivity.java#L225

Thunder07:
Thanks for pointing it out.

jpd002:
Looking good! Thanks!

Thunder07:
Any time.

PR #305, opened 2015-08-18
Remove unused selectors (Handled by GridView)

LoungeKatt:


jpd002:
`game_selector` is still being used here: https://github.com/AbandonedCart/Play-/blob/iOS-WIP/build_android/res/layout/game_list_item.xml#L9

Was that meant to be removed?

LoungeKatt:
I removed it a while back in testing. It doesn't do anything with the GridView because the GridView's selector overlays the one for the individual view. It's similar to how the GridView's OnClick can take focus from the view's OnClick. There is a way to push the view's selector up to the top, but it's unreliable and wouldn't really look good with the covers anyway.

jpd002:
Seems to work well when using a controller, but the item is not highlighted when you touch it with your finger. Is this the intended behavior?

LoungeKatt:
That's all the GridView. As it stands, this is just there for show. It isn't actually doing anything.

PR #308, opened 2015-08-20
Android:Fix:reset long click

Thunder07:
if long click is not reset, it will be associated with another view that does not set a new long click, such as ELFs.

### DIFF
--- a/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
+++ b/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
@@ -477,6 +477,7 @@ public class MainActivity extends ActionBarActivity implements NavigationDrawerF
 		} else {
 			
 			((TextView) childview.findViewById(R.id.game_text)).setText(game.getName());
+			childview.findViewById(R.id.childview).setOnLongClickListener(null);
 			
 			final String[] gameStats = gameInfo.getGameInfo(game, childview);
 			


jpd002:
Humm, did you mean to call `setOnLongClickListener` rather than `setOnClickListener`?

Thunder07:
Oops, ya, I'll have fit fix later today.

jpd002:
Thanks for the fix!

LoungeKatt:
Not sure why this didn't go with the whole ELF block then instead of ending up a double declaration when it is reset.

Thunder07:
tbh, I was getting my branch and this repo's master confused.... which was the main reason I didn't...

but as you post this, I went looked back at it.. but I still think this works better, in the rare(?) chance TheGamesDB returns no matches, no longclick would be set, thus would cause this same problem for disc images (as oppose to just ELFs).

LoungeKatt:
Other apps, such as Gamesome, Hyperspin, Arc Browser all have that info as a click submenu to give the description, the image, and give you options such as launch and edit. Since edit is supposedly coming in the future, this would be the better way to go than long click being most likely accompanied by some proprietary detail editing menu. If that were the case, it would remove the need for all this secondary ELF stuff that's only there in case the values are null, since they will just be "" until edited.

Thunder07:
That will be the case with the edit, though as it stands,  it's done but it's ignoring ELFs, because the whole they have no longclick... But mostly because, currently we're using GameDB ID for saving the covers, which is mostly useless for elfs,  so the question is, switch cover saving to use indexdb index id or create a different system for loading covers to elfs.(I prefer the 1st, since that will also give the user flexibility when it comes to having multiple copies of the same game of different regions (can't think why they would, but if they do) each entry would contain a unique index thus won't case changing 1 cover to all of them changing) 

As for longclick vs click, it's a design preference, I for one rather no go through that dialog every time I want to launch a game, thus I favour the current behaviour, and the apps you mentioned are launcher, as oppose to emulator which generally launch a game on click... And would hide any other options behind long click.

LoungeKatt:
The difference between launchers and emulators disappeared when you decided to keep expanding all of the frontend features. This is more a launcher than an emulator as it stands now.
Honestly, details become useless as a long click. Nobody really reads that stuff. It is there to make each game's little intro screen more customized. It isn't a design preference to make it a hidden secondary screen. That is blatantly "I don't want to see this, so let's burry it"
At some point you have to stop and realize that this is still an emulator, not a demonstration of what UI tricks you can pack into something with enough traffic for it to be seen. If something is going to be made a useless feature, such as the details, then it makes more sense to remove it. Save the resources.

Thunder07:
I merely pointed out the difference since ppsspp, psx and maybe dolphin (dont have games to test that), differ from launchers....

but either ya, you do have a point... but long click was created for a reason, for extra option, not so much to burry them... but I'll add that I did make my earlier statement of having it as long click to burry it, since personally and I've mentioned this above, I wouldn't like to see it everytime I start a game.
but, the call is really @jpd002's.

and from what you write, you'd hate me for what I have to say next...
I think that menu would eventually be hidden behind another menu as well :p 
I don't know if @jpd002 would like to support per game settings eventually and if so, I'd propose long click would open up a menu, which contains for e.g, Overview, Game settings, etc. (though that can be included in the in-emulator drawer... but its not practical to have to start a game just to set the settings, though i think it should still be an option)

LoungeKatt:
It's like I said, it's better to just remove it. It's like 75% of the database size and only useful as a stylized per-game launch menu. The way it is now makes it look like just putting it there to use the info.

PR #311, opened 2015-08-27
Android:Database Import improvement

Thunder07:
I've done a couple of tests to find out what's taking the longest to complete during the initial start-up and found the code to import the database takes a full minute to complete.

So, I messed around and tested few different methods to see what can be done.
I tried using applyBatch() & bulkInsert() each with their own variation, such as using insert() inside the beginTransaction() just so I can avoid having to read each value from  ContentValues variable to set bind.... etc

but this reminded the fastest method.... it can do the full database import in a single second :)

combine this with the new search method (the one in the index PR), the very 1st clean initial launch of the app would take a full 2 seconds to complete.

Edit:
beginTransaction() causes any operation that is requested from the database to be placed in a queue/batch, instead of immediately performing it, and so when endTransaction() is called all those updates will be done in one go instead of writing to the database multiple times.

using "compileStatement" means we dont have to constantly tell the database we want to write to that specific table, though it seems like a small thing, the database contains some 4000+ entries, so time spent choosing the table adds up.

Note:( this was added as a revised commit) 
I just noticed this, but in the bulkInsert() method, I'm not using the uri to figure out the table, but instead I've it set directly... at this moment, it isn't a big deal, since the other table is not being used, if ever will be.. but for the sake of it, I'm throwing an exception to avoid the method from going ahead, should someone attempts to use it later.

Edit2:
I appreciate the line edits make it a bit harder to view the changes, but android studio can be a bitch sometimes and wouldn't let me fix that... 

Anyways, there are 3 changes, (1) I'm storing the data I want to import into a new List<ContentList> variable (2)which is passed to bulkinsert() method which replaced the old insert method, (3) the bulkinsert() method itself.

### DIFF
--- a/Source/ui_android/java/com/virtualapplications/play/database/TheGamesDB.java
+++ b/Source/ui_android/java/com/virtualapplications/play/database/TheGamesDB.java
@@ -10,6 +10,7 @@ import android.database.SQLException;
 import android.database.sqlite.SQLiteDatabase;
 import android.database.sqlite.SQLiteOpenHelper;
 import android.database.sqlite.SQLiteQueryBuilder;
+import android.database.sqlite.SQLiteStatement;
 import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Bundle;
@@ -31,27 +32,27 @@ import com.virtualapplications.play.database.SqliteHelper.Games;
 import com.virtualapplications.play.database.SqliteHelper.Covers;
 
 public class TheGamesDB extends ContentProvider {
-	
+
 	// Database Name
 	private static final String DATABASE_NAME = "games.db";
 
     // Database Version
     private static final int DATABASE_VERSION = 2;
-	
+
 	public static final String SQLITE_ID = "_id";
 
 	private static final UriMatcher sUriMatcher;
-	
+
 	private static final int GAMES = 1;
 	private static final int GAMESID = 2;
 	private static final int COVERS = 3;
 	private static final int COVERSID = 4;
-	
+
 	private static HashMap<String, String> gamesMap;
 	private static HashMap<String, String> coversMap;
-	
+
 	private static Context mContext;
-	
+
 	private static class DatabaseHelper extends SQLiteOpenHelper {
 
 		DatabaseHelper(Context context) {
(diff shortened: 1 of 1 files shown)


jpd002:
Looking good! I'll give that a try soon.

LoungeKatt:
TL;DR this is why I was saying a content provider would be useful.

Thunder07:
"TL;DR this is why I was saying a content provider would be useful."

from what point of view.... this 'fix' is to the content provider.

LoungeKatt:
The difference between us is that I read the code, not the comment ;)

Thunder07:
still, the content provider bulkInsert() method had to be overridden to make it useful... as such its a generic method which has no relation to the content provider... which as such can be used without the content provider.

my point is, in no way do we owe our thanks to the content provider.

LoungeKatt:
Whatever. You talk too much.

Thunder07:
"performance gain of over 1000 Percent"... but thats the thing, content provider provides no performance gain at all...
any gain would be because of the overridden code in applyBatch() or bulkInsert()...(since the default functions loop and insert without even a transaction, thus they'll right to the database for each entry)
...
and since you know content provider is a mere wrapper around the database, that same code can be used without the provider...

I'm not saying the content provider is useless, just in this case or any future performance gains, will and can't be directly or even indirectly attributed to the provider.

LoungeKatt:
The thing of it is you're not an expert, just reciting documentation. That makes it a moot point to claim anything is or isn't based on your own interpretation of it.

Thunder07:
"you're not an expert" and you are? and who said I'm not?

I could list the dozens of stackoverflow, post and web pages, non of which say content provider provides speed improvements, as a matter of a fact, some say it's actually slower.. but all of them have "sharing data across apps" as usually the single reason for using the content provider.

LoungeKatt:
I am and I just did. Sorry, but this is one you can't prove by attrition.

Thunder07:
this last revision of the commit just uses new lines for the if statements, remove else bindnull and instead calls stmt.clearBindings(); instead.

jpd002:
Sorry I didn't take a look at this yet, but I will try to do that this week.

Thunder07:
I'm in no rush really :p

jpd002:
Thanks! Gives a great speed improvement to the initial load up time!

PR #317, opened 2015-09-11
Android: Indexing + Games Detail Editing + few others

Thunder07:
@jpd002 sorry for the long PR, unfortunately, since this is a combination of the Indexing PR and commits based on it, it can't really be split up, however, you can still review the Indexing PR separately 1st then do this, since everything upto "Android: Comment out logger" is Indexing.

I've also started college again, So im not sure how much time will I have, so i opted to post everything out.

anyways, back to this PR...
everything upto "Android: FileWalker class License Added"(edit*) is Indexing is from the Indexing PR (So I'll avoid commenting on it here)

Everything from "Android:Allow user to edit games details" to the end has a descriptive enough commit name, if you need me to elaborate on something please ask. 

The only thing I'm not happy about is GameInfo.getImage() method change... https://github.com/jpd002/Play-/commit/4592025f89f60c3dffd6ae3824acbba51c073190#diff-317ea4e5dffe1b9fbacc0f8bb11e13c1R130
not because its bad or anything, I just couldn't think of any other way to do it, as such I have no other implementation to compare it against.

Finally, I'd like your opinion on this, we should switch to saving the images based on their Index ID instead of TheGamesDB, as the current implemention means, should the user (if some unknown reason) have 2 copys of the same game, and should he choose to set a custom cover, both games cover would change, since they both share a non-unique ID, on the otherhand their IndexID would be unique.


jpd002:
Thanks! Now that I've merged the iOS changes and that I'm back from vacation, I'll take some time to look at your PRs. Expect some feedback tomorrow. My goal is to merge that by the end of next week.

Thunder07:
God, how I'd love a holiday right about now.

I rebased this PR on both latest master and Indexing PR to make sure you can merge either or without conflicts and one commit was removed since it was incorporated into another commit.
workload for 1st week of college seems light, so hopefully I can fix anything you might find during this week.

jpd002:
I've tested that branch and it's working great. I still need to finish my code review, but I've noticed that you've included a copy of the Apache license in there. Do we need to add that even though you mention that your code is inspired by Findex's code?

Thunder07:
I'm not entirely sure, as I'm not too familiar with licenses, which is why I opted to add it just to be sure...
but I suspect it's not needed, as while the license is included in their project, it was not applied, as in it's not copyrighted to anyone. (but one could argue it's implied to the original uploader).
though i dont want to come to conclusions, I'd like to think anyone who pushes something with the intent of it being opensource won't be too aggressive, especially, since we're publishing the changes and allowing other to use the code should they like.

I really wish I could be of help, i even went online when i was working on this, but no one could give me a clear answer to.

jpd002:
Well, since you're crediting him inside the code (which is inspired by what he's done rather than a verbatim copy), I think it's safe not to add the license in there.

Thunder07:
I've revised the pull and remove the license from it.

build_android/res/values/strings.xml
```diff
@@ -42,8 +46,14 @@
     
     <string name="preferences_ui">Storage Options</string>
     
-    <string name="pref_ui_reset_folder_title">Reset Folder</string>
-    <string name="pref_ui_reset_folder_summary">Clear stored default folder</string>
+    <string name="pref_ui_rescan_all_title">Rescan Storage</string>
+    <string name="pref_ui_rescan_all_summary">Scan storage for new games</string>
+
+    <string name="pref_ui_clear_index_title">Clear Game List</string>
+    <string name="pref_ui_clear_index_summary">Reset sll of the game\'s list and its associated details</string>
```
jpd002: Small mistake here, should be "all" instead of "sll". Should also be "game list" instead of "game's list".

build_android/res/values/strings.xml
```diff
@@ -42,8 +46,14 @@
...
+
+    <string name="pref_ui_clear_unavailable_title">Clear Unavailable Games</string>
+    <string name="pref_ui_clear_unavailable_summary">Remove any game currently inaccessible from the game\'s list</string>
```
jpd002: Should be "game list" instead of "game's list".

jpd002:
Thanks for the update. I've tried the buttons in the settings activity ("Rescan storage", "Clear Game List", etc.) and I was wondering... are they supposed to disappear after they're being pressed? On my side, they don't disappear at all.

Source/ui_android/java/com/virtualapplications/play/database/GameIndexer.java
```diff
@@ -0,0 +1,202 @@
...
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    public String getSerial(File game) {
+        String serial = NativeInterop.getDiskId(game.getPath());
+        Log.d("Play!", game.getName() + " [" + serial + "]");
```
jpd002: I think that should be commented out to reduce log spam.

Thunder07:
The button disappearing seems to have broken when we sub divided the preferences.
as for the "Rescan storage" not updating the list, its because the list can't

I'm just testing everything out before I push it, should be out within the next 5mins. anyway I have it all under one commit as I don't have much time today, if that's an issue or if you'd like them to be integrated into previous commit (when possible) I can get it done 1st thing this weekend.

jpd002:
Thanks again for the changes. I'll be merging this tomorrow as it is looking good.

I'm not totally sold on the buttons in the settings activity. I think it might be interesting to have a single button in there that would re-synchronize everything we've got from the local filesystem with what we've got in the index database (deleting items that are not on the local filesystem anymore). Anyways, I guess that would be easy to change later on.

And yes, I think every item in the index should be able to have its own disc cover. The Games DB should only be a "source" from which we can pull information for a game inside the index.

Thunder07:
Glad to hear.

as for the button,
"Clear Game List", I guess there is no real need to clear all the games (if only til the next scan they'll be added).

so you want to combine "Rescan Storage" & "Clear Unavailable Games"?
I thought about it, but my concern is to do with people using external sdcard, if they ever ran clear unavailable without the sdcard plugged it, all their games would disappear, as it stands, it isn't all thw big of a deal, but some day in (hopefully) not the distant future you'd probably add per game settings, now with such option, everytime they do this, they'll lose all their games settings (since I'm assuming the setting would be tied to Index ID).

I wont have time to do anything today, something came up.
but I was thinking about how to keep the grid updated one way to do it, it to do a partial scan (folder with games) everytime Play! comes into focus(after being sent to the background) with onResume() or adding a context menu with option to "Refresh"/"rescan folders".
though I prefer the 2nd option, I also don’t want people to expect to find their newly added games (which might have been placed in a new folder), when we're not looking for them.

jpd002:
Oh yeah, I hadn't thought of the case where the user would have removed his/her sdcard temporarily. I guess having a separate "Clear Unavailable Games" button would make sense then. But yeah, "Clear Game List" isn't really needed.

As for keeping the list updated, I think I would prefer the first option. I don't like having too many options and buttons in there, it's better to keep it minimalist and intuitive.

And yeah, GameInfoEditActivity would probably be a better name. You can make that change in another PR if you'd like since I'm merging this one now.

PR #320, opened 2015-09-22
Android:save user sort preference, highlight user selection on startup, display sort selection in NavBar subtitle

Thunder07:
mDrawerListView.post() might seem weird, I think it is anyway, but thats the only way i found for setting the selection highlight on startup, every otherway returns a null.


jpd002:
Looks good to me. I'll do it myself this time, but don't hesitate to comment such "weird" things in the code next time so we can figure out why it was done this way later on.

Thunder07:
Will do thanks.

PR #340, opened 2015-10-21
modularisation changes

bigianb:
These are the changes that I have to support the BGDA HLE app. I'm happy to maintain a fork as the changes are small, but am interesting in comments from JPD.
The changes are summarised as follows:
1. Put all the code except for the main function into a library. This allows all the code to be re-used in a separate project.
2. Allow the EE executor to be set dynamically in the PS2VM - this allows the BGDA app to replace the standard one with a custom executor supporting hle
3. made execute and compile virtual in BasicBlock so that an hle version of the basic block can be used.
4. Allow sub-classes of MainWindow to specify alternative GS Handler implementations.

The repo https://github.com/bigianb/hle_play demonstrates how this is used (in a very initial immature manner)


jpd002:
Thanks for submitting this! Looks quite interesting! I'll take a look when I get the chance.

jpd002:
So, finally got around trying this branch. Unfortunately, I can't seem to find the "snowlib" project. Do you know where I can find it?

bigianb:
Ah it looks like that repo is set to private. I'm away in Germany until weds so I will make it public when I get back. To test, just use a revision from before that was added (you'll need to press x a few times to get past the invisible menus in that case)

Snow lib is just some routines to decode snow blind engine textures and models.

Hle play is currently bgda specific. I need to add some code to look at the sles/ slus Id of the title being loaded and switch in an appropriate hle set.

I suspect you will want to make play lib differently than I did - it's really just the play project without the main function in it. Then play project then just has the main function.

Ian

PR #344, opened 2015-11-06
Android:remove duplicate string

Thunder07:
this string also exists in donttranslate.xml
it also prevents Android Studio from building, with duplicate strings error.


jpd002:
Thanks for the fix!

PR #348, opened 2015-11-22
Android:Layout Clean up

Thunder07:
Setting Cover Scaling through XML layout instead of calling it through java every single time.

I've also noticed that file_list_item.xml isn't used, since we don't have an initial file selector, I've updated it and kept it, instead of removing it, should there be a need for it in the future, if you don't think so, you/I can remove it.


jpd002:
Thanks, a clean up is always welcome! 

I think the `file_list_item.xml` file should be removed if it's not needed and since you're making a cleanup PR, I think you should remove it :smile:. Once it's removed, I'll merge this.

jpd002:
Awesome! Thanks a lot for making the change!

PR #376, opened 2016-01-21
Enable x86_64 Android

maximu:
hi All android x86_64 roms launched so far have a bug in build.prop

ARMv8 prioritize binary load before the 32-bit x86 version if x86_x64 binary not exist.

This bug is fixed in the latest android-x86 6.0 branch source git updates a few days ago.

We needed  add x86_64 binary for prevent this error in bug build.prop roms.

### DIFF
--- a/build_android/jni/Application.mk
+++ b/build_android/jni/Application.mk
@@ -1,5 +1,5 @@
 APP_PLATFORM			:= android-19
-APP_ABI 				:= armeabi-v7a arm64-v8a x86
+APP_ABI 				:= armeabi-v7a arm64-v8a x86 x86_64
 APP_STL 				:= c++_static
 APP_CPPFLAGS 			+= -std=c++11
 NDK_TOOLCHAIN_VERSION 	:= clang


jpd002:
I was planning to enable x86-64 support on Android when a new version of the NDK would be available because I wasn't able to test the 64-bits version yet on my Nexus Player due to a bug with the current version of the NDK. 

But, I don't see any objections to enable it right now if you need it for your tests. So, thanks for the PR.

maximu:
I did not know there was a bug with the NDK.
What happened in nexus player?
I just wanted fix headaches detection in many roms to load ARMv8 emulation before a binary 32bit x86 if x86-64 binary not exist by a misconfiguration on the build.prop

I use Android x86 project for my tests http://sourceforge.net/projects/android-x86/ they are good for this supporting very well perfomance

Thx for merge :+1:

jpd002:
https://code.google.com/p/android/issues/detail?id=179410

maximu:
ah I see thanks 

https://android-review.googlesource.com/#/c/182443/1/sources/cxx-stl/gabi%252B%252B/src/cxxabi.cc

It seems that they have already solved have to use some changes in 18-November last year in the NDK source master google branch to fix that error... hopefully Google will soon upgrade their NDK

maximu:
@jpd002 I can't reproduced you bug said working fine with android-x64 lolipop image  in my hw no crashes detected :) 

screenshot_2016-01-25-13-27-39

PR #380, opened 2016-01-25
Android permission

Thunder07:
getColor was crashing with the new library and so i was replaced it.

Edit: I've also noticed as i was working on the new permission system, that Internet permission is no longer required, as such only 1 permission was needed with this new system.

Edit2:
should have mentioned, for this to work, you'd also need support library v4/v7 updated to 23, and I've made a PR for them as well https://github.com/jpd002/Play-Dependencies/pull/4.

### DIFF
--- a/Source/ui_android/java/com/virtualapplications/play/Constants.java
+++ b/Source/ui_android/java/com/virtualapplications/play/Constants.java
@@ -3,4 +3,6 @@ package com.virtualapplications.play;
 public class Constants
 {
 	public static String TAG = "Play!";
+	public static final int READ_WRITE_PERMISSION = 0;
+
 }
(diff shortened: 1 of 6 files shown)


jpd002:
Thanks for this, I've tested it on my side and it's working good. 

There's just a small issue with indenting though, it should be done with tabs instead of spaces (example: https://github.com/jpd002/Play-/pull/380/files#diff-c4257fa5734cb0df77279a55f969c72aR53).

Thunder07:
thats odd, its all tabbed out here, ill pull the branch from my repo and see whats up later when i get the chance.

jpd002:
Thanks for the fix!

Thunder07:
Sure thing

PR #391, opened 2016-04-03
Fix MacOSX build

Thunder07:
since this commit https://github.com/jpd002/Play-/commit/9a3ca1b2d1a95d59d26c35017615c5882a5d6538 when you started using "glTexStorage2D" in GSH_OpenGL.cpp the OS X build refused to compile as glTexStorage2D was not defined.

 I did a bit of googling and it seems glTexStorage2D is in OpenGL/gl3ext.h this fixed the issue, without breaking android build, but I'm not sure if this will effect iOS as i don't have a device to test.

Edit:
the app crashes after selecting a game to boot, I'm guessing thats because this branch is still in development.

Edit2: i was building with debug, building with release games run fine with the fix.

### DIFF
--- a/Source/gs/GSH_OpenGL/GSH_OpenGL.h
+++ b/Source/gs/GSH_OpenGL/GSH_OpenGL.h
@@ -8,6 +8,9 @@
 #include "opengl/Program.h"
 #include "opengl/Shader.h"
 #include "opengl/Resource.h"
+#if defined(__APPLE__)
+#include "OpenGL/gl3ext.h"
+#endif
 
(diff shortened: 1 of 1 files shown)


jpd002:
Thanks for reporting this, I had not tested that branch on OSX and it indeed fails to compile on that platform. Adding that include looks like the proper fix, but I think the include should be added there instead: https://github.com/jpd002/Play--Framework/blob/master/include/opengl/OpenGlDef.h#L29. That file already has all the "platform specific" includes for OpenGL.

If you do a PR in the Framework repo, I'll merge it there.

PR #393, opened 2016-04-11
Fix building iOS with Xcode

LoungeKatt:
Explained by the commit headers. Nothing overly complex.


LoungeKatt:
This will NOT merge as is. This is simply getting it running before coming up with a version that works with the current layout, as I use a static version of boost to avoid the messy compile.

jpd002:
@LoungeKatt Alright, duly noted. Just tell me whenever the branch will be ready.

LoungeKatt:
Not sure what the write conflicts are, but I restored all the original links from the variation of boost I was using. The ImageIO and Foundation frameworks are only in the pod now, which cleans up the main project a little. The downside is that you don't really get to store the pods in another location, such as dependancies and for this one it requires manually integrating the custom configuration with the configuration from the pod (since their merge resolution is to discard one altogether)
The cocoapods version I use is is installed using `sudo gem install cocoapods -- pre`

LoungeKatt:
Looks like it just needed to be rebased. Didn't realize there were new commits.

Fadi5555:
@jpd002 when you will merge these branches it ready 4 days ago.

LoungeKatt:
@Fadi5555 If you are building it locally, you can cherry-pick the commits for testing. If not, there is absolutely no reason to push for a merge since you cannot verify the changes are working.

Fadi5555:
I didn't mean anything i just want to remind him about the branch and it ready to merge.
That's it. anyway i'm sorry.

LoungeKatt:
It's been nearly a year now. It would be easier to be patient than nagging, apologizing, nagging, apologizing. You are well aware it will be ready when it is ready and not a moment sooner.

Fadi5555:
Thank you so much

jpd002:
After testing on my side, I've decided not to bring all the changes as they are in your branch, because I don't especially agree with the way CocoaPods is used: I feel automatic project integration is a bit dangerous and I'd like to avoid adding dependencies to source control, so I prefer leaving the "Pods" folder out of it.

But I think using CocoaPods is a good idea, so I've made the changes to use it to update SDWebImage. I've also imported the changes you've made to the database and to the way boxart is fetched: https://github.com/jpd002/Play-/commit/3ff69882e506aa1fe2c917d8a6d6089068348e22.

While it might fix some of the issues with boxart, I don't think it'll fix the problem @Fadi5555 has been referring to. AFAIK, none of the database rows are being updated and since most of the rows seem to have a null value for their boxart URL, nothing will be downloaded most of the time. It's not a deal breaker for me, but I'm just settings expectations here.

Fadi5555:
That's so sad. I really hoped to fix boxart issue. at least to matching android version

LoungeKatt:
The pods folder is CocoaPods though... That's how it functions.

build_ios/Pods/Target Support Files/Pods-Play!/GeneralSettings.xcconfig
```diff
@@ -9,13 +9,13 @@ ENABLE_BITCODE = NO
 CODE_SIGN_IDENTITY = -
 
 OTHER_CFLAGS = -fvisibility=hidden $(inherited) -isystem "${PODS_ROOT}/Headers/Public" -isystem "${PODS_ROOT}/Headers/Public/SDWebImage"
-OTHER_LDFLAGS = $(inherited) -ObjC -l"SDWebImage" -framework "ImageIO" -framework boost
```
LoungeKatt: -framework boost should not have been removed from the submission version of this commit

build_ios/Pods/Target Support Files/Pods-Play!/GeneralSettings.xcconfig
```diff
@@ -33,9 +33,9 @@ GCC_WARN_UNDECLARED_SELECTOR = YES
 GCC_SYMBOLS_PRIVATE_EXTERN = YES
 GCC_INPUT_FILETYPE = sourcecode.cpp.objcpp
 
-FRAMEWORK_PATH = ../../Framework/
-CODEGEN_PATH = ../../CodeGen/
-BOOST_FRAMEWORK_PATH = ../../Dependencies/BoostMac/ios/framework/
+FRAMEWORK_PATH = ${PODS_ROOT}/../../../Framework/
+CODEGEN_PATH = ${PODS_ROOT}/../../../CodeGen/
+BOOST_FRAMEWORK_PATH = ${PODS_ROOT}/../../../boost_1_58_0
```
LoungeKatt: ../boost_1_58_0 should be ../Dependencies/BoostMac/ios/framework/ in the submission version of this commit

jpd002:
I admit that there might have been some confusion on my side. It's my first experience with CocoaPods so, I probably misunderstood its purpose.

For the `Pods` folder, the [official documentation](https://guides.cocoapods.org/using/using-cocoapods.html#should-i-check-the-pods-directory-into-source-control) states that it's not necessary to add it in source control. I'm not totally against the idea and I can understand the advantages it offers (the best being not having to run `pod install` every time the repo is cloned), but I feel it makes things a bit more cluttered when external dependencies are added in the repo. Anyways, I wouldn't really mind adding the `Pods` folder to the repo if you feel it should be in there.

As for the direct integration feature of CocoaPods, whilst being an handy feature, it doesn't seem to be compatible with the way the project is currently configured. Direct integration requires the project to use the `xcconfig` file generated by CocoaPods and Play! is currently using `xcconfig` files to store its build settings. Xcode only allows using one `xcconfig` file per target. Furthermore, there's no way to create a proper inheritance chain for properties defined inside an `xcconfig` file using `#include` directives (http://stackoverflow.com/a/1401682). To allow direct integration to work, build settings from `GeneralSettings.xcconfig` would need to be moved back to the project file, which would be possible now that the need for `ExternalDependencies.xcconfig` has been removed. I feel it would be a step back for the cleanliness aspect though: I think it's much easier and clearer to have all settings exposed in a single `xcconfig` file.

Also, from what I understand, the `Pods` folder is kinda like an intermediate build folder, as such, it should be possible to erase it and regenerate it from scratch using the `pod install` command. Considering this, it doesn't seem like a good idea to make manual changes to the files located inside the `Pods` folder as the changes would need to be reapplied again when using the `install` or `update` command. I might be wrong on this because I haven't tested it personally.

But, yeah, not using direct integration also has its downsides. The `GeneralSettings.xcconfig` has to be updated every time the structure inside the `Pods` folder changes. The project file also has to be updated manually since there references to `Pods.xcodeproj` and its build artifacts. This makes the pod adding/updating/removing workflow tedious.

So, that pretty much boils down to getting rid of the `GeneralSettings.xcconfig` file and letting CocoaPods integrating itself in the project using its `xcconfig` files without having to change anything inside the `Pods` folder. I'll have to think about this a bit more to see if there's any major problems with that.

There were also some changes in this PR that didn't seem to compile on my side, namely: https://github.com/jpd002/Play-/pull/393/commits/f7fbe6d261ca549e8cf0186bd46cf67bc10f99e4#diff-e2316950d3eac3e724f60e82af7515b6L80. [Documentation](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIResponder_Class/#//apple_ref/occ/instm/UIResponder/touchesBegan:withEvent:) states the `*` is required and the file won't compile without it.

TLDR; Thanks for the feedback, current project configuration makes CocoaPods integration hard, build settings would need to be moved back to the project.

LoungeKatt:
It is odd the documentation would recommend the * when it was XCode that pointed it out as an issue that could not be built. That may be another of Apple's wonderful "only in XCode version X" issues, although I was trying with both 6 and 7 and neither would accept it.

jpd002:
Alright, it's been a while, but I've moved the build settings from the `GeneralSettings*.xcconfig` files to the project file. CocoaPods' standard project integration can now be used without any change to the configuration.

https://github.com/jpd002/Play-/commits/ios_coverfix

Tell me what you think.

jpd002:
The changes for CocoaPods and the update to the latest SDWebImage library are now in master. The changes from other commits either don't apply anymore or don't build with my setup (Xcode 7.3 with latest iOS SDK), so, I'm not going to bring them in my repo for now. If you fix them, feel free to submit again.

LoungeKatt:
The entire purpose of those commits was support for building with older version of Xcode. There is nothing to fix for 7.3 because they added compatibility for things only available after 7.0
I am sure if the Cydia repository is maintained, that won't be an issue. The problem was changes piling up between published builds. Just make sure to verify support for older models. Not everyone updated to the iPhone 6 and the 5s is still widely used by many iPhone owners.

PR #414, opened 2016-05-24
implement vsprintf. Lego starwars uses it in its sound driver.

bigianb:
The code is mainly a duplicate of the existing printf code. The printf code could be refactored to use vsprintf in a similar manner that the standard library does. For anyone not familiar with the emulator wondering why we don't just call the system vsprintf - the problem is that any %s argument will be a ps2 pointer and so needs converting to a real host pointer before it is de-referenced.

### DIFF
--- a/Source/iop/Iop_Sysclib.cpp
+++ b/Source/iop/Iop_Sysclib.cpp
@@ -1,5 +1,9 @@
 #include "Iop_Sysclib.h"
 #include "../Ps2Const.h"
+#include <stdarg.h>
+#include "string_format.h"
+#include <boost/lexical_cast.hpp>
+#include "lexical_cast_ex.h"
 
 using namespace Iop;
 
@@ -85,6 +89,9 @@ std::string CSysclib::GetFunctionName(unsigned int functionId) const
 		break;
 	case 41:
 		return "wmemset";
+		break;
+	case 42:
+		return "vsprintf";
 		break;
 	default:
 		return "unknown";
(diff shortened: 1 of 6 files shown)


Source/iop/Iop_Sysclib.cpp
```diff
@@ -241,6 +248,16 @@ void CSysclib::Invoke(CMIPS& context, unsigned int functionId)
 			context.m_State.nGPR[CMIPS::V0].nD0 = context.m_State.nGPR[CMIPS::A0].nV0;
 		}
 		break;
+	case 42:
+		// int vsprintf (char * s, const char * format, va_list arg );
+		{
+			context.m_State.nGPR[CMIPS::V0].nD0 = static_cast<int>(__vsprintf(
+				reinterpret_cast<char*>(&m_ram[context.m_State.nGPR[CMIPS::A0].nV0]),
+				reinterpret_cast<const char*>(&m_ram[context.m_State.nGPR[CMIPS::A1].nV0]),
+				reinterpret_cast<va_list>(&m_ram[context.m_State.nGPR[CMIPS::A2].nV0])
```
jpd002: Is it possible to use `va_list` this way? I thought it was only possible to use it with `va_start` and `va_end` along with a `...` parameter.

jpd002:
Thanks for this. 

Actually, I tried using `vsprintf` in an hardware test and I wasn't able to get it to work. The PS2SDK mentions something about that function being broken (https://github.com/ps2dev/ps2sdk/blob/master/iop/debug/sior/src/xprintf.c#L8). So, I'm not quite sure this is actually implemented in the BIOS. One thing I haven't tested yet is to use a "recent" IOPRP image in hardware tests and see if it's implemented there.

That being said, I think it would be easier and nicer to change the already existing `PrintFormatted` function to handle the case where a `va_list` is supplied instead of `...`. It might be necessary to make that function a template to make that possible.

Source/iop/Iop_Sysclib.cpp
bigianb: yes, you have to do it this way because va_list is passed to you. va_start initialises a va_list by taking the argument before the elipsis and setting the va_list value to the correctly aligned pointer. We're passed va_list (which is just a pointer) so we don't want to initialise it. va_end just sets the va_list pointer to null so you cannot accidently use it again. The documentation for va_* isn't really that clear.

bigianb: Actually you can see this use pattern in the implementation of vxprintf in the file you link to below.

bigianb:
I agree about the code duplication with PrintFormatted - hence my comment. I was just trying to get something up for lego star-wars (the sound routines use vsprintf to write logging to stdout). I thought about a template at the time, but it's not simple because there is not a generic interface for both argument accessing methods.
A modern solution would be to pass a lambda which is used to access successive arguments, An easier solution would be to use the pre-processor to have 2 versions of the print formatted routine (one uses va_arg and the other the existing argument accessor).

bigianb:
I've updated the code to use a common formatter now.
Lego batman stdout looks good.

PR #417, opened 2016-05-26
alias cdrom to cdrom0

bigianb:
tribes ariel assault tries to load game data from cdrom: rather than cdrom0:


jpd002:
It should be safe and will be cleaner to add the alias there: https://github.com/jpd002/Play-/blob/9c719bcf75b7ef7f98f0d000578ccad2021ca648/Source/PS2VM.cpp#L367

`CIsoDevice` doesn't own any memory.

jpd002:
Turns out Need for Speed: Carbon also needed that, so I've made the change myself: 9eedb56f70115ed32cfee1568738fcc0bc5dad1e. Thanks.

PR #422, opened 2016-06-06
Lego spike

bigianb:
Deal with the special case where an integer register is written immediately before a conditional branch which uses it. Fixes Lego star wars.

### DIFF
--- a/Source/MIPS.h
+++ b/Source/MIPS.h
@@ -102,6 +102,9 @@ __attribute__((aligned(16)))
 	uint32				cmsar0;
 	uint32				callMsEnabled;
 	uint32				callMsAddr;
+
+	uint32				savedIntReg;
+	uint32				savedIntRegTemp;
 };
 
 #define MIPS_INVALID_PC			(0x00000001)
--- a/Source/ee/VuBasicBlock.cpp
+++ b/Source/ee/VuBasicBlock.cpp
@@ -2,6 +2,8 @@
 #include "MA_VU.h"
 #include "offsetof_def.h"
 
+#pragma optimize("", off)
+
 CVuBasicBlock::CVuBasicBlock(CMIPS& context, uint32 begin, uint32 end)
 : CBasicBlock(context, begin, end)
 {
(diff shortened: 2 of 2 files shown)


jpd002:
Awesome, thanks a lot!

Preliminary testing on my side shows that it fixes rendering in Crash Bandicoot: Wrath of Cortex (only shows loading screen though) and allows Kingdom Hearts 2 to enter in-game (visuals are quite screwed up though).

Fadi5555:
@jpd002 did you mean this will fix some games with black screen like crash bandicoot the warth of cortex?

Other thing when the next build will come? Because you said you'll merge the fix of boxart.
Thanks

bigianb:
The lego series (this also fixes Lego batman) have a similar issue - they are all travellers tales games so I suspect they share a code base. I suspect it may be due to how floating point denormal values are handled as I remember pcsx2 having similar issues several years ago before they fixed up the floating point handling.
Note that there is still an edge case not accounted for in this code - if the block is less than 6 instructions then we will not save the int reg value ... in this case we should save it at the start of the block. It's a simple fix, I just forgot about it last night.

jpd002:
Planning to merge the branch this week-end. I'm going to run over my test plan with some other compatibility fixes I've made.

jpd002:
I've tested this on a few games and found some problems. BGDA is one of the games that don't work with this due to one of the loop it has in its VU microcode:

```
00001260    000002FF 10010040    NOP                                           IADDIU          VI1, VI0, $0040
00001268    000002FF 10020000    NOP                                           IADDIU          VI2, VI0, $0000
00001270    000002FF 100310FB    NOP                                           IADDIU          VI3, VI2, $00FB
00001278    000002FF 81E20B7D    NOP                                           SQI             VF1xyzw, (VI2++)
00001280    000002FF 81E2137D    NOP                                           SQI             VF2xyzw, (VI2++)
00001288    000002FF 81E21B7D    NOP                                           SQI             VF3xyzw, (VI2++)
00001290    000002FF 520317FC    NOP                                           IBNE            VI3, VI2, $00001278
00001298    000002FF 81E2237D    NOP                                           SQI             VF4xyzw, (VI2++)
```

I've isolated the test case in an hardware test here: https://github.com/jpd002/ps2autotests/commit/6ef239aa1b218a5090df1ad0bfeb9671d6e2430c

I don't really know what's supposed to be the "rule" here, but one weird thing about the test result is that the counter goes all the way up to 0x100 even though the test value is 0xFB. I know you have lots of knowledge about BGDA's internals, so you might be able to shed some light on this.

I'll try to find other edge cases to find a way to make sense of all this.

bigianb:
Great thanks for testing. This is due to bad handling of the edge case I mentioned. I'll explain more later and think about resolutions.

bigianb:
So, the rule here is the same as always - the value used is the one set 5 instructions earlier. The edge case here is that the set instruction in question is in the previous iteration of the loop (I guess at address 1288). In this PR, I say that if the block is less than 5 entries long, take the value at the start of the block - but this is wrong because it would take the value set in the delay slot.
I'm not sure of a simple method to fix this properly - but it should be pretty simple to make the code work in the lego games whilst causing no greater regression in other games like BGDA. In the case that the block is less than 5 instructions, just omit the conditional branch test. That way, code like BGDA works like it always did (broken ... although in BGDA I think it makes no difference) whilst games such as lego work.
I won't have a chance to look at this before Tuesday - feel free to make the required change and merge if you have time ... or I'll drop a fix in later this week for review.

bigianb:
Previous commit makes the change I described. The 'fix' is now only performed when we know it is safe. When we cannot be sure when the integer register was set (or when it's value became effective to be precise) then we use the original (but still broken) implementation.

jpd002:
Thanks for the update. Your change fixed BGDA. KH2 doesn't work anymore though, but I can live with that since it didn't work with master anyways.

Been continuing my tests and I've found that we don't handle the "flag testing operation right before a branch" case properly. Documentation states that all flag testing operations don't cause a delay with the value used within branch operations.

I've updated the hardware test accordingly: https://github.com/jpd002/ps2autotests/commit/929c6e852bfc06021a08ad21ea14276935ad140a

I've made some changes on my side that add a new `readFlags` field in `OPERANDSET` which can be used to disable the delay code. I'll push that commit tomorrow.

jpd002:
Added the necessary stuff in 28c96a56f8809ac1625bfdd18e014255b520d758. `readFlags` can be tested to disable the delay code.

I've tested a bit more today and I've seen problems with Romancing Saga which now displays black characters. Still don't know the exact cause yet.

jpd002:
Seems like `ILW` is another exception to the rule even though it's not explicitly documented. That was the reason behind Romancing Saga's problem.

Updated hardware test: c16538652e73ac887f97928405e5431af3f6e74a.

Not too sure about the `readFlags` field in `OPERANDSET` now. I think the hardware test will also need to be improved further to catch all potential problems.

bigianb:
wow, great find on the ILW. I guess we need something to generate code to test all the possible instructions before a conditional branch.

jpd002:
I've updated the test to make testing of single instructions a bit easier ([result](https://github.com/jpd002/ps2autotests/blob/branchdelay_internal/tests/vu/lower/branchdelay.expected)). No huge surprises... `ILWR` exhibits the same behavior as `ILW` and others seem to follow the same pattern as basic integer operations.

Gonna rename `readFlags` to something more sensible and resume testing games.

jpd002:
Ok, after adding a special "case" for BGDA's situation, I've gotten to a point where all games I'm testing works with your changes (KH2, Lego Star Wars and Crash Bandicoot included). Gonna merge that tomorrow.

PR #423, opened 2016-06-14
Linux UI

Thunder07:
following are the required packages that you'd need to build this PR.

for starters you'd need [qt 5.6.*](https://www.qt.io/download-open-source/#section-2), you can also install just the cmdline packages using `sudo apt-get install qt5-default` but this will install 5.5.1 which still doesnt contain this [patch](https://codereview.qt-project.org/#/c/139775/13/src/widgets/kernel/qopenglwidget.cpp), so you'll be spammed with that error, for now i'd avoid the apt package.
note: I'd also avoid using Qt 5.7  due [failed to compile gtk plugin] (https://bugreports.qt.io/browse/QTBUG-53844)
 
```
sudo apt-get install python-dev libbz2-dev libalut-dev libmysqlclient-dev libglew-dev libicu-dev cmake git -y
sudo apt-get install libboost-filesystem-dev libboost-chrono-dev
sudo apt-get install -f
```

~~you'll also need boost, i downloaded and used the latest version, at the time it was/is 1.61.0, if you use a newer or older version, just make sure to edit ./Play/build_unix/ui_unix.pro to reflect that.
when building boost make sure to build it with static linking (just to avoid having the end-user having to install boost), I've also chosen to build it with speed optimisation and multi-threading.
[Boost package](https://sourceforge.net/projects/boost/files/boost/1.61.0/), download 1 directory outside of the project directory, see below of directory tree.~~
```
tar xvf boost_1_61_0.tar.*
cd boost_1_61_0
./bootstrap.sh
./b2 link=static link=shared threading=multi optimization=speed --with-chrono --with-filesystem --with-system
./b2 link=static link=shared threading=multi optimization=speed  --with-chrono --with-filesystem --with-system install #optional, i've avoided installing it for ease of future updates
```

~~~now all you need to start building is~~~
```
cd Play/build_unix/
export BOOST_INCLUDEDIR=/full/path/to/boost_1_61_0
./build.sh
```


~~~this is what my project directory looks like~~~
```
.
├── boost_1_61_0
└── Play-Build
    ├── CodeGen
    ├── Dependencies
    ├── Framework
    └── Play
```


jpd002:
Thanks for this. I know I've said I would take a look at your other Linux UI branch, but I hardly has much time for me recently. I'll try to be quicker this time.

Also, why not install boost using your distro's package manager?

Thunder07:
ubuntus ppa latest boost version is 1.55 which at this stage is just under 3 years old, but boost.org latest release is 1.61 which was upgraded from beta to release a month ago, so i decided to go with that.

Source/PH_Generic.cpp
```diff
@@ -1,4 +1,5 @@
 #include <cassert>
+#include "string.h"
```
jpd002: This should be `#include <cstring>` (angular brackets since this is a system include and `cstring` because it's the preferred way of including C headers in C++ code).

Source/ui_unix/PH_HidUnix.h
```diff
@@ -0,0 +1,74 @@
...
+#include "PadHandler.h"
+#include <memory>
+
+class CPH_HidUnix : public CPadHandler
+{
+public:
+							CPH_HidUnix();
+	virtual					~CPH_HidUnix();
+	
+	void					Update(uint8*);
```
jpd002: Function should be marked as `override` since it overrides the function from `CPadHandler`.

Thunder07:
I've reworked the 1st commit, and removed the commented android lines in GSH_OpenGLUnix.cpp,
seperated " linux: added basic controls" into it own commit, since it was divided between "linux: added game select and run options" & "linux: add settings dialog", finally I've applied the changes you mentioned.

jpd002:
Still testing this branch (sorry for the delay). 

About using GLX, I don't think it's necessary since there's built-in support to create OpenGL contexts within Qt. Here what I was using when I was testing Qt a while back: https://gist.github.com/jpd002/b44fed616bef84681c5f37dcca3b1a1c. I think something like that should be used, because the code is simpler and is more portable.

Thunder07:
no rush (sorry if i rushed you in my much earlier posts, since I was waiting to hear from you before adding more things, but since I ended up going ahead and doing them).

just looking at the code, I'd say you're right, in my own tests I might have over-thought the process (passed qt context to opengl handler and tried yo set it as current context inside opengl handler, which failed), thus I came up with this.

I'll have a look at this within the next few minutes and update this PR accordingly.

jpd002:
Thanks for the updated branch. Is it normal that a new window spawns when the emulator starts? Maybe I did something wrong, but I thought the emulator's video output would be embedded in the main window.

Thunder07:
is it not embedded? I revised the commit a little with little net effect.
creating a QWindow always spawns a new window, which then is embedded into the main window as a widget, as suck you see the output window at start.
as far as i can tell, their isnt a way to embed a QWindow by default, so this seems to be the only way.

I'll keep trying, looking for a better way around it.

jpd002:
Thanks for the updates. Could you create a new PR with the `qt_ui_jpd` branch or update this one to include these new commits? There's still some things I'd like to comment on and it's a bit hard to track with a different branch.

Thunder07:
I've updated this PR.

jpd002:
Awesome, thanks! :+1:

Source/ui_unix/openglwindow.cpp
```diff
@@ -0,0 +1,34 @@
+#include "openglwindow.h"
+#include <QResizeEvent>
+
+OpenGLWindow::OpenGLWindow(QWindow *parent)
+    : QWindow(parent)
+{
+    QSurfaceFormat format;
+    format.setVersion(3, 2);
+    format.setProfile(QSurfaceFormat::CompatibilityProfile);
```
jpd002: Should use `CoreProfile` here. We want to be as close as possible to OpenGL ES and prevent usage of deprecated OpenGL functions.

jpd002: Humm, just tested further and it might not be as trivial as changing this to `CoreProfile`. I've tried removing the `glewExperimental` line (which shouldn't be there), but it crashes a bit further down the road trying to call `glClearDepthf`. From what I can see, it's only supported on OpenGL 4.1+ which might explain the crash. I'll see what's the best course of action for that.

jpd002: Ok, apparently, GLEW doesn't work well with core profiles. Details here: https://www.opengl.org/wiki/OpenGL_Loading_Library#GLEW_.28OpenGL_Extension_Wrangler.29. I need to do some further testing on Windows and OSX to see what would be the best way to handle this.

Source/ui_unix/mainwindow.cpp
```diff
@@ -0,0 +1,228 @@
...
+#include "global.h"
+#include "GSH_OpenGLQt.h"
+#include "tools/PsfPlayer/Source/SH_OpenAL.h"
+
+#include "ui_mainwindow.h"
+
+#define PREFERENCE_UI_SHOWFPS "ui.showfps"
+#define PREFERENCE_UI_SHOWVIRTUALPAD "ui.showvirtualpad"
+#define PREFERENCE_AUDIO_ENABLEOUTPUT "audio.enableoutput"
+#define PREF_CGSH_OPENGL_ENABLEHIGHRESMODE        "renderer.opengl.enablehighresmode"
```
jpd002: OpenGL GS handler related preference strings are not used anywhere in this file. They can be removed.

Source/ui_unix/settingsdialog.cpp
```diff
@@ -0,0 +1,88 @@
+#include "settingsdialog.h"
+#include "ui_settingsdialog.h"
+
+#define PREFERENCE_UI_SHOWFPS "ui.showfps"
+#define PREFERENCE_UI_SHOWVIRTUALPAD "ui.showvirtualpad"
+#define PREFERENCE_AUDIO_ENABLEOUTPUT "audio.enableoutput"
+#define PREF_CGSH_OPENGL_ENABLEHIGHRESMODE        "renderer.opengl.enablehighresmode"
```
jpd002: Preference defines for the OpenGL GS handler and the base GL handler will be available if you include `../gs/GSH_OpenGL/GSH_OpenGL.h". No need to redefine them here.

Source/ui_unix/settingsdialog.cpp
```diff
@@ -0,0 +1,88 @@
+#include "settingsdialog.h"
+#include "ui_settingsdialog.h"
+
+#define PREFERENCE_UI_SHOWFPS "ui.showfps"
```
jpd002: The preference defines specific to this port should probably be moved to a common header since they are also used in `mainwindow.cpp`. Not sure we need a "showfps" setting, it should probably be always on. The "showvirtualpad" preference should also be removed since there's no virtual pad on Linux.

Source/ui_unix/openglwindow.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <QtGui/QWindow>
+
+class OpenGLWindow : public QWindow
+{
+    Q_OBJECT
+public:
+    explicit OpenGLWindow(QWindow *parent = 0);
+        ~OpenGLWindow();
+    void exposeEvent(QExposeEvent *ev) override;
```
jpd002: Should use `Q_DECL_OVERRIDE` for consistency with the ones below.

Source/ui_unix/mainwindow.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+void MainWindow::initEmu(){
+    g_virtualMachine = new CPS2VM();
+    g_virtualMachine->Initialize();
+
+    g_virtualMachine->CreateGSHandler(CGSH_OpenGLQt::GetFactoryFunction(openglpanel));
+    SetupSoundHandler();
+
+    g_virtualMachine->CreatePadHandler(CPH_HidUnix::GetFactoryFunction());
+    padhandler = (CPH_HidUnix*)g_virtualMachine->GetPadHandler();
```
jpd002: Please use `static_cast` here. C style casts are dangerous.

Source/ui_unix/global.h
```diff
@@ -0,0 +1,10 @@
+#ifndef GLOBAL_H
+#define GLOBAL_H
+
+#include "PS2VM.h"
+#include <QTimer>
+
+CPS2VM* g_virtualMachine = nullptr;
+QTimer *fpstimer = nullptr;
```
jpd002: Do we need to declare the timer as a global variable here? Would probably be nicer to have it as a member of `MainWindow` to reduce its scope.

Source/ui_unix/mainwindow.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+    StatsManager = new CStatsManager();
+    g_virtualMachine->m_ee->m_gs->OnNewFrame.connect(boost::bind(&CStatsManager::OnNewFrame, StatsManager, _1));
+
+    createStatusBar();
+
+    fpstimer = new QTimer(this);
```
jpd002: Timer is created but doesn't seem to be deleted anywhere. Probably needs to be deleted in the destructor.

Source/ui_unix/mainwindow.cpp
```diff
@@ -0,0 +1,228 @@
...
+            g_virtualMachine->m_ee->m_gs->Flip();
+        }
+}
+
+void MainWindow::on_actionOpen_Game_triggered()
+{
+    QFileDialog dialog(this);
+    dialog.setFileMode(QFileDialog::ExistingFile);
+    dialog.setNameFilter(tr("ISO/CSO/ISZ files (*.iso *cso *.isz)"));
+    QString fileName;
```
jpd002: `fileName` should be declared inside the if's scope since it's not used anywhere outside. Should also use `auto` type deduction like this:
`auto fileName = dialog.selectedFiles().first();`

Source/ui_unix/mainwindow.cpp
```diff
@@ -0,0 +1,228 @@
...
+        setOpenGlPanelSize();
+    }
+}
+
+void MainWindow::on_actionBoot_ELF_triggered()
+{
+    QFileDialog dialog(this);
+    dialog.setFileMode(QFileDialog::ExistingFile);
+    dialog.setNameFilter(tr("ELF files (*.elf)"));
+            QString fileName;
```
jpd002: Same comment as above.

Source/ui_unix/mainwindow.h
```diff
@@ -0,0 +1,64 @@
...
+}
+
+class MainWindow : public QMainWindow
+{
+    Q_OBJECT
+
+public:
+    explicit MainWindow(QWidget *parent = 0);
+    ~MainWindow();
+    WId getOpenGLPanel();
```
jpd002: `getOpenGLPanel` function is unused. Should be removed.

Source/ui_unix/mainwindow.cpp
Thunder07: @jpd002 ya, Qt has a garbage collector ([1](http://silmor.de/qtstuff.garbcol.php), [2](url)) that takes care of this.... I recall reading that on their actual site, but i can't find that link right now.

Thunder07:
I've revised the commits based on your suggestions.
the only notable change is that I've also revised "linux: clean up" and changed all variables and functions that don’t need to be public (which is all of them) to private, and as you suggested the fps counter is always on now, so i merged the setupfpscounter into createStatusBar() as for everything else is as you suggested with no other changes.

Source/ui_unix/openglwindow.cpp
jpd002: Alright, after testing, I think the best way to go is to create the core context and keep the `glewExperimental` assignation. I've made changes to the way context creation is handled on Win32 to match that: 142513d877a1133b6dd674325aac5742a02daba9.

Thunder07: @jpd002 I'll append the commit in few minutes, just as my laptop reboots into ubuntu.

Source/ui_unix/mainwindow.cpp
```diff
@@ -0,0 +1,198 @@
...
+#include "ui_mainwindow.h"
+
+MainWindow::MainWindow(QWidget *parent) :
+    QMainWindow(parent),
+    ui(new Ui::MainWindow)
+{
+    ui->setupUi(this);
+
+    openglpanel = new OpenGLWindow;
+    openglpanel->show();
```
jpd002: That line shouldn't be here. Makes the output window separate from the main window. (Didn't we remove that line before?)

build_unix/ui_unix.pro
```diff
@@ -0,0 +1,72 @@
...
+
+unix:!macx: LIBS += -L$$PWD/../../../boost_1_61_0/stage/lib/ -lboost_system -lboost_filesystem -lboost_chrono
+
+DEPENDPATH += $$PWD/../../../boost_1_61_0
+
+unix:!macx: PRE_TARGETDEPS += $$PWD/../../../boost_1_61_0/stage/lib/libboost_system.a \
+                              $$PWD/../../../boost_1_61_0/stage/lib/libboost_filesystem.a \
+                              $$PWD/../../../boost_1_61_0/stage/lib/libboost_chrono.a
+
+LIBS += -lX11 -lGLEW -lz -lbz2 -lopenal
```
jpd002: We don't need to link against `libX11` anymore. It should be removed.

Source/ui_unix/mainwindow.cpp
Thunder07: there was a conflict merger, when i merged it back in, ill have it removed.

Thunder07:
anything else?

jpd002:
Thanks. Well, I think we're getting pretty close to something I would want to merge, but there's still a few issues:

- Virtual Machine scope and life time. Currently, it's not stopped or destroyed before the window dies, so, the process will crash randomly at the end of its execution. Also, it's a global variable inside `global.h` but it's only used by `MainWindow` at the moment. It's not necessarily a bad thing if you were planning to have other windows access virtual machine's state, so, it's just nitpicking. (also, just saw that you are declaring the `g_virtualMachine` variable inside an header file which won't work if it's included in two separate cpp files. The variable declaration needs to be inside a separate cpp file. You can take a look at `Globals.cpp`/`Globals.h` from `ui_macosx` for an example.)

- We need to use two separate menu items to launch a disc image. It should be possible to launch a disc image using a single `Boot Disk Image...` menu item. The `Boot ELF` menu item should also have an ellipsis to indicate that it's going to open another dialog.

- Still not sure about that boost thing, but I need to ponder this a bit more.

I don't know how you feel about all those comments, but tell me if you think it's going too far. I mean, I appreciate the work you put in this, but I don't want to make you feel like I'm trying to hold back on merging the branch in master. I just want to make sure we get the best quality we can get for a port to a new platform.

Thunder07:
I don’t mind at all, I get where you're coming from and its much as my wish as it's yours, and I'd do the same if I'm in your shoes.

- Virtual Machine:
so I should call 
```
g_virtualMachine->Pause();
g_virtualMachine->DestroyPadHandler();
g_virtualMachine->DestroyGSHandler();
g_virtualMachine->DestroySoundHandler();
g_virtualMachine->Destroy();
```
in the mainwindows deconstructor and can probably remove the pause and destroy line from `on_actionExit_triggered()` and just leave exit(0);

ya for the global.h, i looked at the osx version before creating it, but failed to realise it also has a cpp counter part and the reasoning behind it, well at the moment, I dont have a reason for including it in global.h/cpp beside perhaps a logger window in the future, assuming it will need access to virtualmachines to get the logs.

- ok, i get the ellipsis indicator, but im not sure what you want me to do with the boot disk option, do you want me to add a submenu with options selected disc, boot disk?

- your call really.

jpd002:
- VM life time: The sequence is about right, it's just missing a `delete g_virtualMachine` and a `g_virtualMachine = nullptr` at the end. I think it would be better to have that code in the destructor because I was using the `X` button on the window to close it and it never goes through `on_actionExit_triggered`. I think `on_actionExit_triggered` should just post a "close window" message and let the process tear itself down normally.

- Boot Disc Menu: What I mean is that there should be a single `Boot Disc Image...` menu item that would replace the current `Open Game` and `Start Game` menu items and that would let the user select a disc image file and launch it automatically.

Thunder07:
Note: with the last commit "linux: create save state folder", I've briefly tried save state on osx but failed to get it to load, I'd assume its because the save state was not created, as Play tried to create them in the binary directory, which it probably had no permission to write to.

jpd002:
I reckon saved states should be stored in "~/Play Data Files/states/" on all platforms. There's [some code](https://github.com/jpd002/Play-/blob/master/Source/ui_win32/MainWindow.cpp#L910) to handle that on Win32 that I might need to move in the `Source` folder so all platforms can use it. But for now, I wouldn't have objections to copy/paste that piece of code unto to other platforms' UI code.

Thunder07:
for commit "linux: added memory card manager", you'd need icu library `apt-get install libicu-dev`, I've reflected 1st post to include this.
also, for the manager, although, I  could get opengl icon code to compile, i couldnt get the icons to in the opengl window, so i left that section out.

for "linux: reset ROM option", I looked at win32 implementation, my way seemed simpler, so I wouldnt mind your opinion on that.

Thunder07:
Hi @jpd002,
quick note in regards to "WIP: linux: configurable controls" commits, I've been working on these commits for sometime now, and it took me long enough to get through it, mainly because I wasn’t too sure how to go about the whole thing, but this is what I ended up with, when you review it and if you felt it wasn’t worth adding, just say so and I can remove them, if you where ok with the final result or want specific changes, let me know so I can merge those commits into one and see what I can do about it, I left them as Parts, mainly so you can see the different methods I tried to get to the final one. the main gist of it was, using a custom QWidget was cleanest way to implement key press detection while coupling it with a Qlabel with the only downside that it appears as an empty widget in Qt Creator.

secondary, while this [commit](https://github.com/Thunder07/Play-/commit/d2346dc393a7b5475cff8941d96f415244b7a4b3) isn't Linux UI (as the name of this PR suggests), its still linux related, without it KingHearts crashes right before the intro screen, while with it, you get the reported green screen intro, shall I add it to this PR or a separate one?(or you could just cherry-pick it)

lastly, I believe i managed to get everything cover in this PR beside FrameDebugger, as such I'll be waiting for your review, if i missed anything let me know.
I did however have a really quick look at FrameDebugger code and found few references to D9, so is it even portable to linux in that state? with that said, I've been able to build the unix build with debug options (`cmake -DCMAKE_BUILD_TYPE=Debug ..`) and set breakpoint thoughout the code, even inside Play's library and they do get triggered.

jpd002:
Sorry I didn't give this PR much attention. I've had an hard time getting time to work on this steadily.

Everything looking quite promising. You can keep d2346dc393a7b5475cff8941d96f415244b7a4b3 in there. As for boost, I think it would be best to use the system's boost installation. We just need the `filesystem` library really (which in turn requires `system`) and then, that library is coming in C++17, so, it's just a matter of time before it's removed completely as a dependency of this project.

Thunder07:
don't worry about it. 
and ill get on them once i get a chance.

note to self: memory card manager, long name warp issue. (i keep intermittently remembering/forgetting this)

Thunder07:
Quick question, how come you've kept `Posix_VolumeStream.h` and `VolumeStream.h` separate?
the slight differences can be managed with #ifdef.
I ask since I was checking to see if i can get this branch working on Win32/OSX (OSX got as far as booting games, Win32 compiled library and waiting for Qt to download), for OSX the changes are few lines and for windows so far the only change I had to make to the CMakeList.txt was replace `Posix_VolumeStream.h` with `VolumeStream.h`.

jpd002:
`VolumeStream.h` and `Posix_VolumeStream.h` are separate because I didn't want to litter code with `#ifdef`s. The required `VolumeStream` class can easily be instantiated by platform specific code, hence, removing the need for any `#ifdef`. 

But, yeah, since you're trying to build a single code base for all platforms, I kinda see why you're having that problem. I think it'll be hard to get the Qt UI code to work properly on all platforms without "hacking" stuff in there (especially on Windows) and I don't know if we should put the effort on getting it to work everywhere. If we limit the scope of the Qt UI to *nix based platforms (I might consider excluding OSX out of that), I think we could get a clean UI code base that doesn't have too much platform conditional code.

jpd002:
Oh yeah, also, just so you know, since I'm finally getting some free time, I'm planning to get that branch merged by September 2nd. Many people have been requesting the Linux port and I think it's time to make it official.

Thunder07:
the `VolumeStream` isn't really an issue, if you don't want to `#ifdef` the code, you can still add OS specific changes through cmake, it's just working through this PR i noticed a lot of the time, I didn't need to use `#ifdef` myself, since that was already layed out in Framework.

As you mentioned, Windows did cause me the most pain to get it to work, but that mostly, because, I removed window from all the PC at home, 1.5 years ago and haven't touched it since.
it did work in the end, and the biggest issue was solving dependencies(lib/bzip2/boost), but that was mostly due to they way window works, no standard directory for source/header like osx/unix systems.

in the end, I thought the required changes were minimal to get it working on OSX/Windows, I'll do a quick clean up and push those commit for you to decided, if its worth it or not.


Glad to hear that, I've been waiting for sometime to be able to say, "Play! is the 1st PS2 emulator to work on all major Desktop/Mobile Platforms natively, that is Windows, OSX, Ubuntu (and other unix based systems), Android and iOS".
I'm als not back to college until 6th Sept, so I should be free most of the times to make any chances you'd need.

jpd002:
You can go ahead and push them, I've just started looking at the code.

Also, could you create a PR for this commit: https://github.com/Thunder07/Play--Framework/commit/e0a54c53ad34467a648088b293d9bcd64b66b49f? It's required to build this. Thanks!

Thunder07:
Created https://github.com/jpd002/Play--Framework/pull/8 and updated this branch, let me know if you need anything.

Source/ui_unix/cdevice.h
```diff
@@ -0,0 +1,61 @@
...
+};
+
+
+
+class CDirectoryDevice : public CDevice
+{
+public:
+    CDirectoryDevice(const char*, const char*);
+    virtual						~CDirectoryDevice();
+    virtual const char*			GetDeviceName();
```
jpd002: Please use `override` for all overridden virtual functions.

Source/ui_unix/cdevice.cpp
```diff
@@ -0,0 +1,164 @@
+#include "cdevice.h"
```
jpd002: Not quite sure about that file's name. I think something like `VfsDevice.cpp/.h` would be more fitting.

Source/ui_unix/cdevice.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+}
+
+bool CCdrom0Device::RequestModification(QWidget *parent)
+{
+    if (true)
+    {
+
+        VFSDiscSelectorDialog* vfsds = new VFSDiscSelectorDialog(m_sImagePath, m_nBindingType, parent);
+        VFSDiscSelectorDialog::connect(vfsds, &VFSDiscSelectorDialog::onFinish, std::bind(&CCdrom0Device::callBack, this, std::placeholders::_1, std::placeholders::_2));
```
jpd002: Not sure if it's possible, but using a lambda here instead of `std::bind` would be nicer.

Source/ui_unix/cdevice.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+void CCdrom0Device::Save()
+{
+    CAppConfig::GetInstance().SetPreferenceString(PS2VM_CDROM0PATH, m_sImagePath.c_str());
+
+}
+
+bool CCdrom0Device::RequestModification(QWidget *parent)
+{
+    if (true)
```
jpd002: Why the always true condition?

Source/ui_unix/cdevice.cpp
```diff
@@ -0,0 +1,164 @@
...
+    {
+
+        VFSDiscSelectorDialog* vfsds = new VFSDiscSelectorDialog(m_sImagePath, m_nBindingType, parent);
+        VFSDiscSelectorDialog::connect(vfsds, &VFSDiscSelectorDialog::onFinish, std::bind(&CCdrom0Device::callBack, this, std::placeholders::_1, std::placeholders::_2));
+        vfsds->exec();
```
jpd002: Does `vfsds` needs to be deleted after we're done with it?

Source/ui_unix/cdevice.cpp
Thunder07: I use that when trying different codes, easy way to enable/disable the code while also isolating its variables. it seems i just forgot to remove it.

Source/ui_unix/cdevice.cpp
Thunder07: @jpd002 Qt should take care of it, since the dialog has a parent widget, i still went ahead and deleted it anyway.

build_unix/build.sh
```diff
@@ -4,3 +4,8 @@ cd build
 cmake ..
 make
 cd ..
+mkdir -p build-ui/
+~/Qt5.6.*/5.6*/gcc*/bin/qmake -o build-ui/
```
jpd002: Why isn't this only `qmake`? If you install Qt, `qmake` should be available as a global command.

Thunder07: initially, because wanted to use qt 5.6, so that it would include this [patch](https://codereview.qt-project.org/#/c/139775/13/src/widgets/kernel/qopenglwidget.cpp), but i don't think thats needed anymore. Ill confirm that once i get back from college tonight.

Source/ui_unix/vfsmodel.cpp
```diff
@@ -0,0 +1,96 @@
...
+        return 0;
+    }
+}
+
+QVariant VFSModel::data(const QModelIndex &index, int role) const
+{
+    if (role == Qt::DisplayRole)
+    {
+        CDevice* m_device = m_devices.at(index.row());
+        const char* val;
```
jpd002: This should be initialized (probably to an empty string). Just trying to avoid UB.

Source/ui_unix/vfsmodel.cpp
```diff
@@ -0,0 +1,96 @@
...
+            break;
+        }
+        return QVariant(val);
+    }
+    return QVariant();
+}
+
+void VFSModel::SetupDevices()
+{
+    m_devices[0] = new CDirectoryDevice("mc0",  "ps2.mc0.directory");
```
jpd002: Doesn't seem to be deleted anywhere. Would probably better to use `shared_ptr` or `unique_ptr` here.

Source/ui_unix/vfsmodel.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+void VFSModel::DoubleClicked(const QModelIndex &index, QWidget* parent)
+{
+     CDevice* m_device = m_devices.at(index.row());
+     m_device->RequestModification(parent);
+}
+
+void VFSModel::Save()
+{
+    for(CDevice::DeviceList::const_iterator deviceIterator(m_devices.begin());
```
jpd002: Should use range-based `for` loop.

Source/ui_unix/controllerconfigdialog.cpp
```diff
@@ -0,0 +1,329 @@
...
+    {
+        if(Rxml.isStartElement())
+        {
+            if(Rxml.name() == "KeyMaps")
+            {
+                Rxml.readNext();
+            }
+            else if(Rxml.name() == "Key")
+            {
+                PadQWidgetExt* child;
```
jpd002: Should be initialized to `nullptr`.

jpd002:
Ok, I think it's pretty good as it is right now, I'm going to merge this tomorrow (giving you some time to do the minor changes listed above). There's still some things that I think should be improved for future versions:

- `VfsDdevice.cpp`/`VfsDdevice.h` should be renamed to `VfsDevice.cpp`/`VfsDevice.h`
- Figure out how to remove some of the code duplicated from other platforms (`StatsManager`, `MemoryCard`, etc.)
- Consistency on indenting and variable naming (my fault because I have not clearly defined the formatting rules for the project)
- Settings for controls should be stored in `config.xml` instead of using a different XML file.

Thunder07:
Ok, So I've done the little tweaks you asked for, except for the shared_ptr, since I don’t know how to use them. I did delete the devices though.
I've merged the `configure controls` commit into one, and correctly renamed VfsDevice this time :)

I dont think managing the duplicate code would be an issue, I'd try to sort out something by the end of the weekend.
and I'll have a thought on how to save the controls to the `config.xml`, but as of right now, I'm drawing blanks, since we'd need a way to load and save the data without specifying the name of every single button (aka some kind of loop, I know you probably have something for windows, so ill look at it)

I also want to go on record and say, I went against my better nature and did not name the 1st commit, `linux: from small beginnings`.

jpd002:
Thanks again for this! It's greatly appreciated.

Don't worry too much about the code duplication, it's a larger problem that spans all platforms. For example, there's the game database querying code (on Android and iOS) that I would like to bring in a common place at some point. I'd like all platforms to have a kind of common feature set, but I still need to write a document describing that.

Thunder07:
Glad i could be of help.

quick note, I've noticed as of this commit https://github.com/jpd002/Play--CodeGen/commit/bbf9543e752cefe3f88421035a4b4d1eef4ccfbc Play! would crash on my desktop, both on ubuntu/windows, since my processor doesn't support SSSE3 instruction set (my laptop works fine and its the one I use). just to keep it in mind should there be an increased incident of "I can't run any games" with this release, since people with older machine running ubuntu might try it.

even though this was a linux PR, I couldn't find a clear cut way to check for cpu features, thats still multiplatform, so i delayed adding that check.

jpd002:
Oh yeah, good point about `pshufb`. I guess we'll have to keep the old `Md_IsZero` and `Md_IsNegative` code around for those who don't have SSSE3 compatible processors.

PR #499, opened 2016-12-07
Xcode fixes.

MaddTheSane:
Have the Xcode project refer to files by relation to groups.
Remove Xcode config files from copy phase: they aren't used there.


jpd002:
Thanks for the fixes!

PR #500, opened 2016-12-08
Madd's Mac Patches

MaddTheSane:
This patches quiet warnings when compiled on Xcode 8.1.
It also fixes the Objective-C initialization methods to better match current best practices.


SolarstrikeX2:
Good idea.

jpd002:
Thanks again for this!

Fadi5555:
@jpd002 please can you give some love for iOS port? It's been long time about that.
Even we still not have save/ load feature. As well as mapping start and select buttons for external controllers like ps3/4

Anyway welcome back we really need you a lot to stay and continue working on this amazing project ever.

PR #526, opened 2017-02-19
Linux: GameID in status bar

Thunder07:
I just noticed that on both Win32/OSX the Game ID is displayed on the title bar, while on this PR I have it set to status bar, if you've issue with that I can have it changed.


jpd002:
Thanks for this.

PR #527, opened 2017-02-19
Screenshot Feature

Thunder07:
This PR also requires [This](https://github.com/jpd002/Play--Framework/pull/11) PR in Framework repo
as such it won't build correctly until that's also merged, so i provided a build log below

[Appveyor](https://ci.appveyor.com/project/Thunder07/play/build/97) & [Travis](https://travis-ci.org/Thunder07/Play-/builds/203220817) builds

Edit: on ubuntu, the screenshot would be saved in `/home/__USER/.local/share/Play..../screenshots/*` which isn't ideal as `.local` is a hidden directory.

but on both OSX/Win32 they're saved into their equivalent of `~/Documents/Play.../screenshots/`

Edit2: ~~just found `MarkNewFrame`, I'll need to test it instead of the signal I added~~

### DIFF
--- a/Source/ScreenShotUtils.cpp
+++ b/Source/ScreenShotUtils.cpp
@@ -1,0 +1,56 @@
+#include "AppConfig.h"
+#include "bitmap/BMP.h"
+#include "ScreenShotUtils.h"
+#include "StdStream.h"
+#include "PathUtils.h"
+
+#include <chrono>
+#include <ctime>
+#include <iomanip>
+#include <sstream>
+
+void CScreenShotUtils::TriggerGetScreenshot(CPS2VM* virtualMachine, Callback completionCallback)
+{
+	virtualMachine->m_ee->m_gs->OnFlipComplete.connect_extended(
+		[=](const boost::signals2::connection &c)->void
+		{
+			c.disconnect();
+			try
+			{
+				auto m_buffer = virtualMachine->m_ee->m_gs->GetScreenshot();
+				auto name = GenerateScreenShotPath(virtualMachine->m_ee->m_os->GetExecutableName());
+				Framework::CStdStream outputStream(name.string().c_str(), "wb");
+				Framework::CBMP::WriteBitmap(m_buffer, outputStream);
+
+				std::ostringstream oss;
+				oss << "Screenshot saved as " << name.filename();
+				auto msgstr = oss.str();
+				completionCallback(0, msgstr.c_str());
+			}
+			catch (const std::exception& e)
+			{
+				completionCallback(-1, "Error occured while trying to save screenshot.");
+			}
+		}
+	);
+}
+
+Framework::CConfig::PathType CScreenShotUtils::GetScreenShotDirectoryPath()
+{
+	auto screenshotpath(CAppConfig::GetBasePath() / boost::filesystem::path("screenshots"));
+	Framework::PathUtils::EnsurePathExists(screenshotpath);
+	return screenshotpath;
+}
+
+Framework::CConfig::PathType CScreenShotUtils::GenerateScreenShotPath(const char* gameID)
+{
+	auto t = std::time(nullptr);
+	auto tm = *std::localtime(&t);
+	auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now().time_since_epoch()).count() % 1000;
+
+	std::ostringstream oss;
+	oss << gameID << std::put_time(&tm, "_%d-%m-%Y_%H.%M.%S.") << ms << ".bmp";
+	auto screenshotFileName = oss.str();
+
+	return GetScreenShotDirectoryPath() / boost::filesystem::path(screenshotFileName);
+}
--- a/Source/ScreenShotUtils.h
+++ b/Source/ScreenShotUtils.h
@@ -1,0 +1,15 @@
+#pragma once
+#include "Config.h"
+#include "PS2VM.h"
+
+class CScreenShotUtils : public Framework::CConfig
+{
+public:
+	typedef std::function<void(int, const char*)>		Callback;
+
+	static void								TriggerGetScreenshot(CPS2VM*, Callback);
+
+private:
+	static CConfig::PathType				GetScreenShotDirectoryPath();
+	static CConfig::PathType				GenerateScreenShotPath(const char* gameID);
+};
--- a/Source/ui_macosx/ApplicationDelegate.h
+++ b/Source/ui_macosx/ApplicationDelegate.h
@@ -7,10 +7,12 @@
 	IBOutlet NSMenuItem*			pauseResumeMenuItem;
 	IBOutlet NSMenuItem*			loadStateMenuItem;
 	IBOutlet NSMenuItem*			saveStateMenuItem;
+	IBOutlet NSMenuItem*			screenCaptureMenuItem;
 }
 
 -(void)applicationDidFinishLaunching: (NSNotification*)notification;
 -(IBAction)showPreferences: (id)sender;
+-(IBAction)screenCaptureMenuSelected: (id)sender;
 -(IBAction)bootElfMenuSelected: (id)sender;
 -(IBAction)bootCdrom0MenuSelected: (id)sender;
 -(IBAction)bootDiskImageMenuSelected: (id)sender;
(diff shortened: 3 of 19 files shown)


Thunder07:
I noticed that in `Add screencapture method for OpenGL` i used the prefix `m_`, just recommitted it (twice :/ ) to fix that.

Source/ui_macosx/ApplicationDelegate.mm
```diff
@@ -100,6 +103,28 @@ -(IBAction)showPreferences: (id)sender
...
+		{
+			c.disconnect();
+			@try
+			{
+				auto m_buffer = static_cast<CGSH_OpenGL*>(g_virtualMachine->m_ee->m_gs)->GetScreenshot();
+				auto name = CAppConfig::GenerateScreenShotPath(g_virtualMachine->m_ee->m_os->GetExecutableName());
+				Framework::CStdStream outputStream(name.string().c_str(), "wb");
+				Framework::CBMP::WriteBitmap(m_buffer, outputStream);
+			}
+			@catch (NSException *exception)
```
jpd002: I'm not sure, but I think this won't catch a C++ exception.

Source/ui_unix/mainwindow.cpp
```diff
@@ -448,3 +451,27 @@ void MainWindow::on_actionController_Manager_triggered()
...
+            try
+            {
+                auto m_buffer = static_cast<CGSH_OpenGL*>(g_virtualMachine->m_ee->m_gs)->GetScreenshot();
+                auto name = CAppConfig::GenerateScreenShotPath(g_virtualMachine->m_ee->m_os->GetExecutableName());
+                Framework::CStdStream outputStream(name.string().c_str(), "wb");
+                Framework::CBMP::WriteBitmap(m_buffer, outputStream);
+            }
+            catch (const std::exception& e)
+            {
+                QMessageBox messageBox;
```
jpd002: Instead of displaying a message box, I think it would be more interesting to write a message in the status bar.

Source/ui_win32/MainWindow.cpp
```diff
@@ -1042,3 +1047,24 @@ void CMainWindow::CLoadElfOpenCommand::Execute(CMainWindow* mainWindow)
...
+			try
+			{
+				auto m_buffer = static_cast<CGSH_OpenGL*>(m_virtualMachine.m_ee->m_gs)->GetScreenshot();
+				auto name = CAppConfig::GenerateScreenShotPath(m_virtualMachine.m_ee->m_os->GetExecutableName());
+				Framework::CStdStream outputStream(name.string().c_str(), "wb");
+				Framework::CBMP::WriteBitmap(m_buffer, outputStream);
+			}
+			catch (std::exception& e)
+			{
+				MessageBox(m_hWnd, _T("Error: failed to save screenshot."), APP_NAME, 16);
```
jpd002: Same as Qt version. Modal dialog boxes can be annoying.

Source/ui_macosx/ApplicationDelegate.mm
```diff
@@ -100,6 +103,28 @@ -(IBAction)showPreferences: (id)sender
 	[self setupSoundHandler];
 }
 
+-(IBAction)screenCaptureMenuSelected: (id)sender
+{
+	static_cast<CGSH_OpenGL*>(g_virtualMachine->m_ee->m_gs)->OnFlipImplComplete.connect_extended(
```
Thunder07: so are you suggesting to create a virtual function in `GSHandler`? (as I'm not sure how generic the current method can get with `glReadPixels()` being OpenGL specific)

jpd002: Yeah, we could have a virtual function returning a `Framework::CBitmap` that's overridden in every specific GS handler, with the OpenGL implementation calling `glGetPixels`, etc.

Thunder07:
noticed inconsistent indentation in Win32/OSX code, and I've just pushed changes to fix that.

Source/ui_win32/Res.rc
```diff
@@ -157,6 +169,8 @@ BEGIN
         MENUITEM "State Slot",                  ID_MAIN_VM_STATESLOT
         MENUITEM "Save State\tF7",              ID_MAIN_VM_SAVESTATE
         MENUITEM "Load State\tF8",              ID_MAIN_VM_LOADSTATE
+        MENUITEM SEPARATOR
```
jpd002: These two lines should be the only changes in this file. All other changes to `Res.rc` are unnecessary and should be removed.

Source/ScreenShotUtils.h
```diff
@@ -0,0 +1,16 @@
...
+class CScreenShotUtils : public Framework::CConfig
+{
+public:
+    typedef std::function<void(int, const char*)>        Callback;
+
+    static void                             TriggerGetSceenshot(CPS2VM*, Callback);
+
+private:
+    static CConfig::PathType                GetScreenShotDirectoryPath();
+    static CConfig::PathType                GetGameScreenShotDirectoryPath(std::string gameID);
```
jpd002: Is there any need to sort screen shots in separate directories for every game? I think it should be sufficient to prepend the game's id in the screen shot's file name.

jpd002:
Submitted another review, mostly minor stuff. It should be good to merge after that.

Source/ui_win32/Res.rc
Thunder07: @jpd002 These changed were done by the IDE itself, do you want me to reverse them?

Source/ui_macosx/ApplicationDelegate.mm
```diff
@@ -5,11 +5,13 @@
 #import "../../tools/PsfPlayer/Source/SH_OpenAL.h"
 #import "Globals.h"
 #import "PreferenceDefs.h"
+#include "ScreenShotUtils.h"
 #import "../ee/PS2OS.h"
 #import "../ScopedVmPauser.h"
 #import "../PS2VM_Preferences.h"
 #import "../AppConfig.h"
 
+
```
Thunder07: There probably was an include that I didn't delete probably, I'll have this fixed.

Source/ui_macosx/ApplicationDelegate.h
```diff
@@ -24,5 +26,6 @@
 -(void)bootFromElf: (NSString*)fileName;
 -(void)bootFromCdrom0;
 -(void)updateTitle;
+-(void)OnNewFrame;
```
Thunder07: Yup, it'll be removed.

Source/ScreenShotUtils.h
Thunder07: It's mostly for cosmetics and from organisational point of view, without folders can get messy really quick, (I know this isn't PS4, but on my PS4, I'd take atleast 100 screenshot per game, I'd imagine that'd be alot of images to scroll through), so I'll leave this choice to you.

Source/ui_win32/Res.rc
jpd002: Yeah, VS has a bad habit of changing stuff in resource files when the users locale config don't match. It's better to filter that noise out whenever possible.

Source/ScreenShotUtils.h
jpd002: I prefer not having separate folders for every game.

Thunder07:
I also noticed i was using `m_buffer` as variable name in `CGSH_OpenGL::GetScreenshot` method, so that's fixed. and so are the other requests.

Source/ScreenShotUtils.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "ScreenShotUtils.h"
+#include "StdStream.h"
+#include "PathUtils.h"
+
+#include <chrono>
+#include <ctime>
+#include <iomanip>
+#include <sstream>
+
+void CScreenShotUtils::TriggerGetSceenshot(CPS2VM* virtualMachine, Callback completionCallback)
```
jpd002: Shouldn't that be `TriggerGetScreenshot` instead of `TriggerGetSceenshot`?

jpd002:
There's just a typo that needs to be fixed. I'll merge this PR once it's done.

Source/ScreenShotUtils.cpp
Thunder07: sorry for the silly mistake, Fixed.

Thunder07:
Hi, you probably spotted the Android build failure due to failed signing. perhaps that should be moved to before deploy, as that not triggered by PRs.

jpd002:
Indeed. I'll fix that, sorry.

Thunder07:
Thanks for merging

PR #530, opened 2017-02-21
Cleanup

Thunder07:
Replace WIN32 preprocessor with _WIN32


jpd002:
Thanks for cleaning this up.

PR #531, opened 2017-02-25
Qt update

Thunder07:


Source/ui_unix/mainwindow.cpp
```diff
@@ -309,8 +309,7 @@ void MainWindow::saveState()
     ui->menuLoad_States->actions().at(m_stateSlot-1)->setText(QString("Load Slot %1 - %2").arg(m_stateSlot).arg(datetime));
 }
 
-void MainWindow::loadState(){
-    int m_stateSlot = sender()->property("stateSlot").toInt();
+void MainWindow::loadState(int m_stateSlot){
```
jpd002: Prefixing names with `m_` is reserved for class member variables. This should be `stateSlot`.

Source/ui_unix/mainwindow.cpp
```diff
@@ -278,29 +281,26 @@ void MainWindow::SetupSaveLoadStateSlots()
...
-            connect(loadaction, SIGNAL(triggered()), this, SLOT(loadState()));
+            connect(saveaction, &QAction::triggered, std::bind(&MainWindow::saveState, this, i));
+            connect(loadaction, &QAction::triggered, std::bind(&MainWindow::loadState, this, i));
         }
 
     }
 }
 
-void MainWindow::saveState()
+void MainWindow::saveState(int m_stateSlot)
```
jpd002: Prefixing names with `m_` is reserved for class member variables. This should be `stateSlot`.

jpd002:
Awesome, thanks!

PR #532, opened 2017-02-25
Play Qt OSX/Win32 builds

Thunder07:
if you recall with my initial qt linux client, I tried to get the same qt client to work with OSX/Win32, but OSX had the weird resize issue and win32 crashed on resize.

for OSX it seemed like the issue was with `swapBuffers` and as [Qt](http://doc.qt.io/qt-5/qopenglcontext.html#swapBuffers) states that `makeCurrent()` must be called after `swapBuffers()` but somehow on linux, it worked fine without.
as of this PR osx client work but has 2 issues, (1) on exit the client crashes (crash stack is below)
 (2) the memory card manager can't read games name, so returns `???`, i recalled seeing a [commit](https://github.com/MaddTheSane/Play--Framework/commit/85db58817b412a3fdeb21deca96d5953850b679e) a while back from @MaddTheSane, that fixed this issue.

for Win32, the issue seemed with `CMailBox::SendCall` more specifically with `m_callFinished.wait_for(callLock, std::chrono::milliseconds(100));` as Qt will report resize multiple times until full expanded causing a mutex lock?

```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fffd727cdda __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fffd7368787 pthread_kill + 90
2   libsystem_c.dylib             	0x00007fffd71e2420 abort + 129
3   libc++abi.dylib               	0x00007fffd5d4185a abort_message + 266
4   libc++abi.dylib               	0x00007fffd5d66b72 default_terminate_handler() + 46
5   libc++abi.dylib               	0x00007fffd5d63d69 std::__terminate(void (*)()) + 8
6   libc++abi.dylib               	0x00007fffd5d63df0 std::terminate() + 64
7   com.virtualapplications.Play  	0x000000010fd8862c CEeExecutor::~CEeExecutor() + 28
8   com.virtualapplications.Play  	0x000000010fd8648c Ee::CSubSystem::~CSubSystem() + 252
9   com.virtualapplications.Play  	0x000000010fd8662e Ee::CSubSystem::~CSubSystem() + 14
10  com.virtualapplications.Play  	0x000000010fe2d3b9 CPS2VM::~CPS2VM() + 345
11  com.virtualapplications.Play  	0x000000010fe2d5ce CPS2VM::~CPS2VM() + 14
12  com.virtualapplications.Play  	0x000000010fcdfef1 MainWindow::~MainWindow() + 177
13  com.virtualapplications.Play  	0x000000010fcde1d4 main + 84
14  libdyld.dylib                 	0x00007fffd714e255 start + 1
```
I've tried following the stack for anything obvious with no luck :/

Side Note: I've one more [PR](https://github.com/Thunder07/Play-/tree/psf_build) (PsfPlayer cmake) which I'll open should/n't this get merge as one has to be rebased onto the other, as it'll cause merge conflict and will cause some changes to the `cmakelists.txt` to be lost, after which I'll stop annoying you with PR until summer break :p

Edit: I'll rebase this PR when I'm free. but do you want the Qt ui to build through travis/Appvoyer as well, I currently have it setup this way to make sure the code remains multiplatform going ahead, although the OSX Qt build might be more feature full than Qt already.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -8,6 +8,11 @@ travis_before_install()
         sudo add-apt-repository --yes ppa:ubuntu-toolchain-r/test;
         sudo apt-get update -qq;
         sudo apt-get install -qq qt56base gcc-5 g++-5 cmake libalut-dev;
+    elif [ "$TARGET_OS" = "OSX" ]; then
+        if [ "$BUILD_UI" = "QT" ]; then
+            brew update
+            brew install qt5 || true
+        fi;
     elif [ "$TARGET_OS" = "Android" ]; then
         sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y #is this needed?
         sudo apt-get update -y
@@ -53,7 +58,11 @@ travis_script()
             cmake .. -G"$BUILD_TYPE" -DCMAKE_PREFIX_PATH=/opt/qt56/;
             cmake --build .
         elif [ "$TARGET_OS" = "OSX" ]; then
-            cmake .. -G"$BUILD_TYPE"
+            if [ "$BUILD_UI" = "QT" ]; then
+                cmake .. -G"$BUILD_TYPE" -DCMAKE_PREFIX_PATH=/usr/local/opt/qt5/ -DUSE_QT:BOOL=ON
+            else
+                cmake .. -G"$BUILD_TYPE"
+            fi;
             cmake --build . --config Release
         elif [ "$TARGET_OS" = "IOS" ]; then
             cmake .. -G"$BUILD_TYPE" -DCMAKE_TOOLCHAIN_FILE=../../../Dependencies/cmake-ios/ios.cmake -DTARGET_IOS=ON
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,11 +11,17 @@ matrix:
       env:
         - BUILD_TYPE=Xcode
         - TARGET_OS=OSX
+        - BUILD_UI=NATIVE
     - os: osx
       osx_image: xcode8.3
       env:
         - BUILD_TYPE=Xcode
         - TARGET_OS=IOS
+      osx_image: xcode8.3
+      env:
+        - BUILD_TYPE=Xcode
+        - TARGET_OS=OSX
+        - BUILD_UI=QT
     - os: linux
       dist: precise
       language: android
--- a/Source/MailBox.cpp
+++ b/Source/MailBox.cpp
@@ -1,5 +1,5 @@
 #include "MailBox.h"
-#if defined(_WIN32)
+#if defined(_WIN32) && !defined(_QT)
 #include "win32/Win32Defs.h"
 #endif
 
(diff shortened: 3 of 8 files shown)


zminhquanz:
Qt UI have easier to coding , nice work

Thunder07:
I've tried to rebase this PR onto https://github.com/jpd002/Play-/pull/572 and it proved too much work, more like to much `#if def` here and there. it just looked to hacky to push the changes out, so for the mean while, I'll close this PR.

I'll probably maintain a local build, since I switch between OSX/Ubuntu frequent enough, and I might figure out a way to re unit them again.

jpd002:
Sorry for taking so long to comment on this PR.

So, while I wouldn't mind being able to build the Qt UI for other platforms, I would rather keep the native versions official. Not sure I want them to build on AppVeyor/Travis either, because there's already lots of variations being built right now.

With that being said, I think the problems you are pointing should be addressed. 

I've merged @MaddTheSane's `string_cast_sjis` commit, so, memory card save titles should show up properly now. 

We also need to get rid of the Win32 specific stuff in `CMailBox::SendCall`. It causes lots of problems even on the native Win32 version due to re-entrancy. I've been working on getting rid of some of the synchronous calls (https://github.com/jpd002/Play-/tree/save_state_async), but there's still some reorganization to be done in the way the messages are sent to the UI. 

I don't really know what could cause the crash on exit... Is there a message in the debugger's output log?

Thunder07:
no worries, we've all other things to tend to every now and then.

you mean qt debugger? that was a while back but I don't recall anything specific.

looking at the code a 2nd time I suspect the thread started by this line https://github.com/jpd002/Play-/blob/5ddc73c6ca4d1fc3b4eebc16db7999746a99007b/Source/ee/EeExecutor.cpp#L71 which isn't stopped anywhere is the cause of this issue(it'll be odd still, since it's not crashing in the native client but is when using the qt ui), I should be free this week to test that.

with that said, as I mentioned earlier, maintaining this branch at the moment isn't a priority since I've open the GamePad PR, but I'll push the changes for `makeCurrent()` since it should be the correct implementation.

jpd002:
Ah, makes sense that would be the culprit. Humm, I don't know if it's going to drop messages with the timeout. Also, the thread will still be "stuck" even though there's a timeout, right? I mean, ideally, we would join the thread in EeExecutor's destructor to make sure everything has been cleaned up alright. Another thing to consider is that there's two `mach_msg` calls in there and both have their specific use. I don't think the second one will block for a long time since it's sending back a message to the kernel, but that would need some testing. If you want to test, I know FFX will triggers the code page write-protection.

Thunder07:
oh I should have mentioned in my test I added `std::atomic<bool>` variable which I set to false in `~CEeExecutor()` as to break the loop if we're closing the app(deconstructing `CEeExecutor`) else if it just timeout, it will loop again.

I'll have a look, but as you said I won't imagine the 2nd `mach_msg` will be an issue, and if so, it might actually be tough to isolate for a test.

Thunder07:
Ok, I've tested it break points and FFX seems to trigger `mach_msg` multiple times (5-6) but would immediately finish the call, and since it's a send request as oppose to receive I don't think it's likely to be blocked long enough to be noticeable.

PR #534, opened 2017-02-27
CMake Debugger build

Thunder07:
Current Issue with this build PR, starting a game creates a small/empty non-closable window, restarting the game or a new game causes that window to grow to cover the full screen.... is that an issue you had before? (quick note: lexical_cast wasn't included from boost and noticed one of your earlier commit, so i just followed suit and replaced it)

over the weekend, I've been attempting to setup windows Debugger build as it was setup in visual studio. i.e: `cmake --build . --config Debug/ReleaseWithDebugger` however I've ran into issues, as that would create different build per config, and cmake didn't support that ( there is a work around, but that would involve wrapping the debugger source (aka`DEBUG_SRC` in `CMakeLists.txt`) with `#ifdef debugger_included`).
e.g how that would look during build. (branch [multiconfig](https://github.com/Thunder07/Play-/tree/multiconfig))
```
cmake .. -G"Visual Studio 14 2015"
cmake --build . --config ReleaseWithDebugger
```
the advantage of this that, you can build Release,Debug,ReleaseWithDebugger,DebugWithDebugger and they'll automatically be build into their respective folders.

While for this PR, Debug/Release get their own folder, but Debugger version builds into Debug/Release as well, however what you can do is create `build` and `build_debugger` folders to keep them separate.
which is what I'm proposing in this PR.
```
cmake .. -G"Visual Studio 14 2015" -DDEBUGGER_INCLUDED=on
cmake --build . --config Release
```

edit: spelling + link (branch [multiconfig](https://github.com/Thunder07/Play-/tree/multiconfig))

### DIFF
--- a/Source/ee/PS2OS.cpp
+++ b/Source/ee/PS2OS.cpp
@@ -2,6 +2,7 @@
 #include <stdlib.h>
 #include <exception>
 #include <boost/filesystem/path.hpp>
+#include "string_format.h"
 #include "PS2OS.h"
 #include "StdStream.h"
 #ifdef __ANDROID__
@@ -3185,7 +3186,7 @@ BiosDebugThreadInfoArray CPS2OS::GetThreadsDebugInfo() const
 			threadInfo.stateDescription = "Sleeping";
 			break;
 		case THREAD_WAITING:
-			threadInfo.stateDescription = "Waiting (Semaphore: " + boost::lexical_cast<std::string>(thread->semaWait) + ")";
+			threadInfo.stateDescription = string_format("Waiting (Semaphore: %d)", thread->semaWait);
 			break;
 		case THREAD_SUSPENDED:
 			threadInfo.stateDescription = "Suspended";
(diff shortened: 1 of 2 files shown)


jpd002:
I'm fine with the additional command line option to generate the "WithDebugger" build variants even though they don't show up in the Configuration selection combo box in VS. 

I wouldn't build that config on AppVeyor/Travis yet because it will probably interfere with the deployment stuff, but I think it would be a good idea to build it later.

The CMake config is just missing a reference to `../Source/ui_win32/ElfViewRes.rc`. Without that, it crashes when trying to open `ELF File Information` in the debugger.

Thunder07:
Ok, I've rebased on the latest master, added the missing `ElfViewRes.rc` file (Crash fixed), and removed the appvoyer build commit.

and the issue I mentioned earlier, small/empty non-closable window, seemed to have resolved itself.

jpd002:
Thanks for this!

PR #543, opened 2017-04-01
Linux Deb Package

Thunder07:
I've been looking for the best way to create deb package for Linux, but their isn't a straightforward to manage qt dependency across different Linux platforms and to be honest, even on a single platform. 
.e.g. I'm running Ubuntu 16.10 with qt versions 5.5.1, while we're compiling for 5.6. 

This solution should allow the users to run Play if they installed qt using the installer, through  beineri ppa and in theory if the latest distro version is higher than 5.6.0(this I haven't tested).


build_cmake/CMakeLists.txt
```diff
@@ -523,6 +523,25 @@ if (UNIX AND NOT APPLE AND NOT ANDROID)
 
 	add_executable(Play ${QT_SOURCES} ${QT_MOC_SRCS} ${QT_RES_SOURCES} ${QT_UI_HEADERS})
 	target_link_libraries(Play ${PROJECT_LIBS})
+
+	install(TARGETS Play DESTINATION lib/Play/bin)
```
Thunder07: looking at this, I think `Play` should be renamed to `Play.bin` and installed to `bin` directory
while script `Play` can either stay as is, or get renamed to `Play.sh`

that should allow the install to be clearer in some sense.

PR #551, opened 2017-04-22
Qt UI: trigger bilinear filtering pref correctly

Thunder07:


jpd002:
Just checking this PR is not missing anything: The only purpose of this is to rename the function so it's consistent with others?

Thunder07:
yes, Qt uses a specific naming convention to automatically setup signal/slots. which is why you don't see a `connect()` for any of these checkboxes signals.
in this case this is the slot function for this [checkbox](https://github.com/jpd002/Play-/blob/47ece5f3df3285cea93f2af8070f6cd9cdb8eb10/build_unix/settingsdialog.ui#L149).

jpd002:
Ah, nice, I didn't know Qt did that. Thanks!

Thunder07:
ya, Qt is great for stuff like that, you can check the self generated ui_*.h in the build directory for more detail more specifically this line `QMetaObject::connectSlotsByName(SettingsDialog);`  [Link](http://doc.qt.io/qt-5/qmetaobject.html#connectSlotsByName), the side effect is, if you manually  add something like this, you don't get any error or warnings as it's technically correct code, it just does nothing.

with that said, I usually use qt creator as it generates a correctly named functions, but it seems in this particular case, I didn't.

PR #553, opened 2017-04-22
Skip travis deployment

Thunder07:
this should help avoid build failures due to failed deployment on forks


jpd002:
Makes sense! Thanks!

PR #555, opened 2017-04-29
Android: build date fix

Thunder07:
Someone posted about this not working correctly on G+ community.

Not sure when this broke, but if i had to guess it was with the gradle upgrade.
I've added the build hour into the date as well, as it might be useful should you release multiple updates within the same day.

I got this fix from StackOverflow [post](http://stackoverflow.com/questions/3540739/how-to-programmatically-read-the-date-when-my-android-apk-was-built)

### DIFF
--- a/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
+++ b/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
@@ -169,23 +169,6 @@ public class MainActivity extends AppCompatActivity implements NavigationDrawerF
 		));
 	}
 
-	private static long getBuildDate(Context context) 
-	{
-		try
-		{
-			ApplicationInfo ai = context.getPackageManager().getApplicationInfo(context.getPackageName(), 0);
-			ZipFile zf = new ZipFile(ai.sourceDir);
-			ZipEntry ze = zf.getEntry("classes.dex");
-			long time = ze.getTime();
-			return time;
-		} 
-		catch (Exception e) 
-		{
-
-		}
-		return 0;
-	}
-
 	private void displaySimpleMessage(String title, String message)
 	{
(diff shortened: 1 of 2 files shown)


Source/ui_android/java/com/virtualapplications/play/MainActivity.java
```diff
@@ -260,8 +243,8 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
 
 	private void displayAboutDialog()
 	{
-		long buildDate = getBuildDate(this);
-		String buildDateString = new SimpleDateFormat("yyyy/MM/dd", Locale.getDefault()).format(buildDate);
+		Date buildDate = BuildConfig.buildTime;
+		String buildDateString = new SimpleDateFormat("d MMM yyyy hh:mm a", Locale.getDefault()).format(buildDate);
```
jpd002: Is there any reason behind the date format change? I think it should stay the same as before, if possible.

Thunder07: the hours was meant to allow you to identify different release on the same day... or are you referring purely to the date format/structure?

jpd002: Yeah, I was just referring to the date format. I think adding time is a good idea, so you can add it back if you want. Thanks!

jpd002: Yeah, I was just referring to the date format. I think adding time is a good idea, so you can add it back if you want. Thanks!

Thunder07: the american date format just doesn't click with me XD

jpd002:
Awesome! Thanks a bunch!

PR #557, opened 2017-05-04
Cleanup

Thunder07:


Source/ee/IPU.cpp
```diff
@@ -734,7 +734,9 @@ void CIPU::COUTFIFO::RequestGrow(unsigned int size)
 	while(m_alloc <= (size + m_size))
 	{
 		m_alloc += GROWSIZE;
-		m_buffer = reinterpret_cast<uint8*>(realloc(m_buffer, m_alloc));
```
jpd002: Indeed, the code doesn't handle the case where `realloc` fails.

But your proposed change seems dangerous. If `realloc` fails, `m_buffer` is not updated while `m_alloc` is. Thus, the rest of the code will assume `m_buffer` points to a memory area of `m_alloc` size when it's not, causing potential memory overruns.

Unless there's a specific reason to address the failing `realloc` case, I wouldn't touch this code. If the allocation fails, there's not much we can do anyways (besides crashing).

Source/ee/SIF.cpp
```diff
@@ -291,7 +291,7 @@ void CSIF::SaveState(Framework::CZipArchiveWriter& archive)
 	{
 		CStructCollectionStateFile* callRepliesFile = new CStructCollectionStateFile(STATE_CALL_REPLIES_XML);
 		for(CallReplyMap::const_iterator callReplyIterator(m_callReplies.begin());
-			callReplyIterator != m_callReplies.end(); callReplyIterator++)
+			callReplyIterator != m_callReplies.end(); ++callReplyIterator)
```
jpd002: Should use a [range for loop](http://en.cppreference.com/w/cpp/language/range-for) instead.

Source/ee/Vif1.h
```diff
@@ -21,7 +21,7 @@ class CVif1 : public CVif
 private:
 	void			ExecuteCommand(StreamType&, CODE) override;
 
-	void			Cmd_DIRECT(StreamType&, CODE);
+	void			Cmd_DIRECT(StreamType&, CODE&);
```
jpd002: Kinda pointless and possibly wasteful on 64-bits machines. `CODE` is a 4 bytes value, a reference is either 4 bytes or 8 bytes.

Source/ee/IPU.cpp
Thunder07: ah I see... and subtracting m_alloc would cause a potential infinite loop. but won't sending old buffer better than sending null back?

Source/MIPSAssembler.cpp
```diff
@@ -42,15 +42,14 @@ void CMIPSAssembler::CreateLabelReference(LABEL label)
 
 void CMIPSAssembler::ResolveLabelReferences()
 {
-	for(LabelReferenceMapType::iterator labelRef(m_labelReferences.begin());
-		m_labelReferences.end() != labelRef; labelRef++)
+	for (auto& labelRef : m_labelReferences)
```
jpd002: Please keep no spaces between `for` and parenthesis. Also, I think `labelRef` could be made `const`.

Source/MemoryMap.cpp
```diff
@@ -70,10 +70,8 @@ void CMemoryMap::InsertMap(MemoryMapListType& memoryMap, uint32 start, uint32 en
 
 const CMemoryMap::MEMORYMAPELEMENT* CMemoryMap::GetMap(const MemoryMapListType& memoryMap, uint32 nAddress)
 {
-	for(MemoryMapListType::const_iterator element(memoryMap.begin());
-		memoryMap.end() != element; element++)
+	for (const auto& mapElement : memoryMap)
```
jpd002: Please keep no spaces between `for` and parenthesis.

Source/MipsExecutor.cpp
```diff
@@ -148,10 +148,8 @@ bool CMipsExecutor::MustBreak() const
 {
 	uint32 currentPc = m_context.m_pAddrTranslator(&m_context, m_context.m_State.nPC);
 	CBasicBlock* block = FindBlockAt(currentPc);
-	for(auto breakPointIterator(m_context.m_breakpoints.begin());
-		breakPointIterator != m_context.m_breakpoints.end(); breakPointIterator++)
+	for(auto breakPointAddress : m_context)
```
jpd002: Not sure that compiles. Should probably be `m_context.m_breakpoints`

Thunder07: you're right, i missed that it was inside a debug if preprocessor, so i assumed it compiled when I tested it, when in fact it wasn't even being included.

Source/MipsExecutor.cpp
```diff
@@ -363,14 +361,13 @@ void CMipsExecutor::PartitionFunction(uint32 functionAddress)
 	//Create blocks
 	{
 		uint32 currentPoint = -1;
-		for(PartitionPointSet::const_iterator pointIterator(partitionPoints.begin());
-			pointIterator != partitionPoints.end(); pointIterator++)
+		for(auto partitionPoint : partitionPoints)
```
Thunder07: @jpd002 I just want you to confirm if my naming choice(`partitionPoint`)  is correct here?

jpd002: Yep, all good!

Source/MipsExecutor.cpp
```diff
@@ -364,7 +364,7 @@ void CMipsExecutor::PartitionFunction(uint32 functionAddress)
 	{
 		uint32 currentPoint = -1;
 		for(PartitionPointSet::const_iterator pointIterator(partitionPoints.begin());
-			pointIterator != partitionPoints.end(); pointIterator++)
+			pointIterator != partitionPoints.end(); ++pointIterator)
```
jpd002: Humm, dunno why I said that, but never mind, you can't use a range for loop here.

Thunder07: it's all good, I've noticed the iterator below and figured it out.

jpd002:
Thanks!

PR #558, opened 2017-05-04
Cleanup

Thunder07:
Silence:"warning: '0' flag ignored with precision and ‘%X’ gnu_printf format"
more specifically on ubuntu builds.

I've kept the 2 Cleanup PR separate since this is warning silencing, while the other one is potential optimisation, if you'd like them to be the one PR, i can do that.

edit:
```
   grep -lR "%0.*X" . | awk '{print $1}' | sort -u | xargs sed -i '' -e "s/%0\.4X/%04X/g"
   grep -lR "%0.*X" . | awk '{print $1}' | sort -u | xargs sed -i '' -e "s/%0\.8X/%08X/g"
   grep -lR "%0.*X" . | awk '{print $1}' | sort -u | xargs sed -i '' -e "s/%0\.2X/%02X/g"
   grep -lR "%0.*X" . | awk '{print $1}' | sort -u | xargs sed -i '' -e "s/%0\.6X/%06X/g"
   grep -lR "%0.*X" . | awk '{print $1}' | sort -u | xargs sed -i '' -e "s/%0\.3X/%03X/g"
   grep -lR "%0.*i" . | awk '{print $1}' | sort -u | xargs sed -i '' -e "s/%0\.2i/%02i/g"
   grep -lR "%0.*d" . | awk '{print $1}' | sort -u | xargs sed -i '' -e "s/%0\.2d/%02d/g"
   grep -lR "%0.*d" . | awk '{print $1}' | sort -u | xargs sed -i '' -e "s/%0\.8d/%08d/g"
```
just wanted to update this post with how I went about replacing/removing the zero's.

### DIFF
--- a/Source/BasicBlock.cpp
+++ b/Source/BasicBlock.cpp
@@ -115,7 +115,7 @@ void CBasicBlock::Compile()
 		jmethod.method_size = m_function.GetSize();
 		jmethod.line_number_size = 0;
 
-		auto functionName = string_format("BasicBlock_0x%0.8X_0x%0.8X", m_begin, m_end);
+		auto functionName = string_format("BasicBlock_0x%08X_0x%08X", m_begin, m_end);
 
 		jmethod.method_name = const_cast<char*>(functionName.c_str());
 		iJIT_NotifyEvent(iJVM_EVENT_TYPE_METHOD_LOAD_FINISHED, reinterpret_cast<void*>(&jmethod));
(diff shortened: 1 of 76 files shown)


jpd002:
I started using `%08x`, `%04x` and such in new code to avoid those warnings. Could you adapt your script to change the format specifiers to this format instead?

Thunder07:
Done. I've also updated the posted script for future reference.

jpd002:
Awesome! Thanks!

PR #561, opened 2017-05-06
OSX: add 'bin' disk format as supported filetype

Thunder07:


jpd002:
Thanks. 

Personal notes: At some point, it would be nice to have that supported disc image format list stored somewhere that would be the same for all platforms. Also, reading disc data directly from the bin file isn't really the best way to go. We'll need some cue-sheet support soon.

PR #567, opened 2017-05-17
Fix a typo in time formatting

LoungeKatt:
This was displaying 24 hour format with AM/PM (ex. 23:17 PM). 
It would either be "hh:mm" or "K:mm a", but not a combination of both.

### DIFF
--- a/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
+++ b/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
@@ -244,7 +244,7 @@ public class MainActivity extends AppCompatActivity implements NavigationDrawerF
 	private void displayAboutDialog()
 	{
 		Date buildDate = BuildConfig.buildTime;
-		String buildDateString = new SimpleDateFormat("yyyy/MM/dd hh:mm a", Locale.getDefault()).format(buildDate);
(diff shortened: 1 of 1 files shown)


Thunder07:
I just tested the [latest](http://playbuilds.s3-us-east-2.amazonaws.com/0723ecb2/Play-release.apk) Travis build, which is the cleanup commit from `aoe2` branch and it correctly showed `03:25 PM`, any chance you're running a Samsung device? (they tend to mess with android's inner workings ALOT)

LoungeKatt:
If you have the choice between compatibility with all devices or telling every user of the most prominent manufacturer that they "tend to mess with android's inner workings ALOT" then it seems irrational to choose the tedious route of having to explain that every time to avoid changing an "hh" to a "K".

Thunder07:
anyway, google's own [documentation](https://developer.android.com/reference/java/text/SimpleDateFormat.html), states `h = Hour in am/pm (1-12)` while `K = Hour in am/pm (0-11)`, I've personally never seen 00:00 PM/AM being used before, but ill leave the call to Jean.

LoungeKatt:
I am glad you understood the purpose of the change. As you said:

> it correctly showed 03:25 PM

> I've personally never seen 00:00 PM/AM being used before

Despite the contradiction, you can see where this change would not only be compatible but also a little less "unorthodox" lol

LoungeKatt:
I was not aware of any situation where that call was yours, but thanks for permission anyway ;)

PR #568, opened 2017-05-17
Time format compatibility, Cleanup

LoungeKatt:
This aims to resolve a few issues with compatibility for "the rest of us"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,8 @@
 .gradle
 build_android/build
-build_android/src/main/libs
-build_android/src/main/obj
 build_android/src/main/assets/patches.xml
 build_android/src/main/jni/ExternalDependencies.mk
+build_android/keystore.properties
 build_android/local.properties
 build_android/proguard-project.txt
 build_android/.*
--- a/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
+++ b/Source/ui_android/java/com/virtualapplications/play/MainActivity.java
@@ -244,7 +244,7 @@ public class MainActivity extends AppCompatActivity implements NavigationDrawerF
 	private void displayAboutDialog()
 	{
 		Date buildDate = BuildConfig.buildTime;
-		String buildDateString = new SimpleDateFormat("yyyy/MM/dd hh:mm a", Locale.getDefault()).format(buildDate);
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -10,6 +10,7 @@ build_android/.*
 build_android/*.apk
 build_android/Play.iml
 build_android/projectFilesBackup/
+build_ios/build
```
jpd002: Do we need to ignore that directory? The build folder for iOS was removed a while ago now that CMake is used to generate the project. I might be wrong, but there's probably no references to "build_ios" anywhere in the project at the moment.

LoungeKatt: Not sure why directory deletions never seem to sync. I have the new folders, but also the old ones.

LoungeKatt:
This doesn't resolve the complete issue, since boost also looks for icu4c

```
CMake Warning at /Users/loungekatt/Public/Dropbox/PlayEmulator/Dependencies/boost-cmake/libs/locale.cmake:26 (find_package):
  By not providing "FindICU.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "ICU", but
  CMake did not find one.

  Could not find a package configuration file provided by "ICU" with any of
  the following names:

    ICUConfig.cmake
    icu-config.cmake

  Add the installation prefix of "ICU" to CMAKE_PREFIX_PATH or set "ICU_DIR"
  to a directory containing one of the above files.  If "ICU" provides a
  separate development package or SDK, be sure it has been installed.
```

README.md
```diff
@@ -61,6 +61,11 @@ cmake --build . --config Release
 ### Building for macOS & iOS ###
 If you don't have CMake installed on your system, you can install it using brew with the following command: `brew install cmake`.
 
+You will also need the icu4c library installed.
```
jpd002: It's not required to have ICU installed. I want to keep it optional because ICU is so huge and it wouldn't make sense to add all that weight to the executable just to be able to decode Shift-JIS strings. Most OS have good built-in functions that can be used. Linux/FreeBSD doesn't provide anything built-in and that's where ICU is handy.

LoungeKatt: The build fails without it installed because boost depends on it.

LoungeKatt:
26c17b0a7df1c830cec1180c5742b2bbb495b80d is the hash for just the cleanup.

jpd002:
Thanks!

PR #572, opened 2017-05-23
Reworked Unix Controls + Add Gamepad support

Thunder07:
I've been working on adding GamePad support to the linux Qt build and I found it easier to rebase all the qt controls  to the Win32 counter part, since all the logic was sorted already.

(most of the work for the gamepad support has been done, I just need to rebase it on the current set of commit, but to keep review simple, and since it's stil a WIP, ill have it in it's own PR once done)

P.S. the work was done in one go, then broken up into the 3 commit, because of that, the 2nd commit wont compile without the 3rd.

Edit: It's very likey that I'll change device from integer to byte array, since this is an example of unique ID you'd get for a device `00:22:44:CC:DD:FF`, Edit2: unique id has no standard format :/ (and not always set, using product/vender means we cant use 2 of the same product)

Edit3: I'm renaming this to a WIP, since the 2nd part would could cause issues if this is merged 1st, especially with the change from `int device` to `std::array<>`, as such it'll be part of this PR when done.


Source/ui_unix/InputBindingManager.cpp
```diff
@@ -0,0 +1,577 @@
...
+	}
+	else
+	{
+		return QString("Key: %1").arg(libevdev_event_code_get_name(m_binding.type, m_binding.id)).toStdString();
+	}
+}
+
+uint32 CInputBindingManager::CPovHatBinding::GetValue() const
+{
+	if(m_value != m_refValue) return 0;
```
Thunder07: @jpd002 quick comment/question about this.


in the Win32 implementation, you had the line `if(m_value == -1) return 0;`.
however, this didn't work in my testing with a PS4 controller, a POVHAT (in this case the D-Pad was reported as such) returned a value of `-1,0,1` along a single axis (Left/Right & Up/Down), so returning 0 meant that Left & Up did nothing.

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
```diff
@@ -0,0 +1,230 @@
...
+		do
+		{
+			if (!m_running) goto cleanup;
+		} while(poll(fds, 2, 500) == 0);
+
+		int length = read(fd, buffer, EVENT_BUF_LEN);
+
+		if ( length < 0 )
+		{
+			perror("read");
```
jpd002: Same as above. Can we continue executing if `read` fails?

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
```diff
@@ -0,0 +1,230 @@
...
+		}
+
+		for (int i=0; i < length;)
+		{
+			struct inotify_event *event = (struct inotify_event *) &buffer[i];
+			if (event->len)
+			{
+				if (std::string(event->name).find("event") != std::string::npos)
```
jpd002: Why not use `strstr` instead of creating a temporary `std::string`?

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+	std::string devinput_dir("/dev/input");
+	int wd = inotify_add_watch(fd, devinput_dir.c_str(), WATCH_FLAGS);
+	char buffer[EVENT_BUF_LEN];
+
+	while (m_running)
+	{
+		do
+		{
+			if (!m_running) goto cleanup;
```
jpd002: I would prefer avoiding `goto` at all costs. I don't think that inner do/while loop is needed. Could something like that work?
```
while(m_running)
{
  int pollRes = poll(fds, 2, 500);
  if(pollRel == 0) continue;
  ...
}
```

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
```diff
@@ -0,0 +1,230 @@
...
+					else if (event->mask & IN_DELETE)
+					{
+						RemoveDevice(event->name);
+					}
+				}
+			}
+			i += EVENT_SIZE + event->len;
+		}
+	}
+cleanup:
```
jpd002: Label unnecessary if `goto` is removed.

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
```diff
@@ -0,0 +1,230 @@
...
+	if(access( inputdev_path.string().c_str(), R_OK ) == -1)
+	{
+		return false;
+	}
+	struct libevdev *dev = NULL;
+	int fd;
+	int rc = 1;
+	bool res = false;
+
+	fd = open(inputdev_path.string().c_str(), O_RDONLY);
```
jpd002: Why not declare and init `fd` with the result of this call?

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
```diff
@@ -0,0 +1,230 @@
...
+	if (fd < 0) {
```
jpd002: Bracket style.

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
```diff
@@ -0,0 +1,230 @@
...
+		libevdev_free(dev);
+		fprintf(stderr, "CGamePadDeviceListener::IsValidDevice: Error Failed to open (%s)\n", inputdev_path.string().c_str());
+		return res;
+	}
+
+	rc = libevdev_new_from_fd(fd, &dev);
```
jpd002: `rc` could be declared and initialized here since it's not used before that line.

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
```diff
@@ -0,0 +1,230 @@
...
+	if (rc < 0) {
+		fprintf(stderr, "CGamePadDeviceListener::IsValidDevice: Failed to init libevdev (%s)\n", strerror(-rc));
+		libevdev_free(dev);
+		close(fd);
+		return res;
+	}
+
+    std::array<uint32, 6> device = CGamePadUtils::GetDeviceID(dev);
```
jpd002: Identation. Also, would probably be nicer to use `auto`.

Source/ui_unix/GamePad/GamePadDeviceListener.h
```diff
@@ -0,0 +1,48 @@
...
+	typedef std::function<void(std::array<uint32, 6>, int, int, int, const input_absinfo*)> OnInputEvent;
+
+	CGamePadDeviceListener(bool f = false);
+	CGamePadDeviceListener(OnInputEvent, bool f = false);
+
+	~CGamePadDeviceListener();
+
+	struct inputdevice
+	{
+		char *name;
```
jpd002: Why not use a `std::string` here to avoid memory ownership issues?

Source/ui_unix/GamePad/GamePadDeviceListener.h
```diff
@@ -0,0 +1,48 @@
...
+		std::array<uint32, 6> uniq_id;
+		char *path;
```
jpd002: Same as above.

jpd002:
Thanks! I haven't tested the feature yet, but I've started reviewing the code. Gonna update the review soon.

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
```diff
@@ -0,0 +1,230 @@
...
+#include <poll.h>
+#include <sys/inotify.h>
+#include <csignal>
+
+#define EVENT_SIZE		( sizeof (struct inotify_event) )
+#define EVENT_BUF_LEN	( 1024 * ( EVENT_SIZE + NAME_MAX + 1) )
+#define WATCH_FLAGS		( IN_CREATE | IN_DELETE )
+
+CGamePadDeviceListener::CGamePadDeviceListener(OnInputEvent OnInputEventCallBack, bool filter)
+	: OnInputEventCallBack(OnInputEventCallBack),
```
Thunder07: I didn't know this was possible, Thanks.

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+	m_thread = std::thread(std::bind(&CGamePadDeviceListener::InputDeviceListenerThread, this));
+	UpdateDeviceList();
+
+}
+
+CGamePadDeviceListener::CGamePadDeviceListener(bool filter)
+	: m_filter(filter)
+{
+	m_thread = std::thread(std::bind(&CGamePadDeviceListener::InputDeviceListenerThread, this));
```
Thunder07: sure, ill do that.

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
```diff
@@ -0,0 +1,230 @@
...
+	UpdateDeviceList();
+
+}
+
+CGamePadDeviceListener::~CGamePadDeviceListener()
+{
+	m_running = false;
+	m_GPIEList.clear();
+	if(m_thread.joinable()) m_thread.join();
```
Thunder07: I initially though a thread could become unjoinable if it throw an exception, it seems i was mistaken, making the check redundant.

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
```diff
@@ -0,0 +1,230 @@
...
+}
+
+void CGamePadDeviceListener::UpdateOnInputEventCallback(CGamePadDeviceListener::OnInputEvent OnInputEventFunction)
+{
+	OnInputEventCallBack = OnInputEventFunction;
+	for (auto& device : m_devicelist)
+	{
+		for (auto& GPI : m_GPIEList)
+		{
+			if(GPI.first == device.first)
```
Thunder07: haha yes, I knew there had to be a better way :/

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
```diff
@@ -0,0 +1,230 @@
...
+{
+	m_devicelist.erase(key);
+	m_GPIEList.erase(key);
+}
+
+void CGamePadDeviceListener::InputDeviceListenerThread()
+{
+	int fd = inotify_init1(IN_NONBLOCK);
+	if ( fd < 0 ) {
+		perror("inotify_init");
```
Thunder07: No, You're right.

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
Thunder07: I'll do this.

Source/ui_unix/GamePad/GamePadDeviceListener.cpp
Thunder07: not through the remaining of the loop, but calling `continue` shouldn't be an issue.

Thunder07:
I've rebased and applied all the suggestion to both `CGamePadDeviceListener` and `CGamePadInputEventListener`

Source/ui_unix/GamePad/GamePadUtils.cpp
```diff
@@ -0,0 +1,48 @@
...
+				for(int i = 0; i < strlen(tmp_id) && i < 6; ++i)
+					device.at(i) = tmp_id[i];
+			}
+		}
+	}
+	if(device == std::array<uint32, 6>{0})
+	{
+		int vendor = libevdev_get_id_vendor(dev);
+		int product = libevdev_get_id_product(dev);
+		int ver = libevdev_get_id_version(dev);
```
Thunder07: Using these 3 id's can result in a non unique ID (if you have 2 of the same controller), but gives the user the freedom to plug the controller into any port on their PC.
Using `libevdev_get_id_bustype` this will give a unique id but you'd lose that freedom.

Source/ui_unix/controllerconfigdialog.cpp
```diff
@@ -1,329 +1,99 @@
...
+		std::string button(PS2::CControllerInfo::m_buttonName[i]);
+		std::transform(button.begin(), button.end(),button.begin(), ::toupper);
+
+		InputEventSelectionDialog IESD;
+		IESD.Setup(button.c_str(), m_inputManager,
+				   static_cast<PS2::CControllerInfo::BUTTON>(i), m_inputDeviceManager);
+		if(IESD.exec())
+		{
+			m_inputDeviceManager.get()->DisconnectInputEventCallback();
+			std::this_thread::sleep_for(std::chrono::seconds(2));
```
Thunder07: I'm currently using this to give the users few seconds to release the recently pressed key, to avoid retriggering of the same key into a different binding.

I don't think it's ideal, so perhaps mapping all the keys' values in their neutral state to compare against before allowing the user to proceed.

Thunder07:
@jpd002 I want to quickly comment and get your opinion about my design choice between `CGamePadDeviceListener` and `CGamePadInputEventListener` compared to your relative implementation for the windows build.

I decided to couple them this way as oppose to your way, as it seemed to me that it will be easier to add a 2nd gamepad into the mix this way, what do you think?(maybe I missed something on the windows build?)

jpd002:
So, I finally tested that on my Linux machine and everything works nicely with my PS4 controller hooked up to the machine. Nice work :+1: .

From what I gather, `CGamePadDeviceListener` is analogous to `DirectInput::CManager` and `CGamePadInputEventListener` is analogous to `DirectInput::CDevice`? So, the implementations don't seem very different from what I can see.

Looking over the code again, I would maybe move more input handling code inside a PadHandler class instead of having that code in `MainWindow` (for example, right now the `CGamePadDeviceListener` is owned by `MainWindow`), just to keep concerns separated as much as possible. Also, since this is a Linux specific thing, it doesn't make much sense to have that code in the generic Qt section (which could be built on any other platform). Anyways, it's mostly cleanup at that point and it can be done in another PR if you'd like. I'll merge this in as it is right now.

Thunder07:
may bad, it seems that was only a thought I had as I was working on this :p

also, im not sure if I mentioned this before, but I've tested this with PS4 control with BT and USB, PS3 BT, and 3rd party usb PS3 controller.

also would enabling players 2 work at the moment, as I've noticed this fixed variable here.
https://github.com/jpd002/Play-/blob/20b3a8579778a8615e1235c62980dd56006264a5/Source/iop/Iop_PadMan.cpp#L101-L107

jpd002:
Humm, I don't remember what `PADNUM` is supposed to do, but `CPadMan` is only there for homebrew. For commercial games, I think feeding 1 as `padNumber` should work to enable 2 players. For players 3 and 4, I couldn't really say (maybe it has to go through multitap?).

Thunder07:
Also, you said you want to move this from Unix/qt section, so should it go to framework?

jpd002:
It's been a while, but I think I meant that we should try to make the Unix pad handling logic independant of the Qt code. For instance, we could want to instantiate the Unix pad handler with a single `new PH_HidUnix()` wether you're running a Qt UI or a X/Motif based UI. In other words, the code would stay in the `Play` repo but be "UI toolkit" independant.

In any case, it would just be a nice to have for now.

Thunder07:
I think the initial decision was in attempt to keep `PH_HidUnix()` conferment to the other `PH_*`,
and having access to `CInputBindingManager` ptr made configuring the buttons on the fly a bit faster, though a quick reload after returning from setting menu would work just as well (I can always have a return it from `PH_HidUnix` anyway).

but ya sure, I don't mind having another look at it, (I also want to see 2 player would work, but got distracted).

PR #574, opened 2017-06-05
Android: update games db

Thunder07:
added roughly 118 new serial code with their associated thegamedb ID, plus associated another 70 serial to their TheGameDB counterpart.

edit: note to future self, [TheGamesDB.java](https://github.com/jpd002/Play-/blob/master/Source/ui_android/java/com/virtualapplications/play/database/TheGamesDB.java#L40) version must be updated to trigger Db updates.

Edit2: the 2nd commit associates 600 serial to TheGamesDB ID counterpart.


Thunder07:
Consider this comment for discussion or my own internal monolog :p

side note:the 2nd commit was achieved by comparing the game's names to a cached=>converted to sqlite db list from TheGamesDB complete PS2 games [list](http://thegamesdb.net/api/GetPlatformGames.php?platform=11), narrowed down using sqlite `soundex` then manually selected by my self. (A basic web UI was created for this)

Now, I'm working on the remaining 2600 serials, using the names **already** in the DB, I'm sending a search query to TheGamesDB, again displaying the results and manually matching them, I've noticed out of the 100 names I've processed that when the search returns a successful match, it has always been the 1st option. pros we've an almost complete serial to name database, using it could potentially increase successful queries to TheGamesDB, cons user loses control over query should it return a false results (the chances are, if it's a false result, you won't be able to get a cover for that gain anyway)

I mention this, since currently we're querying TheGamesDB API using the file name. pros give the user the power to control the search term, cons I've games names RG CoT etc which won't match.

Best course, switch to using DB names for search + give the user the option customise the search query. This is easy enough to achieve, but I'll see if I'll get around to it.

jpd002:
Thanks for this. Going to comment on the game cover look up algorithm in your other PR.

iOS4all:
Unfortunately iOS still has no games boxart db

PR #576, opened 2017-06-09
Close stream after process without risking fail

LoungeKatt:
This ensures the stream is complete before attempting close without the
risk to kill off the functionality if close cannot be processed.


jpd002:
Oh yeah, good catch. Would it be possible to use [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) feature from Java 7 to make this code even more robust?

LoungeKatt:
That would be identical to including the close within the statement. This is meant to separate close from the code performed using the resource. There is also a bit of incompatibility with Java 7 methods and Android. They are hit or miss.

jpd002:
Alright, fair enough. Thanks!

PR #577, opened 2017-06-11
Android cleanup

Thunder07:
continuation of the other cleanup [PR](https://github.com/jpd002/Play-/pull/575).

While, I can't separate the commits, I'm opening a separate PR, to allow you to review the changes separately if you so wish.

```
TODO:Adding the serial to the database this way is not ideal, as a mismatch will persist
but as it stands that's the only way to get the pcitures and prevent them from being downloaded every run
```
this is probably the most important change, I believe `games.db` should be considered a read only reference, so we're no longer storing info to it.

### DIFF
--- a/Source/ui_android/java/com/virtualapplications/play/GameInfoEditActivity.java
+++ b/Source/ui_android/java/com/virtualapplications/play/GameInfoEditActivity.java
@@ -69,10 +69,9 @@ public class GameInfoEditActivity extends AppCompatActivity {
 
             @Override
             public void onClick(View view) {
-                gi.removeBitmapFromMemCache(intent.getStringExtra("gameid"));
-                gi.setCoverImage(intent.getStringExtra("gameid"), viewHolder, intent.getStringExtra("cover"), false, 0);
+                gi.removeBitmapFromMemCache(intent.getStringExtra("indexid"));
+                gi.setCoverImage(intent.getStringExtra("indexid"), viewHolder, intent.getStringExtra("cover"), false, 0);
                 default_cover = true;
-
             }
         });
     }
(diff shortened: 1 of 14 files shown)


jpd002:
Just to be sure, this contains all the changes from #575? If so, I think it's ok to only have a single PR.

And sorry, I merged #576 first because I thought it was an important fix (and I think it should be included in this PR too).

Thunder07:
> Just to be sure, this contains all the changes from #575?

Yes, consider it closed.

> And sorry, I merged #576 first because I thought it was an important fix (and I think it should be included in this PR too).

I've seen that PR, and I figured it will need a rebase, shouldn't be an issue.

Thunder07:
@jpd002 this should be pretty much it for this PR.

Just one notice, when I was working on the android project.
Since (if I'm not mistaken) was worked on before your setup the `Submission Guidelines`, I felt a bit lost if I should follow the old style for consistency or the new guidelines... well what ended happening was probably something in between.
Anyways, Android Studio has a really powerful and customisable style formatter `Preference->Code Style->Java`, perhaps it'll be worth while if you had a look at it, customised it, exported it and posted it up with the project, then applied it to the project as well(`right click on the java folder ->reformat code`).

This commit 118473e5f42a85491d02e5a6d0a61ecea580bd83 is mere 5 minutes worth of configuration.

jpd002:
Oh yeah, good point about the automatic formatting. I should definitively look into that. There's also clangformat that should be used for C++.

jpd002:
Thanks for this.

I agree that `games.db` should be a read only reference. I'm not quite sure about the game information edition use case. I don't think it's really interesting to be able to do that inside the emulator. I think users would want to report bad information or supply missing information, but I don't think they would want to customize a game's appearance on the emulator's launch screen. Anyways, we can leave it there for now, but we'll have to re-evaluate at some point.

Also, I think all the "game information fetching logic" should be ported to C++ at some point because the emulator should have a single method of obtaining a game's information that is the same on all platforms.

Thunder07:
> Also, I think all the "game information fetching logic" should be ported to C++ at some point because the emulator should have a single method of obtaining a game's information that is the same on all platforms.

I agree, this is something that has been on my radar for a while, but it always boils down to "Should I do X or Y 1st", at this stage, I have one more android clean up PR and then I might make the next one that. (I'm busy for this month so won't be until August, I know I've said that before)

> I don't think it's really interesting to be able to do that inside the emulator.

again, I agree, I see 2 issue though, the error could be as simple as changing GameID in the DB, or it might have to be changing/adding covers on TheGamesDB.net site (while the latter is ideal for many reasons), the issue is someone would need to do it..... the 2nd issue is, where would users report the issue, while G+ can work, it's not ideally organised for this, which goes back to an old topic, a forum is needed with a dedicated section for reporting covers.

PR #584, opened 2017-06-27
Android Cleanup

Thunder07:
Calculate sample size without having to rely on ImageView.
move the common image sampling function to it's own class.


jpd002:
Thanks!

PR #585, opened 2017-06-27
Android: Add GridView Cover Animation

Thunder07:
added basic fade/zoom animation, both as an aesthetic feature, and also to give the background thread few (milli)seconds to load the cover image.


jpd002:
Interesting effect, but not really my cup of tea. I think it would be more interesting to have a "spinner" on the covers to indicate that the image is being loaded and such.

I'm not going to merge this PR, but thanks for the submission.

Thunder07:
I can have a look when I get the chance, but I can't visualise it looking nice, since it'll be visible on each scroll up and down.

PR #617, opened 2017-08-18
CPU instruction set warning (readme)

atomic83GitHub:
Add a warning about the instruction set.
Play! don't support instruction set under SSSE3.


jpd002:
Thanks!

This reminds me that the whole readme file would need to be revamped. So much old information in there.

PR #623, opened 2017-09-01
Remove Spongebob and Incredibles patch

atomic83GitHub:
These games has been fixed and patch is not needed anymore.


jpd002:
Thanks. I'll merge this once the proper fix is in master.

jpd002:
I had already removed the Spongebob patch in my work branch that I merged yesterday and I cherry picked your commit that removes the patches for the Incredibles games. Should be all good now.

PR #628, opened 2017-09-05
[DO NOT MERGE] Add pre-compiled binaries for Linux, via AppImage, update Qt on linux to 5.9.1

hcorion:
This adds pre-compiled linux binaries.

---

This won't work right now, because it requires adding `make install` [step to cmake](https://cmake.org/cmake/help/v3.0/command/install.html).

Namely, the `install` step needs to install:
- A Play.desktop
- An icon
- The executable

You can look at the [Arch Linux package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=play-emu-git) for [inspiration](https://aur.archlinux.org/cgit/aur.git/tree/?h=play-emu-git).

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -57,7 +57,7 @@ travis_script()
         if [ "$TARGET_OS" = "Linux" ]; then
             if [ "$CXX" = "g++" ]; then export CXX="g++-5" CC="gcc-5"; fi
             export PATH=/opt/cmake-3.8.1-Linux-x86_64/bin/:$PATH
-            source /opt/qt591/bin/qt591-env.sh || true
+            source /opt/qt59/bin/qt59-env.sh || true
             cmake .. -G"$BUILD_TYPE" -DCMAKE_PREFIX_PATH=/opt/qt591/ -DCMAKE_INSTALL_PREFIX=./appdir/usr;
             cmake --build .
             cmake --build . --target install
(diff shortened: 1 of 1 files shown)


Thunder07:
@jpd002 I had initially looked at `linuxdeployqt` to create a release package for linux systems, but the resulting package was over 100mb which i thought was less than ideal, as such I dismissed it.

if you do want to go down that route, I can send you a PR for it, else there is my `workaround` in the other PR.

hcorion:
IMO deb is the wrong way to go about providing packages like this, as they do not work on all of the common distros.

Thunder07:
And 100mb package is? Beside completely killing AWS storage space. 

As the name suggests, deb packages will work on debian based distros as they are meant to, while I hasn't tested anything beside Ubuntu, I made it with that in mind.

hcorion:
Yeah, 100 MB is ridiculous, but I'm not sure how you got that. RPCS3's AppImage, which has a rather large codebase, llvm libs, libpng, ffmpeg and of course the latest Qt libs, is 43 MB.

Thunder07:
Well, I'll give you that, that was 8 months ago, so the number could be fuzzy, but those numbers are exactly why I ditched appimage as well, when Jean decides, I'll have another look.

hcorion:
:+1: sounds good to me.

Should probably get `make install` working in this PR...

PR #636, opened 2017-09-27
Osx fix

Thunder07:


Source/ee/EeExecutor.cpp
```diff
@@ -41,7 +41,10 @@ CEeExecutor::CEeExecutor(CMIPS& context, uint8* ram)
 
 CEeExecutor::~CEeExecutor()
 {
-
+#if defined(__APPLE__)
```
jpd002: Just to stay consistent with other platforms, I think this cleanup code should be moved in `RemoveExceptionHandler`.

Thunder07: Sorry for taking this long to get back to this.

I've just checked, moving this to `RemoveExceptionHandler` still causes the same crash, most likely since the class is deconstructed before the thread is stopped.

jpd002:
Thanks. 331cf44917fa04a5a62a99bfed89d87003ece54b was needed to prevent a crash on iOS.

Thunder07:
cool, Thanks.

PR #641, opened 2017-10-18
Chinese Localization

qingan1996:


Thunder07:
im a bit late with my comment, but you shouldn't have merged this one, instead, the text should have been moved out into `strings.xml`then localised... now you have 2 layouts to worry about.

jpd002:
Oh yeah, right! I had not thought about that! I'll fix it.

PR #640, opened 2017-10-16
Please merge my Chinese Translation

qingan1996:

PR #643, opened 2017-10-18
Chinese Localization

qingan1996:


Margen67:
Is 4 PRs really necessary? #643(current) #642 #641 #640

jpd002:
Working good, thanks for the localization @qingan1996!

PR #649, opened 2017-11-17
PsfPlayer CMake build

Thunder07:
Since PsfPlayer requires PlayCore to build, I'd imagine it's a pain trying to build with the current setup, this should rectify that.

Quick question, I have a Qt ui for PsfPlayer for ubuntu (which at the time (10ish months ago) worked for OSX/Win32 as well), however, I've no intent to maintain it, so I haven't push it with this PR, if you'd like and if you feel it's worth while pushing, let me know and I'll include it here.

Note: Win32 hasn't been tested yet, so don't merge.
also, you'd need to merge the other PRs in all the other Repo for this to work.

Edit: ~~note to self, `SpuRegView.h` change `m_listView` && `m_extrainfo` to unique pointers~~
`CLayoutWindow` will handled deleting those pointers once it itself has been deleted.


Thunder07:
To comment on Win32 Build, the last 2 commit would allow it to nearly build, it's currently failing due to missing `d3dx3.h` which has been depreciated and even dropped from this project

Good news is, there isn't much reference to `d3dx3.h`, the bad news I haven't dealt with DirectX before, so this might take some time.

Thunder07:
ok `d3dx9` dependency removed from PsfPlayer.

Note:I've noticed I have less flickering when I moved the refresh into a separate `std::thread` and used `chrono` for timing, so it could be a combination of that, or maybe `chrono` is just more accurate as such the view is more insync with monitor so flickers less?

the only thing remaining is to test that the reset of the build didn't break, I'll do that tomorrow.

Thunder07:
ok, so hopefully thats the last of that, build should be working correctly for all builds now Play/PsfPlayer across the supported platforms. (I didn't add PsfPlayer iOS support, since I wouldn't be able to test it).

I pushed the Qt UI, since it'll be a waste since It's fully functional and works on all Ubuntu/OSX/Win32.

Thunder07:
@jpd002 as I working on this PR, I've noticed that cmake needed some work, and even with this PR I felt that there was more that needed to be done, now especially since the other project solution has been removed.

so for the past 2 days I've reorganised all of cmake in this repo (the other repos aren't bad since they don't have much dependencies) and I based it on this PR since it would be a night mare trying to navigate 2 PR with huge cmake changes... these are the results so far (hopefully for the better)[cmake_cleanup_rebased_psf_build](https://github.com/Thunder07/Play-/tree/cmake_cleanup_rebased_psf_build), last few commits needs merging back but as far as I can tell that should be most of it.

the question I came here for, since it's based on this PR and this PR has a lot of cmake changes, would I open in it, into a new PR or push it to this? (if I fix the branch before you reply, ill push it to a new PR and will work from there)

Thunder07:
merged as part of #660

PR #654, opened 2017-12-07
Better ObjC property declarations.

MaddTheSane:
As a general rule, delegates and `IBOutlets` are `weak`, `NSString`s should be `copy`.


jpd002:
Thanks for this!

PR #658, opened 2018-01-10
Qt gamepad cleanup master

Thunder07:
added 3 second button registration window when configuring gamepad, this allows us to remove the 2 second sleep i previously added to avoid the same button getting registered multiple time.

I've removed direct calls to UI when configuration controllers as that was on occasion causing crashes.

lastly, control filter that was setup to prevent certain buttons from spamming the configuration also prevented the range condition `if (value < midVal + triggerVal && value > midVal - triggerVal)` from triggering, so now we'll check the last 20% of the range, which allows us enough range to set an on/off region

edit, you also asked to maybe move the gamepad files our of `ui_unix`, would framework be a better place?


jpd002:
Thanks for this! Yeah, the files in the `GamePad` folder would be well suited for the `Framework` project, but no need to move them right now since they would only be used by Play!.

PR #660, opened 2018-01-13
CMake Cleanup

Thunder07:
- Update cmake to expect 3.5, since we've made few changes to how certain libraries are listed for build (Bzip, zlib)
- completely reorganised cmake, into how it's usually used
  - This also helped organise dependencies
- No longer make any calls to `include_directories` since each dependency will provide it's own includes with if its linked
- made test builds optional but on by default
- Can also build PsfPlayer as part of the same Play build, thus avoiding completely separate build where you'd have to build PlayCore for each

I've can confirmed that for commit `aae67c7` I can perform a [fresh build for each](https://gist.github.com/Thunder07/ab8bdc7f67b62a6c491664bdd11e6296) `CMakeLists.txt` for Win32/Linux/OSX/iOS, I couldn't run the same test for android since I couldn't get standalone cmake to work with it. I've also got a [travis](https://travis-ci.org/Thunder07/Play-/builds/328367819?utm_source=github_status&utm_medium=notification) and [AppVeyor](https://ci.appveyor.com/project/Thunder07/play/build/353) to build that commit (with changes to Play-Build and the build scripts).

as for commit `d3612e3` I've done a fresh Ubuntu build with the same above script.
I've recompiled OSX and Win32 builds(, Android iOS shouldn't be effected, so I didn't build them).

Lastly, commit upto `initial PsfPlayer Qt Client 84e435b` are part of this PR  #649 
This PR would need branch `cmake_2` of `Framework` & `Dependencies` and branch `cmake` for `CodeGen` and`Nuanceur`.
(I didn't open a PR for Framework & Dependencies, since I wanted to see if you want to review this PR and #649 or as separate ones)


jpd002:
Nice! My time is very limited right now, but I'll test that on my side and give you feedback ASAP.

Thunder07:
Alright 👍 
the only comment I forgot to mention was in relation to window's `GSH_Direct3D9*.cpp`,
OpenGL GS has it's own cmake file since the files has been placed in a separate folder while DirectX haven't, I thought about moving them to `gs` folder as well, but wasn't too sure if there is a reason to why you kept it in the ui folder, so I didn't?

I also noticed the merger conflict, and it seems I didn't remove common Win32 definition from AutoTest, and ill get to both within the next 2 days.

Thunder07:
PR has been rebased as of `master` https://github.com/jpd002/Play-/commit/d153162bbdea30fc57f2509fcb0d5c1e75ddd3d2, with above included cleanup.

[AppVeyor](https://ci.appveyor.com/project/Thunder07/play/build/360)* & [Travis](https://travis-ci.org/Thunder07/Play-/builds/329866043).
* AppVeyor builds but fails to upload, since it doesnt have the AWS keys, still haven't figured if there is a proper way to disable it... however, its only an issue in my CI and wont be in master

Thunder07:
This PR would need branch `cmake_2` of Framework & Dependencies repos.

I just push an update to them, but I haven't had the chance to test them out on anything beside OSX, so I'll do so tomorrow and open a PR for them as well.

Thunder07:
I can confirm with the PR for Framework & Dependencies you should be able to build `Play!` for OSX, Win32 (Win32 ran the stress test).
I don't think anything has changed yet that would require me to run stress test for OSX/iOS/Unix but should I get the time, I'll do just that to be sure.

jpd002:
Thanks for this, it's a great cleanup! The only thing that I don't really like is the fact that `CMakeLists.txt` are mixed up with source code. I would like to keep all build config files in `build_cmake` directory to keep them separate from source. Otherwise, everything looks good to me.

Thunder07:
`CMakeLists.txt` are somewhat analogous to `Makefile`, they're meant to be with the source code they contain, and to manage it dependency, lets say someone wanted to integrate `PlayCore` source into there own (lets say without cloning this repo, to avoid ui, tools, test and other resources), they just need to copy `source`, (they can delete the UI if they wanted), and just `add_subdirectory` to their project and they're done.

but if you move the `CMakeLists.txt` to an arbitrary relative out of source directory, that previous scenario would need revision as `CMakeLists.txt` automatically breaks down, since all it's source files are now in the wrong relative directory.

continuing with the `Makefile` analogy, the change would require creating multiple folders in `build_cmake` to contain each `CMakeLists.txt`, and I think you'd agree that it would be rather good, doing the same thing for `Makefile`

with that said, I know the relative path to external dependencies, in this case `Framework`, `CodeGen`, `Dependencies`, etc would fail, in both cases, but you probably noticed how the main `CMakeLists.txt` does `add_subdirectory(Source/ui_unix/)` without having to specify and output directory, which is the ideal structuring method which would also prevent their failure, but its not really achievable in this cases without also placing those repos directly inside this(as submodules) and also restructuring those (which is how that should really be done).

if you really want it so, I can get around to doing so, but I'd highly highly recommend against it.

jpd002:
Thanks for the explanation, the `Makefile` analogy makes sense. I've looked at other projects and it looks like a common approach to place `CMakeLists.txt` with source code. Let's keep it that way then.

Lemme run some other tests and I'll merge that PR if everything works good.

jpd002:
Building with debugger (setting `DEBUGGER_INCLUDED` option) on Win32 doesn't seem to work. I haven't investigated, but I guess the setting isn't propagated to other CMakeLists for some reason? Were you able to build it properly with that setting on?

Thunder07:
I just tried it, and it's not working, I'm behind the screen now, so ill check it out now.

Thunder07:
this commit should fix it, but PsfPlayer DEBUGGER_INCLUDED build still fail since I'm not compiling the debugger for it, ill have a look at that now.

jpd002:
Great work, thank you very much! There's some minor issues, but I'm going to merge this PR since everything is functional. 

The issues I saw:
- Root project name is not set (CMake generates Project.sln instead of something more meaningful).
- Manifest issues in PsfPlayer (dpi scaling doesn't work).
- TDEmu*.dll not copied automatically to executable directory.

I'm going to change Travis and AppVeyor to build PsfPlayer automatically.

jpd002:
Fixed project name: e24267b13e709b147e34e3c2c6a9f87e403b2ea4.

Thunder07:
Sure thing :)
I'll see what I can do with the other 2 issues.

jpd002:
Thanks!

Seems like the project name fix broke the build on Android for some reason... I've reverted it for now.

Thunder07:
that could be because I used the name `Play` for the android `CMakeLists.txt` causing a naming conflict, while the other UI use the name `PlayUI`.
anyway, I didn't feel it's correct to name it `PlayUI` since thats technically wrong.

jpd002:
Tried to rename the Android project to `PlayUI` and added `project(Play)` in the root `CMakeLists.txt` and it's still broken. If `project(Play)` appears before `include(Header)` in `CMakeLists.txt`, everything works fine on Android. I guess the `project` command needs to appear before `include` but I know it causes some issues on macOS because of the deployment target. Guess we'll need to reorganize things a bit if we want to change the project name.

Thunder07:
🤔 sounds like a lot of work for a little return

PR #666, opened 2018-03-04
MacOSX: Add Thumb & Trigger button mapping

Thunder07:


jpd002:
Makes sense! Thanks a lot!

PR #668, opened 2018-03-05
Psfplayer fixes

Thunder07:
would need the following [PR ](https://github.com/jpd002/Play-Dependencies/pull/12) in `Dependencies` to work correctly


jpd002:
Works great! Thanks!

PR #682, opened 2018-04-08
Restore a missing Max Payne commit

atomic83GitHub:
A commit was not ported from the Max Payne branch to the Master.

Changes made by jpd002


jpd002:
Thanks. 

I left that commit out because it was breaking more things that it fixed. The whole SPU timing subject is quite tricky and I decided to stay away from that till I get a better idea on the way tests could be built to prevent such breakage.

Issue #676, opened 2018-04-06
Sticks Initialized Wrong


Nine-H:
what it says on the tin, I used padtest.elf to test the input because I noticed input bugs on every game, turns out the sticks are initialized to top right and moving one axis moves the other a little, but there's a workaround, moving the sticks:

initialized:

![screenshot from 2018-04-06 20 46 43](https://user-images.githubusercontent.com/20123543/38417739-d907043e-39dc-11e8-9790-27c89199025a.png)

after some waggling:

![screenshot from 2018-04-06 20 47 57](https://user-images.githubusercontent.com/20123543/38417651-7def5010-39dc-11e8-91d9-bfba4e4a1f03.png)

http://www.theisozone.com/downloads/playstation/tools/padtest-controller-test-app/
here's a link to the binary I used, it might be worth tracking down the author to integrate the software as a unit test for input.

here's some search terms so this doesn't get duped: joystick, analog, input, analogs, dualshock

Thunder07:
ummm... the Qt build, has the analog initiated to `0x7F` (127), so I don't see why this is happening.

@Nine-H you said you had bugs in every game, what games have you tested, because while I haven't used `pastiest.elf`, and my gaming collection is small I don't recall having issues in KH, FFXII and few others that I've tested.

also, are you using the keyboard or controllers/joysticks?
I'll boot into ubuntu later to see if anything sticks out.

Nine-H:
I was using keyboard, is ubuntu using libinput or xinput because that's all I can think of, I'm also building against qt5-default for 16.04 which may be out of date.

Thunder07:
I don't know about Ubuntu itself but the answer should be neither, atleast neither directly, the keyboard input is acquired through qt events, which is event based, so there shouldn't be any trigger to change the initial values (0x7F) before direct user input (granted I didn't test the keyboard, but how stupid would it be if the app started with keys pressed by default and what are the chances that only happening to the analog buttons? Stupid or not, you can still test that theory by assigning different keys to the analog, preferably keys that are used by other buttons that are not currently actived)

Nine-H:
maybe check the keyboard, I noticed that after cancelling the bug by moving the sticks clicking "reset to default" in the controller manager brought the bug back.

Thunder07:
damn this was an annoying bug to hunt, it was the keyboard or more precisely `CSimulatedAxisBinding` which is only available when using the keyboard,  there was an issue with a condition and value setting.
[...]

PR #685, opened 2018-04-11
Qt fixes

Thunder07:
the 1st commit in the PR should fix this issue https://github.com/jpd002/Play-/issues/676


Source/ui_unix/mainwindow.h
```diff
@@ -66,12 +66,16 @@ class MainWindow : public QMainWindow
     struct lastOpenCommand
     {
         BootType type;
-        const char* filename;
-
-        lastOpenCommand(BootType m_type, const char* m_filename) : type(m_type),filename(m_filename){}
+		std::string filename;
+		lastOpenCommand() = default;
+		void reset(BootType m_type, std::string m_filename)
```
jpd002: Why have a reset function instead of a constructor that takes 2 parameters? Also, I would expect a 'reset' function to clear all members/release all resources/etc..

Thunder07: fail on my part, not sure why but I initially tried using a constructor and it failed, so I went straight to this, but it's working now, ill push fix in few.

jpd002:
Thanks for this.

Yeah, running clang-format on the Unix UI files would be nice, but can be done in a separate PR. We'll have to run it on everything else too at some point (dunno if it makes sense to add it to the CI process too).

Thunder07:
> dunno if it makes sense to add it to the CI process too
I was actually going to suggest that as well.

> We'll have to run it on everything else too at some point
I think that should be done now, as formatting can causing annoying merge conflicts and right now, you and I are the only 2 that are visibly working on the project and you only have few active branches with small changes that can be easily managed, but should patron pickup support and people start to noticing the project, others might join in, then it will be too late to run it, as that could conflict with their changes leading to plenty of headaches.

jpd002:
Yeah, you're right, we'd better get that up and running ASAP. I was already pretty satisfied with the clang format settings, so, it shouldn't be too problematic to run it on everything.

PR #687, opened 2018-04-21
WIP:Clang format

Thunder07:


Thunder07:
going to close this to avoid spamming you with revision, the long story short, this works on my desktop but fails on travis :(

PR #689, opened 2018-04-25
enable clang-format travis build

Thunder07:
I pushed the 1st commit 1st in hopes to see the build fail in action, but it seems the 2nd commit stopped the build, you should be able to manually start it up to test it, just note that the resulting patch file is bigger than the limit set by termbin so it won't produce a usable file, but that shouldn't be a problem after the 2nd commit, which formats the entire project.

edit: didnt expect build to break after formatting, ill have a look at it.

edit2: ok, it seems clang-format changed the order of includes which is now causing issues, as it seems some includes dependent on others, but instead of having the include within the dependent code file, it relied on others including them.

edit3: ok this should build now on all platforms without a issue. i do recommend you restart the 1st commit travis build just to see termbin in action.

edit4: haha you need to run clang-format on your own commit too :P but this is a good example of clang-format in action

```
IncludeCategories:
  - Regex:           '^"(llvm|llvm-c|clang|clang-c)/'
    Priority:        2
  - Regex:           '^(<|"(gtest|gmock|isl|json)/)'
    Priority:        3
  - Regex:           '.*'
    Priority:        1
```
^^ this is the default `IncludeCategories` aka the one getting applied before the changes I made, didn't want to modify it too much, so I kept the llvm...json references there.


jpd002:
The infrastructure works great! Thanks for this!

I've looked at the changes made by clang-format and there's some things I don't like about the formatting. Two major things:

- Odd bracing format for [constant array defs]( https://github.com/Thunder07/Play-/blob/e0bb765d8b4716917b4bbbd6c5a6c868e728f998/Source/COP_SCU.cpp#L10). I’ll probably add clang-format on/off tags around some of the arrays (mainly reflection tables) because it doesn’t make sense to have them formatted that way.
- Formatting of lambdas. It seems that version of clang-format is having issues indenting lambdas. I’ve tried a tool (https://zed0.co.uk/clang-format-configurator/) with the same clang-format settings and the lambda looks fine on there… So, maybe it’s a clang-format version issue?

I'll keep playing with settings to see if I can find something that I would be more ok with.

Thunder07:
if there isn't any. configuration for it, then yes clang on/off would be the only option, in such case, I can remove the entire project format commit, as it'd be easier to add them tags 1st then apply clang format to entire project.

I wouldn't expect to have varying within a version except perhaps the default settings, to get around that you should run `clang-format --dump-config` in `Play!` root directory which will read the config file but also dump all the configs being used, in this way we'll have the same settings being use irrespective of the defaults (since they're specified now).

Thunder07:
do you want me to keep 70881a98fa52077a1ce218314b75de2086f0be0c aka `Build Fix`, it's no longer needed put at the same time it should really be included (renamed to cleanup)

jpd002:
Thanks for this!

PR #694, opened 2018-05-08
Android: Fix / optimize external sdcard code

LoungeKatt:
Forgot you guys were still using my pre-Android Studio version of this, so here is the update that does not append the String onto itself (reduces writes and fixes an Android Studio lint warning) and prevents a lint warning by adding something into the catch block (even if it still isn't used at runtime).


jpd002:
Thanks for this!

PR #698, opened 2018-05-21
Updated boost include

BlackLotus:
Replaced deprecated include. Should compile not only on the newest boost version, but on old ones as well.

### DIFF
--- a/Source/saves/Icon.h
+++ b/Source/saves/Icon.h
@@ -2,12 +2,7 @@
 
 #include "Types.h"
 #include "Stream.h"
-
-#ifdef _WIN32
 #include <memory>
-#else
-#include <boost/config/no_tr1/memory.hpp>
-#endif
 
 class CIcon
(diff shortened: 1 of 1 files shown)


Thunder07:
is the boost included even needed anymore?
I assume it was used for `std::share_ptr` which is now part of the std on all platforms

BlackLotus:
You are probably right that it can be removed. In the end boost/config/no_tr1/memory.hpp only does an #include <memory> as well. Just didn't want to break any builds. Anyway either the if clause should be removed or the new include could be added.

Thunder07:
if thats the case, just redo the commit and remove the preprocessor if statement, It shouldn't really break anything, and boost is slowly being removed as possible, so this will have to go anyway.

BlackLotus:
Done and thank you for your comments :)

jpd002:
Thanks!

PR #700, opened 2018-05-24
Update TheGamesDB to legacy domain

Zer0xFF:
the current api will be completely removed in time for now we're slow preparing to replace the current(aka old) site with the beta(new) site as such the main domain name will eventually drop the old api, and the legacy domain name will be available during the transition stage to the new api.

P.S. Legacy subdomain will also be eventually be put offline, but by then I'd have PR for it as well.


jpd002:
Thanks for this!

PR #701, opened 2018-06-04
OSX: Qt UI

Zer0xFF:
allowing Qt ui to work on OSX.

note the currently created OSX bundle doesn't include the Qt dependencies,
however, that can be easily fixed with `~/Qt/5.9/clang_64/bin/macdeployqt Play.app/` before packing it into `dmg`, if you'd like to immediately replace the current automated build with qt, I can go ahead and include that in this PR, else I'd keep it until I start working on using Qt GamePad


jpd002:
Thanks for this. The changes needed to make the Qt UI official on macOS can be done later (I wanna change a few things myself first).

PR #722, opened 2018-10-15
Qt OSX Gamepad

Zer0xFF:
Just tested with linux, beside the 2 commit that I have pushed, everything else is fine.

### DIFF
--- a/Source/ui_unix/ControllerConfig/inputeventselectiondialog.cpp
+++ b/Source/ui_unix/ControllerConfig/inputeventselectiondialog.cpp
@@ -106,6 +106,80 @@ void InputEventSelectionDialog::SetupInputDeviceManager(std::unique_ptr<CGamePad
 	GPDL.get()->UpdateOnInputEventCallback(onInput);
 }
 
+#elif defined(__APPLE__)
+void InputEventSelectionDialog::SetupInputDeviceManager(std::unique_ptr<CGamePadDeviceListener> const& GPDL)
+{
+	auto onInput = [=](std::array<unsigned int, 6> device, int code, int value, IOHIDElementRef elementRef) -> void {
+		IOHIDElementType type = IOHIDElementGetType(elementRef);
+		bool is_axis = type == kIOHIDElementTypeInput_Axis || type == kIOHIDElementTypeInput_Misc;
+		if(!is_axis)
+		{
+			if(!setCounter(value)) return;
+		}
+		QString key = QString("btn-").append(QString::number(code));
+		if(is_axis)
+		{
+			if(IOHIDElementGetUsage(elementRef) != kHIDUsage_GD_Hatswitch)
+			{
+				CFIndex triggerRange = IOHIDElementGetLogicalMax(elementRef) / 100 * 20;
+				CFIndex triggerVal1 = IOHIDElementGetLogicalMax(elementRef) - triggerRange;
+				CFIndex triggerVal2 = IOHIDElementGetLogicalMin(elementRef) + triggerRange;
+				if(value < triggerVal1 && triggerVal2 < value)
+				{
+					setCounter(0);
+					return;
+				}
+				setCounter(1);
+				setSelectedButtonLabelText("Selected Key: " + key);
+				m_key1.id = code;
+				m_key1.device = device;
+				m_key1.type = type;
+				m_key1.bindtype = CInputBindingManager::BINDINGTYPE::BINDING_SIMPLE;
+			}
+			else
+			{
+				m_key1.id = code;
+				m_key1.device = device;
+				m_key1.type = type;
+				m_key1.value = value;
+				m_key1.bindtype = CInputBindingManager::BINDINGTYPE::BINDING_POVHAT;
+				setSelectedButtonLabelText("Selected Key: " + key);
+				setCounter(1);
+			}
+		}
+		else
+		{
+			if(PS2::CControllerInfo::IsAxis(m_button))
+			{
+				if(click_count == 0)
+				{
+					setSelectedButtonLabelText("(-) Key Selected: " + key);
+					m_key1.id = code;
+					m_key1.device = device;
+					m_key1.type = type;
+					m_key1.bindtype = CInputBindingManager::BINDINGTYPE::BINDING_SIMULATEDAXIS;
+				}
+				else
+				{
+					m_key2.id = code;
+					m_key2.device = device;
+					m_key2.type = type;
+					setSelectedButtonLabelText("(+) Key Selected: " + key);
+				}
+			}
+			else
+			{
+				setSelectedButtonLabelText("Selected Key: " + key);
+				m_key1.id = code;
+				m_key1.device = device;
+				m_key1.type = type;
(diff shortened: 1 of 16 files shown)


Source/ui_unix/ControllerConfig/inputeventselectiondialog.cpp
```diff
@@ -103,9 +103,82 @@ void InputEventSelectionDialog::SetupInputDeviceManager(std::unique_ptr<CGamePad
...
+			{
+				CFIndex triggerRange = (255 * 20) / 100;
+				CFIndex triggerVal1 = 255 - triggerRange;
+				CFIndex triggerVal2 = 0 + triggerRange;
+				if(value < triggerVal1 && triggerVal2 < value)
+				{
+					setCounter(0);
+					return;
+				}
+				setCounter(1);
```
Zer0xFF: need to double check this edit

Zer0xFF: I was meant to tag the lower `setCounter(1);`. but thats fine as well

Source/ui_unix/GamePad/GamePadDeviceListener_OSX.cpp
```diff
@@ -0,0 +1,351 @@
...
+			CFRelease(usageCFNumberRef);
+		}
+	}
+	return result;
+}
+
+void CGamePadDeviceListener::InputValueCallbackStub(void* context, IOReturn result, void* sender, IOHIDValueRef valueRef)
+{
+	auto device_info = reinterpret_cast<DeviceInfo*>(context);
+	if(!device_info->OnInputEventCallBack)
```
jpd002: This tests the std::function's pointer validity and not the actual std::function's value (which can be set to nullptr by DisconnectInputEventCallback). This should be `if(!(*device_info->OnInputEventCallBack))`.

Source/ui_unix/mainwindow.h
```diff
@@ -65,7 +67,8 @@ class MainWindow : public QMainWindow
 	CPS2VM* m_virtualMachine = nullptr;
 	bool m_deactivatePause = false;
 	bool m_pauseFocusLost = true;
-#ifdef HAS_LIBEVDEV
+	CFutureContinuationManager m_futureContinuationManager;
```
jpd002: Did that get there in a rebase? I thought I got rid of that (CFutureContinuationManager) in master.

Source/ui_unix/GamePad/GamePadDeviceListener_OSX.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+CGamePadDeviceListener::CGamePadDeviceListener(bool filter)
+    : CGamePadDeviceListener(nullptr, filter)
+{
+}
+
+CGamePadDeviceListener::~CGamePadDeviceListener()
+{
+	m_running = false;
+	CFRunLoopStop(CFRunLoopGetCurrent());
```
jpd002: This doesn't seem necessary, the loop in the thread will exit because of m_running being set to false. I also think that might have unintended side effects since we're stopping the run-loop for the current thread that might be used for other system stuff. Should be removed.

Source/ui_unix/GamePad/GamePadDeviceListener_OSX.cpp
```diff
@@ -0,0 +1,350 @@
...
+		CFRelease(matchingDict[2]);
+		IOHIDManagerSetDeviceMatchingMultiple(m_hidManager, array);
+	}
+
+	IOHIDManagerRegisterDeviceMatchingCallback(m_hidManager, onDeviceMatched, this);
+	//IOHIDManagerRegisterDeviceRemovalCallback(m_hidManager, onDeviceRemoved, this);
+
+	IOHIDManagerOpen(m_hidManager, kIOHIDOptionsTypeNone);
+	IOHIDManagerScheduleWithRunLoop(m_hidManager, CFRunLoopGetCurrent(), CFSTR("CustomLoop"));
+	while(CFRunLoopRunInMode(CFSTR("CustomLoop"), 0, true) != kCFRunLoopRunStopped && m_running)
```
jpd002: Specifying 0 for the seconds parameter in CFRunLoopRunInMode will cause 100% CPU usage. Using 1 instead of 0 seems to fix the problem.

Source/ui_unix/GamePad/GamePadDeviceListener_OSX.cpp
```diff
@@ -0,0 +1,350 @@
...
+CGamePadDeviceListener::CGamePadDeviceListener(bool filter)
+    : CGamePadDeviceListener(nullptr, filter)
+{
+}
+
+CGamePadDeviceListener::~CGamePadDeviceListener()
+{
+	m_running = false;
+	CFRunLoopStop(CFRunLoopGetCurrent());
+	IOHIDManagerClose(m_hidManager, 0);
```
jpd002: The thread should be joined before closing the HID manager. I had random crashes here when quitting the app. Also, it seems `m_hidManager` could be made local to the `InputDeviceListenerThread` function. Would make sense to move the call there and remove the member from `CGamePadDeviceListener`.

jpd002:
Just a note: There seem to be another situation where CPU usage goes to 100% : when you select the button to be bound to a control. Gonna track it down.

Source/ui_unix/ControllerConfig/inputeventselectiondialog.cpp
```diff
@@ -103,9 +103,82 @@ void InputEventSelectionDialog::SetupInputDeviceManager(std::unique_ptr<CGamePad
...
+				m_key1.device = device;
+				m_key1.type = type;
+				m_key1.bindtype = CInputBindingManager::BINDINGTYPE::BINDING_SIMPLE;
+			}
+			else
+			{
+				m_key1.id = code;
+				m_key1.device = device;
+				m_key1.type = type;
+				m_key1.value = value;
```
jpd002: Don't know if that applies to DS3/DS4, but it seems I need a check for value being less than 8 here. My Moga Pro 2 reports value = 15 when the dpad is being released. Values are in range [0, 8[ when pressing the directions.

Zer0xFF: DS3 DPad is reported as individual digital buttons (to be more exact reported as a bit value per direction).
DS4 reports dpad as a range from 0-7 and 8 when released/neutral.
so the DS4 would have similar issue, however, since value is only reported when changed, I didn't find it to be an issue(aka it's only an issue when configuring, but clicking any other button and holding or even releasing would cause it to be ignored, granted not my finest solutions).

jpd002: Well, adding a check for `value < 8` would probably be a good solution. If it's less than 8, you call `setCounter(1)` and call `setCounter(0)` otherwise. That's what I did on my end and it seems to do the job.

jpd002:
Going to merge this as it is right now since it works. Going to push my fixes soon after.

PR #737, opened 2019-01-06
Update TGDB API

### LINKED ISSUE #736: Updating TheGamesDB API
Just wanted to give you a heads up, as of today, I've switched off the legacy server for TGDB API, which means Play! cover fetcher would stop working.

but I've also started working on C++ implementation for `Play!`, I'm currently using https://github.com/nlohmann/json for son parser, since we don't have one for this project.

I know you've entertained the idea of removing cover editor, so I might get that out of the way as well.

Zer0xFF:
once merged close #736 

The API key will need to be changed, I've DM'ed you my private one, but you'd need to sign up to http://forums.thegamesdb.net/ to get one

- [X] implement TGDB New API (batch requesting + pagination)
- [X] Replace Java calls with native calls
- [X] Remove redundant Java functions 
- [x] Fix Remove unavailable games
- [x] Fix sorting
- [x] Test Elf
- [x] stop ScanBootables() overwriting TGDB title with game file name
- [x] remove ExtractDiscIds()? Disc ID shouldn't change, so this could be made as part of RegisterBootable() to reduce redundant writes

note, I couldn't make java edits smaller, since the minute I switch the code to native, thus use Bootable vs GameInfoStruct, it requires a lot of code change, and breaks in some parts, e.g the editor, but since we're removing all of it, I saw little point in making small changes(fixes) until the final result.

I also have this set to merge with `shared_covers` branch.
Associated PR in Dependencies https://github.com/jpd002/Play-Dependencies/pull/13


Source/DiskUtils.cpp
```diff
@@ -128,7 +128,7 @@ static std::string GetDiskIdFromPath(const std::string& filePath)
 	auto regionCode = subFilePath.substr(0, 4);
 	auto serial1 = subFilePath.substr(5, 3);
 	auto serial2 = subFilePath.substr(9, 2);
-	return regionCode + "_" + serial1 + "." + serial2;
+	return regionCode + "-" + serial1 + serial2;
```
Zer0xFF: this will probably break iOS database scan (not that it matters since it'd be using the old API, which won't work anyway)

Zer0xFF:
Alright for the android build error, `cxx_std_11` error is caused by json parser library using that keyword, which was only added on cmake 3.8.2, so we have 2 solution, force 3.8 requirement, which has its own issues, since android studio repos only up to 3.6 so far, or we can maintain an our fork of json parse, or the simpler solution yet, create a CMakeLists.txt without that line and drop it into `Play-Dependencies/build_cmake/` as we've done with other dependencies, which is what im leaning towards.

Zer0xFF:
after a wasteful day trying to get 3.10.2 working on travis, I still dont know why it works on my local machine but not travis, I opted for creating our own cmake script for use with this :/ PR in dependencies in a bit.

jpd002:
Thanks! So, I've tested this on iOS and Android, and this works nicely. The `FetchGameTitles` function will require a try/catch block because an exception can occur when the request to TheGamesDb is executed, due to a network error (ex.: no internet connection available). 

Otherwise, it's all good!

Zer0xFF:
Great, I see that you've taken care of that already :thumbsup:

PR #740, opened 2019-01-24
Resolution Multipliers

Zer0xFF:
Notes:
~~Windows: UI needs updating, Im aware that passing some DROPDOWN flag to CListBox allows me to create a drop down menu to add resolution multipliers, but how would I create a label for it?~~
Since current Win32 UI we'll be replaced with Qt, no need to worry about this

iOS: no experience in iOS (/UI), so I can't do much about it (if it even has the option now for higher resolution?)

Qt: tested on OSX/Win32 works fine
~~Android: 👍, though I noticed some indentation inconsistency in the xml, so will fix that.
edit:
for android I noticed you're using `CSettingsManager` which then calls `CAppConfig`, is that necessary?~~

### DIFF
--- a/Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
+++ b/Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
@@ -285,7 +285,7 @@ void CGSH_OpenGL::LoadState(Framework::CZipArchiveReader& archive)
 void CGSH_OpenGL::RegisterPreferences()
 {
 	CGSHandler::RegisterPreferences();
-	CAppConfig::GetInstance().RegisterPreferenceBoolean(PREF_CGSH_OPENGL_ENABLEHIGHRESMODE, false);
+	CAppConfig::GetInstance().RegisterPreferenceInteger(PREF_CGSH_OPENGL_RESOLUTION_FACTOR, 1);
 	CAppConfig::GetInstance().RegisterPreferenceBoolean(PREF_CGSH_OPENGL_FORCEBILINEARTEXTURES, false);
 }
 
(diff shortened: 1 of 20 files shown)


jpd002:
So, only the iOS part is missing for this? I can add the necessary UI.

Zer0xFF:
iOS and the 2 android comments I mentioned above, I can sort both of them 2morrow since, I'm free.

but to point to it:
> for android I noticed you're using CSettingsManager which then calls CAppConfig, is that necessary?

jpd002:
It should go through `CSettingsManager` for consistency's sake.

Also, saw the multiplier options, I think we should limit the choices to powers of 2 (1x, 2x, 4x, 8x, 16x). Does setting a multiplier to a non power of 2 work?

I'll take care of the iOS part.

Thanks!

Zer0xFF:
> I'll take care of the iOS part.

👍 great, but by both, I meant the android issues :P
P.S haven't looked at the code in a while but looking at it now, it seems half the android code is in the wrong function XD

> Does setting a multiplier to a non power of 2 work?

ill that confirm tomorrow, but I can't imagine a hardware in the 21st that can't handle non power of 2

> I think we should limit the choices to powers of 2

But this might still be best

Zer0xFF:
so to confirm, my phone atleast honor 10 can run Kingdom Hearts with multipler of 3x, 5x (non power of 2).

that aside, PR should be ready, tested on OSX, Win32, Android.

Margen67:
Can you make the resolution scale based on the size of the window like in Dolphin, or is that outside the scope of this PR?

Zer0xFF:
@Margen67 at the moment, my time is fairly limited, so while it shouldn't be too hard to implement, I won't be looking into it, in the scope of this PR.

jpd002:
Out of curiosity... how does it work in Dolphin? Does the multiplier change to make the buffer size match the closest window size?

jpd002:
Added the iOS UI (and some more).

I don't know how it worked, but seems like I was able to push in your fork... Anyways, gonna merge once it compiles.

PR #742, opened 2019-01-29
Rayman Revolution

bigianb:
Rayman Revolution now works (except some graphical glitches that look like path 3 masking things).
Changes are:

1. Implement QueryTotalFreeMemSize
2. sif check init returns true (ubisound.irx checks for this and loops until it is true)
3. Stack space is initialised to 0xFF (this seems to be what the bios does unless a flag is passed to explicitly disable it). There is a bug in ubisound.irx where an uninitialised variable is used on the stack and the initial value is assumed to be non-zero.
4. Need to fully initialise the sif queuedata structure as previously we were assuming it was zero initialised (which is a bug because it is just stack space and so will be uninitialised depending on the previous call activity).

### DIFF
--- a/Source/iop/IopBios.cpp
+++ b/Source/iop/IopBios.cpp
@@ -869,8 +869,8 @@ int32 CIopBios::GetCurrentThreadIdRaw() const
 uint32 CIopBios::CreateThread(uint32 threadProc, uint32 priority, uint32 stackSize, uint32 optionData, uint32 attributes)
 {
 #ifdef _DEBUG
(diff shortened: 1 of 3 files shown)


bigianb:
image

bigianb:
The problem now is that you can fall though the floor when playing. I remember pcsx2 had that issue in the early days ... so I'm guessing that could be to do with math rounding.

rcaridade145:
@bigianb  this https://github.com/PCSX2/pcsx2/issues/2452 what the pcxs2 issue :)

bigianb:
That would be the ideal case ... the question is how to find it. It’s not a hard error we can trap so it’s a case of unpicking what I guess is a physics library and tracing though it to find the collision detection routine.

PR #749, opened 2019-03-30
Qt UI cover list

Zer0xFF:
as mentioned on discord, most of this was done in Jan but was just busy to push it.

~~rebased into master, and it's should be a completed implementation but I'm marking it as WIP until I review the changes again (tomorrow) and fully test it out.~~

Edit: tested on OSX/Win32
Edit2: issue on Win32 handling paths, causes same game to show up twice, e.g as c:\iso\game.iso and as c:\iso/game.iso
ummm... easiest solution is to do replace all `/` `\` on database access on `_WIN32`

edit3:
```c++
std::string GetPlatformPathString(const boost::filesystem::path& path)
{
	std::string s_path = Framework::PathUtils::GetNativeStringFromPath(path);

#ifdef _WIN32
	std::replace(s_path.begin(), s_path.end(), '/', '\\');
#endif

	 return s_path;
}
```
replaced every instance of `Framework::PathUtils::GetNativeStringFromPath(path);` with the above call and that fixed duplicate game entry... should this perhaps be part of `GetNativeStringFromPath()` itself?

Screenshot 2019-03-30 at 00 25 10


jpd002:
Really nice!

Only thing that's bugging me is the path format thing. AFAIK, the path slashes should not matter since we don't do any manipulation on them anywhere.

I'll do a quick test on my end to see if I can find out why that happens.

Zer0xFF:
a had a quick look at this again, and apparently `boost::filesystem::path` has function `make_preferred()`which converts the slashes to native win32, but does nothing on unix system, but replaces slashes on windows (exactly as the function I've included), however it modifies the calling object itself, so if you prefer to use that, we'd need to modify `GetNativeStringFromPath(..)` to either remove `const` from it or create a new path object to call that on.

jpd002:
Getting some errors while running this on my machine. Namely, this:

`ASSERT failure in QCoreApplication::sendEvent: "Cannot send events to objects owned by a different thread. Current thread 0x0x270bd463020. Receiver 'qt_scrollarea_viewport' (of type 'QWidget') was created in thread 0x0x270b67f0550", file kernel\qcoreapplication.cpp, line 577
Debug Error!
`

Zer0xFF:
Ummmm... Interesting, I take it you're testing this with debug build? Since I didn't get it before, but I guess I know the line causing it, it's probably caused the scroll call, it's under the force redraw comment.

I'm not home now, I'll check it tomorrow.

jpd002:
Great! Thanks!

Zer0xFF:
commit pushed, using qt signal/slots we can avoid direct UI access from a different thread.

do note that Qt also supports a form of lambda for the connect call, however that for some reason still relies on the calling thread.

```
	connect(this, &BootableListDialog::AsyncUpdateCoverDisplay, [this](){ UpdateCoverDisplay(); });
```

Source/ui_qt/bootablelistdialog.cpp
```diff
@@ -0,0 +1,163 @@
...
+	}
+}
+
+void BootableListDialog::on_listView_doubleClicked(const QModelIndex& index)
+{
+	bootable = model->GetBootable(index);
+	accept();
+}
+
+void BootableListDialog::on_refresh_button_clicked()
```
Zer0xFF: should make this entire function async to avoid UI lockup

Zer0xFF:
just to note, I was just on linux to test the linux deployment PR, and i tested this branch, and i realised that we dont have unix based http implementation, with that said, this PR still works on linux, it just can't make any requests for now.

jpd002:
Seems like the slash thing is a Qt related problem. The file dialog always returns paths that have '/' even on Windows. They have this function that can be used to convert the slashes to the native format: https://doc.qt.io/qt-5/qdir.html#toNativeSeparators.

[This](https://github.com/jpd002/Play-/commit/11f5bf314cf3d609973bb347ad5d9767c5e7c463) would be a proper fix for the problem.

Still getting some random crashes, I need to find out why they happen. Also, I think we're missing a vertical scroll bar.

jpd002:
And about the Unix based HTTP client, I've got WIP code here: https://github.com/jpd002/Play--Framework/commits/httpclient_curl. I think it works fine, I just need to merge it in master.

Zer0xFF:
I hid the scroll bar by default, as its taken into account on windows, since its always on, but on osx its hidden by default and become visible during scroll because of that it doesnt occupy a space on osx.

(I'm trying to figure out how to center the covers, to make it less apparent as below, do note that given it more space does make it look normal)

image

more space
image

Zer0xFF:
hopefully last force push >_<

jpd002:
I can't make it crash anymore. Might be related to updating to Qt 5.12.3 on my machine. Going to merge this since it's working good.

One thing I'm worrying about is that there's many ways to start games now on the desktop version. Would be nice to unify that at some point, maybe by having the cover list show by default in place of the emulator's viewport and maybe having a single menu item to start a game from an arbitrary path. "Boot from cdrom0" could be easily replaced by having the "last played" view. Files from S3 could also show in the cover list, so, we wouldn't be limited to files that are on the user's drive. Also, we could have a special view for physical drives to let user boot from real discs.

Zer0xFF:
if its Qt version related, then that might still be half an issue on linux build, linux travis uses 5.6 (win32/osx use 5.12)

and we should probably officially move support to 5.12 LTS, since Qt 5.6 support has officially ended 2 months ago.

I could look into having it overlay the emu's view, but you'd still need a menu item to toggle it on/off, so that doesn't really change anything?

PR #753, opened 2019-05-10
fix cover art url string formatting

Zer0xFF:
extra / causes cdn to create a unique cache to that url, which we over TGDB dont actually clear when new image is uploaded


jpd002:
Thanks!

Issue #696, opened 2018-05-19
Play Ubuntu Arm64 Building Error


natehaxx:
`-- Build files have been written to: /home/switch/Play-Build/Play/build
switch@switch:~/Play-Build/Play/build$ cmake --build .
[  0%] Built target glew_s
[  1%] Built target bz2
[  3%] Built target Boost_system
[  5%] Built target Boost_filesystem
[ 18%] Built target Framework
[ 18%] Built target Framework_Http
[ 20%] Built target Framework_OpenAl
[ 28%] Built target CodeGen
[ 66%] Built target PlayCore
[ 67%] Built target Framework_OpenGl
[ 69%] Built target gsh_opengl
[ 69%] Building CXX object Source/ui_unix/CMakeFiles/Play.dir/GSH_OpenGLQt.cpp.o
In file included from /usr/include/aarch64-linux-gnu/qt5/QtGui/QOpenGLContext:1:0,
                 from /home/switch/Play-Build/Play/Source/ui_unix/GSH_OpenGLQt.cpp:3:
/usr/include/aarch64-linux-gnu/qt5/QtGui/qopenglcontext.h:55:2: warning: #warning qopenglfunctions.h is not compatible with GLEW, GLEW defines will be undefined [-Wcpp]
 #warning qopenglfunctions.h is not compatible with GLEW, GLEW defines will be undefined
  ^~~~~~~
/usr/include/aarch64-linux-gnu/qt5/QtGui/qopenglcontext.h:56:2: warning: #warning To use GLEW with Qt, do not include <qopengl.h> or <QOpenGLFunctions> after glew.h [-Wcpp]
 #warning To use GLEW with Qt, do not include <qopengl.h> or <QOpenGLFunctions> after glew.h
  ^~~~~~~
In file included from /usr/include/aarch64-linux-gnu/qt5/QtGui/qopengl.h:105:0,
                 from /usr/include/aarch64-linux-gnu/qt5/QtGui/qopenglcontext.h:60,
                 from /usr/include/aarch64-linux-gnu/qt5/QtGui/QOpenGLContext:1,
                 from /home/switch/Play-Build/Play/Source/ui_unix/GSH_OpenGLQt.cpp:3:
/usr/include/GLES3/gl31.h:528:60: error: ‘void __glewActiveTexture(GLenum)’ redeclared as different kind of symbol
 GL_APICALL void GL_APIENTRY glActiveTexture (GLenum texture);
                                                            ^
[...]

PR #754, opened 2019-05-16
Build on aarch64 linux

Specht2010:
Regarding issue #696 :

To build and run Play on L4T Ubuntu while not affecting other platforms I've made some changes in the [Dependencies](https://github.com/Specht2010/Play-Dependencies/commit/10d620244920e9a39a700f6c3e8f43700263372a) and [Framework](https://github.com/Specht2010/Play--Framework/commit/4505c2443b7a7f2fe6a611534508d0ec823e5d60) submodules as well as the changes in this PR.  The submodule changes are required to build so I've linked those commits.  If these changes are fine I can send a PR for those as well.

For the Dependencies change I added `TARGET_PLATFORM_UNIX_AARCH64` to Header.cmake using `${CMAKE_SYSTEM_PROCESSOR} MATCHES "^(aarch64.*|AARCH64.*)"` to set it to true in the Unix section. This will happen after `TARGET_PLATFORM_UNIX` is set to true.  This change isn't required if I use the `CMAKE_SYSTEM_PROCESSOR` line in the other makefiles, though I found this to be cleaner.

In Framework I excluded `TARGET_PLATFORM_UNIX_AARCH64` from using glew project libs in FrameworkOpenGL/CMakelists.txt.  Also, in include/opengl/OpenGlDef.h I added `#elif defined(__arch64__) && defined(__linux__)` above the linux #elif includes to include GLES3 rather than glew.

Finally, in Play- I enabled gles compatibility under this platform, excluded glew from the ui_qt project libs for this platform, and excluded aarch64 linux where Apple had already been excluded from using glew libs in GSH_OpenGLQt.cpp

On my Switch I'm able to build and run games with these changes using the original instructions in the readme and I believe the changes I've made shouldn't affect building for other platforms.  I've tested Metal Gear 2 in the 2nd MGS3 Subsistence disc and Devil May Cry 3 and they run as I'd expected them to with graphics and audio.


Zer0xFF:
ummm this comment is more targeted at how we've included `Header.cmake` than this PR itself, but wonder about the use of `TARGET_PLATFORM_UNIX_AARCH64` in Framework repo, since `Header.cmake` is not explicitly included there, so it will work correctly in our case since we're referring to `Framework` cmakes after including `Header.cmake` in this repo but could/will fail if `FrameworkOpenGl` was included directly into other projects.

jpd002:
Thanks! Looks pretty solid. There's what @Zer0xFF mentioned, but it doesn't need to be taken care of now.

Is there a way to build for Linux AArch64 on TravisCI (or other)? I would love to have some CI on this to make sure it doesn't break.

Specht2010:
Sure, I can try to get TravisCI set up with this.  I've seen some projects using aarch64 for Android with it so this might be fine.  Though I may be away from home for the next few days so I'll get back to this when I can.

Zer0xFF:
Android tool contains full tool chain for cross compiling, so it's a different case.

I'm 100% sure it can be done, just not as straightforward as android cross compiling is.

jpd002:
Yeah, I came to the same conclusion while doing some research on that. Going to merge this for now, the CI for AArch64 Linux can be done later.

jpd002:
Oops, didn't mean to close this without merging.

PR #756, opened 2019-05-17
Gamepad profile

Zer0xFF:
~~this PR would require these changes https://github.com/jpd002/Play--Framework/commit/3ab44ec526647666ebdd73f95dcce8ed6da9f8e9 in Framework repo.~~

~~however, I'm not sure how you feel about those particular changes? more specifically the `Framework` ones? making path editable and Load accessible like that.~~

I also moved gamepad config to their own xml, i know a while back, you didnt really want that, under slightly different context (aka gamepad1 vs gamepad2), however in this case, i think the config file will be overflooded and harder to manage if we keep all the profiles in the same `config.xml` file

TODO:

~~- [ ] add few error msg dialogs~~
~~- [ ] clean up commits, use descriptive names~~
- [x] Address the TODO comments

### DIFF
--- a/Source/ui_qt/PreferenceDefs.h
+++ b/Source/ui_qt/PreferenceDefs.h
@@ -3,3 +3,4 @@
 #define PREFERENCE_AUDIO_ENABLEOUTPUT "audio.enableoutput"
 #define PREF_UI_PAUSEWHENFOCUSLOST "ui.pausewhenfocuslost"
 #define PREF_VIDEO_GS_HANDLER "video.gshandler"
+#define PREF_INPUT_PAD1_PROFILE "input.pad1.profile"
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -198,6 +198,8 @@ set(COMMON_SRC_FILES
 	FpUtils.h
 	FrameDump.cpp
 	FrameDump.h
+	InputConfig.cpp
+	InputConfig.h
 	GenericMipsExecutor.h
 	gs/GsCachedArea.cpp
 	gs/GsCachedArea.h
--- a/Source/InputConfig.cpp
+++ b/Source/InputConfig.cpp
@@ -1,0 +1,46 @@
+#include "AppConfig.h"
+#include "InputConfig.h"
+#include "PathUtils.h"
+
+#define PROFILE_PATH ("inputprofiles")
+
+CInputConfig::CInputConfig(const Framework::CConfig::PathType& path)
+    : CConfig(path)
+{
+}
+
+bool CInputConfig::IsValidProfileName(std::string name)
+{
+	static const std::string valid_chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-";
+	for(auto c : name)
+	{
+		if(valid_chars.find(c) == std::string::npos)
+		{
+			return false;
+		}
+	}
+	return true;
+}
+
+std::unique_ptr<CInputConfig> CInputConfig::LoadProfile(std::string name)
+{
+	auto path = GetProfilePath() / name;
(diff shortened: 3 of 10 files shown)


jpd002:
At first glance, this looks good and I'm realizing this is needed at least for testing multiple pad types. I'm going to test it locally and give you feedback.

jpd002:
The changes in Framework seem unnecessary. It's possible to construct a CConfig object with a path. If you need to replace a config profile, it would just be a matter of creating a new CConfig object and replacing the previous one with the new one. 

No need to "recycle" the `m_config` object. Not sure if it's possible, but it might be possible to make it movable, but otherwise, we could just make `m_config` a `unique_ptr<CConfig>` in `CInputBindingManager`.

Source/GamePadConfig.cpp
```diff
@@ -0,0 +1,41 @@
+#include "AppConfig.h"
+#include "GamePadConfig.h"
+#include "PathUtils.h"
+
+#define PROFILE_PATH (L"Profiles")
```
jpd002: Please make this `profiles` to be consistent with everything else that's already in the Data Files folder.

Zer0xFF: can do that.

Zer0xFF:
> No need to "recycle" the m_config object.

ya I realise I now complicated this, since I was fixated on this :/

Zer0xFF:
rebased, and made `m_config` `unique_ptr`, its still a WIP, more specifically UI stuff need to be cleaned up a little bit (as per commented noted)

Source/ui_qt/PreferenceDefs.h
```diff
@@ -3,3 +3,4 @@
 #define PREFERENCE_AUDIO_ENABLEOUTPUT "audio.enableoutput"
 #define PREF_UI_PAUSEWHENFOCUSLOST "ui.pausewhenfocuslost"
 #define PREF_VIDEO_GS_HANDLER "video.gshandler"
+#define PREF_UI_GAMEPAD1_PROFILE "ui.gamepad1.profile"
```
jpd002: Let's rename these to `PREF_INPUT_PAD1_PROFILE` and `input.pad1.profile`.

Source/GamePadConfig.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "PathUtils.h"
+
+#define PROFILE_PATH ("profiles")
+
+CGamePadConfig::CGamePadConfig(const Framework::CConfig::PathType& path)
+    : CConfig(path)
+{
+}
+
+bool CGamePadConfig::isValidProfileName(std::string name)
```
jpd002: Please rename this to `IsValidProfileName` (capital i).

Source/GamePadConfig.cpp
```diff
@@ -0,0 +1,46 @@
+#include "AppConfig.h"
+#include "GamePadConfig.h"
+#include "PathUtils.h"
+
+#define PROFILE_PATH ("profiles")
```
jpd002: Please rename this to `inputprofiles`. `profiles` is too generic.

jpd002:
So, few comments/questions regarding this:

- We should rename `GamePadConfig` to `InputConfig`. I'd like to prevent adding new names for stuff, right now we have `PH_GenericInput` that has `InputBindingManager` and it would be more consistent if it had `InputConfig`s instead of `GamePadConfig`s. I've made requests for some other minor naming changes, just to prevent changing setting/folder names in the future.

- We should probably save the config when changing the combo box selection. Right now you loose everything if you do.

- Just to make sure (but I think we agree here) the intent is to have a profile per input. Right now you have preference "input.pad1.profile", but we will add "input.pad2.profile", "input.pad3.profile", etc. later.

I'll merge after these changes are made.

jpd002:
Great work! Thanks a bunch!

PR #757, opened 2019-05-18
Linux deployment

### LINKED ISSUE #635: linux builds
is there a way to add linux to the weekly build roster, it doesn't have to be fancy: just binaries in a tar.gz like  citra does would be awesome.

Zer0xFF:
an amalgamation of #628 (app image) #543 (install)

for testing initially built on Ubuntu 16.04 (using Qt 5.12):
resulting package tested on Ubuntu 16.04, arch linux (PS4) (dont know how versioning work on arch XD)

Note: I can't test travis deployment, but assuming its straight forward, that should work just fine as committed

resolves #635 closes #628 closes #543

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -77,8 +77,16 @@ travis_script()
             if [ "$CXX" = "g++" ]; then export CXX="g++-5" CC="gcc-5"; fi
             export PATH=/opt/cmake-3.8.1-Linux-x86_64/bin/:$PATH
             source /opt/qt56/bin/qt56-env.sh || true
-            cmake .. -G"$BUILD_TYPE" -DCMAKE_PREFIX_PATH=/opt/qt56/;
+            cmake .. -G"$BUILD_TYPE" -DCMAKE_PREFIX_PATH=/opt/qt56/ -DCMAKE_INSTALL_PREFIX=./appdir/usr;
             cmake --build .
+            cmake --build . --target install
+
+            # AppImage Creation
+            wget -c "https://github.com/probonopd/linuxdeployqt/releases/download/continuous/linuxdeployqt-continuous-x86_64.AppImage"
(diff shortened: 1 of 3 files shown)


jpd002:
I'm not familiar with this, but I guess once this is merged, we will get Play.AppImage files on the S3 bucket where other builds are stored? I guess it would make sense to add links to this file on the website then?

Zer0xFF:
yes, and yes. 1 thing to note that create AppImage will be named `Play!-{SHORT_HASH}-x86_64.AppImage` (incase thats relevant to linking it), we could always rename it if it does matter.

The AppImage should contain all the needed libraries to run `Play!` which means that image will run across many if not all linux based OSs.

jpd002:
Sounds pretty cool! Thanks!

hcorion:
Thanks for cleaning this up and and getting it merged, fantastic to see!

.travis.sh
```diff
@@ -111,6 +119,9 @@ travis_before_deploy()
     if [ -z "$ANDROID_KEYSTORE_PASS" ]; then
         return
     fi;
+    if [ "$TARGET_OS" = "Linux" ]; then
+        cp ../../build_cmake/build/Play*.AppImage .
```
Zer0xFF: @jpd002 it'd probably easier for you to test fix instead of, since PRs dont deploy, but looking at the logs, i think path should be `../../build/Play*.AppImage`?

Zer0xFF:
@hcorion well thanks for getting it back into my head, after i dismissed it as an option.

.travis.sh
jpd002: I've changed it, working good now.

Zer0xFF: that's great :)
I tested the uploaded image and it works great on Ubuntu 16.04, will test on arch later and will report if anything worth mentioning.

PR #761, opened 2019-05-24
Few android changes

Zer0xFF:


jpd002:
Regarding that last commit (the fps counter padding), it seems to move the counter right? It seems the intention was to move it down?

Could we use [this](https://developer.android.com/guide/topics/display-cutout) to handle the notch problem? It's nitpicking, but I'd like to keep the fps counter on the edge of the screen and not have a gap.

The three first commits are good, can you make a PR with those only? Thanks!

Zer0xFF:
It's suppose to move it the the right.

We could, I already have a local commit that does this, the only problem was, that the screen "seems" off centred due to the notch. It's not that big of a deal regardless. 

I'll redo force this PR without that commit and push those changes to a different PR so you can have a look.

jpd002:
Thank you very much!

Zer0xFF:
any time :)

Zer0xFF:
to mention the display-cutout, im not sure how it looks on other devices, but the link you posted doesnt actually fix the problem.

you see as is right now, the screen ends before the notch and there is a virtual rounded border
image
note how its rounded, even though its not the true edge of the screen.

if we enable the notch option, the entire screen is used, and the fps will be cut this time by the physical edge
image

as i said, its not that big of a deal, just something cosmetic i noticed.

PR #762, opened 2019-06-04
Boost removal

Zer0xFF:
- [x] remove boost::scoped_ptr (merged)
- [x] remove boost::lexicasl_cast (merged)
- [x] boost::noncopyable (merged)
- [x] remove boost::bind (merged) 
- [x] remove boost::replace_all (merged)
- [x] remove boost::erase_all (merged)
- [x] remove boost::is_any_of (merged)
- [x] remove boost::split (merged)
~- [ ] remove boost::any~ 
~- [ ] remove boost::any_cast~ boost::any_* will be dealt with in a separate PR.
- [x] remove boost::filesystem
- [x] remove boost::signals2::* (merged) 
  - [x] boost::signals2::signal (merged) 
  - [x] boost::signals2::trackable (merged)
  - [x] boost::signals2::connection (merged)
- [x] boost::flyweight

~~though these commit touch `Win32_UI`, since its to be removed, I haven't build it to test their validity,
however, as of f29638a the changes are simple enough to presume they're valid.~~

You'd need this PR https://github.com/jpd002/Play--Framework/pull/18 and https://github.com/jpd002/Play-Dependencies/pull/17

Note to self: c1d1235 is probably an issue with my own setup, need to check it out.

### DIFF
--- a/Source/ui_qt/win32/DebugSupport/ELFSymbolView.cpp
+++ b/Source/ui_qt/win32/DebugSupport/ELFSymbolView.cpp
@@ -282,7 +282,7 @@ void CELFSymbolView::GetItemInfo(LVITEM* outItem) const
 {
 	if(outItem->iItem >= m_items.size()) return;
 	const ITEM& item(m_items[outItem->iItem]);
-	const PooledString* itemText(NULL);
+	const std::tstring* itemText(NULL);
 	switch(outItem->iSubItem)
 	{
 	case 0:
@@ -304,5 +304,5 @@ void CELFSymbolView::GetItemInfo(LVITEM* outItem) const
 		itemText = &item.section;
 		break;
 	}
-	outItem->pszText = const_cast<TCHAR*>(static_cast<const std::tstring&>(*itemText).c_str());
+	outItem->pszText = const_cast<TCHAR*>(itemText->c_str());
 }
--- a/Source/ui_qt/win32/DebugSupport/ELFSymbolView.h
+++ b/Source/ui_qt/win32/DebugSupport/ELFSymbolView.h
@@ -3,7 +3,9 @@
 #include "win32/Window.h"
 #include "win32/ListView.h"
 #include "ELF.h"
-#include <boost/flyweight.hpp>
+
+#include <vector>
+#include <string>
 
 class CELFSymbolView : public Framework::Win32::CWindow
 {
@@ -12,16 +14,14 @@ public:
 	virtual ~CELFSymbolView();
 
 private:
-	typedef boost::flyweight<std::tstring> PooledString;
-
 	struct ITEM
 	{
-		PooledString name;
-		PooledString address;
-		PooledString size;
-		PooledString type;
-		PooledString binding;
-		PooledString section;
+		std::tstring name;
+		std::tstring address;
+		std::tstring size;
+		std::tstring type;
+		std::tstring binding;
+		std::tstring section;
 	};
 
 	typedef std::vector<ITEM> ItemArray;


Source/iop/Iop_Spu2.h
```diff
@@ -1,13 +1,13 @@
...
 
-#include <boost/utility.hpp>
+#include "Noncopyable.h"
 #include <functional>
 #include "Iop_Spu2_Core.h"
 
 namespace Iop
 {
-	class CSpu2 : boost::noncopyable
+	class CSpu2 : Framework::CNoncopyable
```
jpd002: The noncopyable thing was useful back when C++11 was not widely available. IMO, now, the proper way of carrying this intention would be to mark the copy constructor and copy assignment operator as deleted. ex: `CSpu2(const CSpu2&) = delete;`

No need for a Noncopyable utility in Framework.

Source/gs/GSHandler.cpp
```diff
@@ -552,7 +552,7 @@ void CGSHandler::WriteRegisterImpl(uint8 nRegister, uint64 nData)
 
 void CGSHandler::FeedImageDataImpl(const void* pData, uint32 nLength)
 {
-	boost::scoped_array<const uint8> dataPtr(reinterpret_cast<const uint8*>(pData));
+	std::unique_ptr<const uint8> dataPtr(reinterpret_cast<const uint8*>(pData));
```
jpd002: Not sure it's going to work on all compilers, but using a `std::unique_ptr<const uint8[]>` would seem like a better idea since we're using `new []` in `FeedImageData` and we want to make sure `delete []` is called.

Source/iop/Iop_Spu2.h
Zer0xFF: ah, ill change it so

Source/gs/GSHandler.cpp
Zer0xFF: 👍  just reading [cppreference ](https://en.cppreference.com/w/cpp/memory/default_delete) right now, it seems to be a c++17 feature, which means we'd need to move standard up (but as you mentioned, pending compiler support, cough apple cough)

Zer0xFF:
Comments:

`boost::any`, `boost::any_cast`
both these objects are used in `MediaCenter`, which didnt get any CMake loving.
however, since its a windows only project, msvc now has `std::any` and `std::any_cast` , should we just fast forward this to c++17? (note: im struggling trying to get the build to work, since we need to change to toolchain to v141 for c++17, but we also need to change boost packages to v141, which doesnt have 1.60 version etc etc etc, either-way, its best to test this once the other changes are made, since  can just delete boost at that stage)

`boost::flyweight`: reading the description on boost is scary XD
ill figure something out when i get to it

last `boost::signals2::* `
I already have an implementation for `signal` which seems to work fine.
`connection` as per our use (`*.disconnect()`) should be easy to add as well
but I still can't understand what `trackable` is suppose to do, can you elaborate on that?


Also, ill rebase once i get through everything (unless you really want to merge the current changes :P)

uyjulian:
Apple LLVM version 10.0.1 (clang-1001.0.46.4) supports `-std=c++17`.

jpd002:
The problem is Android. I think we need a somewhat recent version of CMake to add the C++17 flag to the compiler's command line, but the Android Gradle plugin doesn't like that version. I've tried a month ago or so, might be working now.

Source/gs/GSHandler.cpp
jpd002: Oh, I didn't notice the C++17 part... Well, there's another way to handle this case. I'll make the change and push it in master.

jpd002:
I think we can leave MediaCenter aside, but it would make sense to update to C++17 for that project and use std::any.

`trackable` enables auto disconnection of objects that are connected to signals. Once the object is destroyed, it will disconnect itself from the signal if it was previously connected. It seems its usage is not [recommended](https://www.boost.org/doc/libs/1_70_0/doc/html/boost/signals2/trackable.html), so I guess we should get rid of it (and I know it doesn't work well with std::bind and lambdas, only with boost::bind).

Also, are you rewriting boost::signals? I don't think we'll be able to get rid of boost completely, so, we might as well keep this one?

I don't think the flyweight is essential in ELFSymbolView. I think we could just replace it with std::tstring and it should be fine. But this is in the Win32 debug support and since we're probably not be able to get rid of boost completely, it might be ok to just leave it there.

Just to make sure I understand well, the scope of this PR is to remove boost usage to be able to compile for targets that are not be able to compile boost? Like I mentioned, I'm not sure it's possible to get rid of it completely (mainly because of boost::filesystem).

Zer0xFF:
```
The problem is Android [....] cmake
```
just top of my head, I believe newer version of cmake was introduced a while back.

```
I think we can leave MediaCenter aside, but it would make sense to update to C++17 for that project and use std::any.
```
cool, so pending work completion I'd look into it. (if all goes well, it should be a simple switch)

`trackable`, but thats the thing, object being the class that inherits `trackable` doesn't itself have any "connection". you can however, declare a `connection` object in the class, but that object has no understanding of the class that created it.

```
Also, are you rewriting boost::signals? I don't think we'll be able to get rid of boost completely, so, we might as well keep this one?
```
yes, I already have a working implementation, though with limited testing. (ill push it out as WIP for `Framework` PR if you want to have a look at it)

`flyweight`, if it doesn't encore too much overhead, I might just do that XD

```
Just to make sure I understand well, the scope of this PR is to remove boost usage to be able to compile for targets that are not be able to compile boost?
```
essentially, for devices with limited libc/c++ support (still no guarantee that will help, but only one way to find out), but I recall it being a long term goal, and so the 2 reasons came together.
ofc there is also the maintenance advantage with less dependencies (thus potential faster build time)

`boost::filesystem`, yes this will be an issue, I found few filesystem c++11 alts (based on c++17 filesystem implementation), however, initial test worked on OSX, but not on windows, however, I almost never work linearly, so i was pushing signal changes, filesystem changes, and bind changes all at once, and that was error'ing something crazy that made no sense, so I've gone back to the start and have been changing and testing the changes 1 by 1 (as they appear in the commit), but I pushed it back to the end of the queue to see if the issue is with the library or something else i was doing.
(yes we're replacing one dependencies with another, but the hope is that this one is lighter, header only).

and so by that, removing `boost` completely should be possible.

Zer0xFF:
rebased on master.

so I've said earlier that I've got a WIP signal replacement, but there is a small issue, disconnecting callback when object is deleted, its not a massive issue as it can be solved as I mention below, its just that it won't be a simple replacement drop in like with the previous changes.

in short, the callback could be invalid by the time the signal is called, there are 2 ways solve this problem:
1) to pass a weak reference to the object to check before the callback is made, however, to do so, those object have to be created as `shared_ptr` (in order to get a `weak_ptr` from it).
2) we manage the connection with `unique_ptr`, each connection to a signal returns a connection, we can add a destructor to it, so it will be deleted with end of scope (or with `unique_ptr`, until object owning the ptr is deleted)

Source/ui_qt/win32/DebugSupport/FrameDebugger/Vu1ProgramView.cpp
```diff
@@ -19,8 +19,8 @@ CVu1ProgramView::CVu1ProgramView(HWND parent, const RECT& rect, CVu1Vm& virtualM
 	Create(0, Framework::Win32::CDefaultWndClass::GetName(), NULL, WNDSTYLE, rect, parent, NULL);
 	SetClassPtr();
 
-	virtualMachine.OnMachineStateChange.connect(boost::bind(&CVu1ProgramView::OnMachineStateChange, this));
-	virtualMachine.OnRunningStateChange.connect(boost::bind(&CVu1ProgramView::OnRunningStateChange, this));
+	m_OnMachineStateChangeConnectionPtr = virtualMachine.OnMachineStateChange.connect(std::bind(&CVu1ProgramView::OnMachineStateChange, this));
+	m_OnRunningStateChangeConnectionPtr = virtualMachine.OnRunningStateChange.connect(std::bind(&CVu1ProgramView::OnRunningStateChange, this));
```
Zer0xFF: drop Ptr postfix

Zer0xFF:
and with the latest commit, boost has been expelled from the project ✌️ 

Note:
as far as a81416f have been tested on OSX, 
049325d on windows

I'll Linux, Android, iOS haven't been tested yet.

Source/ee/DMAC.cpp
```diff
@@ -7,6 +8,7 @@
 #include "../COP_SCU.h"
 #include "placeholder_def.h"
 
+
```
Zer0xFF: remove

tools/PsfPlayer/Source/psp/PspBios.cpp
```diff
@@ -181,7 +181,7 @@ CBios::CBios(CMIPS& cpu, uint8* ram, uint32 ramSize)
     , m_threadFinishAddress(0)
     , m_idleFunctionAddress(0)
 {
-	BOOST_STATIC_ASSERT(BIOS_CALCULATED_END <= CBios::CONTROL_BLOCK_END);
+	assert(BIOS_CALCULATED_END <= CBios::CONTROL_BLOCK_END);
```
jpd002: This should be a `static_assert`.

Zer0xFF:
rebased. updated `StringUtils` and `static_asserts`

Source/ui_qt/win32/DebugSupport/FunctionsView.cpp
```diff
@@ -225,11 +225,6 @@ uint32 CFunctionsView::GetFunctionGroupId(uint32 address)
 
 void CFunctionsView::SetContext(CMIPS* context, CBiosDebugInfoProvider* biosDebugInfoProvider)
 {
-	if(m_functionTagsChangeConnection.connected())
-	{
-		m_functionTagsChangeConnection.disconnect();
-	}
-
```
Zer0xFF:
```suggestion
	m_functionTagsChangeConnection.reset()
```

Zer0xFF:
as far as I can tell, everything here beside `boost::filesystem` was merged, and since that's already in the work by @jpd002 I'd remove from this PR, so the only remaining thing is `boost::flyweight`.

if this is not needed please close.

jpd002:
I've removed flyweight and the remains of boost::filesystem. boost is now officially out! 

Thank you very much @Zer0xFF for making this happen!

Zer0xFF:
😁 glad to be part of the change

PR #764, opened 2019-06-25
Libretro port

### LINKED ISSUE #9: Libretro port
Unlike other PS2 emulators, Play! is much more portable. Have you thought about a port to Libretro? It would give much more attention to the project, and would allow it to be ported to a wide variety of platforms. 

libretro.com

Zer0xFF:
Note:
I haven't tested `-fPIC` on regular builds yet(will do), however, i believe `PIC` is on by default for Android (& OSX), and doesn't apply to Windows build. (thus its safe to assume nothing will break because of it?)

Retroarch controls the OpenGL Context, so running GS Async isn't an option, which is why I had to kill the thread, to process those calls on thread/call provided by retroarch.

~~26643e4 might be an overkill? but i figured since mailbox isn't getting processed all the time, it could end up locking other part of the emu until its processed, but then when the mutex guard is released, it will be for all calls at once, instead after their own call has been processed.~~ removed since it was an over kill, this only ever need to release the last call, since we can't add more messages in such case anyway.

~~edit: just noticed I havent added Elf loading yet~~

close #9 on merge

~~Note to self:
split f284e6521b72c027252b940f9d5c869f66e2fe71 into 2 commits
merge f0838a9b932bc1a0945c3ee637cce4b36a71d8f3 && a0ac902a63004bb35a11a5448a1ec1f2f104d977~~


Zer0xFF:
change to WIP, since last commit 5be3bac  is not compatible with master.
I'll be look at this hopefully around next week to the best way to do this (perhaps a toggle option for different audio types)

Source/ui_libretro/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
+cmake_minimum_required(VERSION 3.5)
+
+set(CMAKE_MODULE_PATH
+	${CMAKE_CURRENT_SOURCE_DIR}/../../../Dependencies/cmake-modules
```
Zer0xFF:
```suggestion
	${CMAKE_CURRENT_SOURCE_DIR}/../../deps/Dependencies/cmake-modules
```

appveyor_build.cmd
```diff
@@ -20,6 +21,7 @@ if "%BUILD_PLAY%" == "ON" (
 	cd ..
 	mkdir %REPO_COMMIT_SHORT%
 	move Play-Build\Play\installer_win32\*.exe %REPO_COMMIT_SHORT%
+	move Play-Build\Play\build\Source\ui_libretro\Release\play_libretro.dll %REPO_COMMIT_SHORT%
```
Zer0xFF:
```suggestion
	move build\Source\ui_libretro\Release\play_libretro.dll %REPO_COMMIT_SHORT%
```

jpd002:
My only major blocker for merging this are the changes in `CMailBox`.

I think `CGSHandler` should be responsible for handling the "no-thread" mode. Dunno if that would be the best way to do it, but there could be a bool passed in the `CGSHandler` constructor that would say if it's threaded or not-threaded.

There could be a wrapper to `SendCall` which would force `waitForCompletion` to be false when we're in no-thread mode and would just forward all the other parameters to `CMailBox::SendCall`. 

`CGSHandler::ProcessSingleFrame` would be only usable in no-thread mode and would execute all the commands stored in `CMailBox` till a flag set in `FlipImpl` is set to true (equivalent to the breakpoint concept you had). That last `SendCall` would have `waitForCompletion` set to true to block the VM thread.

That way, `CMailBox` would be left intact.

Would that make sense?

Zer0xFF:
ya that makes sense, I too didn't really like the way I effectively hacked `CMailBox` to get it to work like this.

Zer0xFF:
made GS in charge of its own threading, and I seems to be working just the same.

P.S I plan on merging the change back to a more appropriate commit, once I'm back home (2days)

jpd002:
Great! Looking forward to merge this in master soon 😄.

jpd002:
It's finally in! 😄 Thank you very much @Zer0xFF !

I've made a small improvement to your `SendGSCall` function: https://github.com/jpd002/Play-/commit/4c13b3bf4f7c8598a33a2676d2a0164eb088ec7b. Could you tell me if it works ok in RetroArch?

Zer0xFF:
🎆 
quick test suggest it works. though I've my doubts if it gets called? since the previous call is overloaded as well?

jpd002:
Great! Thanks a bunch for confirming!

PR #780, opened 2019-07-22
Apply offset to bilinearly filtered texture

Zer0xFF:
related to #773


celerizer:
image
image

---

image
image

The font in Disgaea looks a lot better with this commit

PR #783, opened 2019-07-29
[WIP] opengl text overlay (based on Gs clutcaching branch)

Zer0xFF:
this PR is pretty much complete, unfortunately I'm flying away tomorrow. (and will be busy for the 1st week, and won't have access to a computer the following week).
So im marking it as WIP, since I still need to test this on iOS, Android, Linux, Windows Installer.
Has been tested on OSX, Windows. (though haven't tested last few commits on windows).

main point of this PR is to print profiling info over the game, as oppose a side widget, for one it takes less space, for 2nd since its OpenGL based, It should work on all platform just the same without specific platform tweaks.

from what I can tell, profiling doesn't seem to add much overhead, might be worth adding as a UI option?

Note: I've added frame time + fps, as I thought those can be nice to have during fullscreen profiling.

Edit: this won't build without [freetype dependency](https://github.com/Zer0xFF/Play-Dependencies/tree/freetype) (I'm not opening a PR now, for the above stated reason, but if you feel like moving ahead with this, please feel free)
screenshot:
image


jpd002:
This looks very nice! Unfortunately, I've got some problems with it:
- This will only work with OpenGL. If we add another back-end (ie.: Vulkan) we'll need to maintain the rendering code in there at well. Also, text rendering code is hard to get right, so, it can get messy real quick.
- I don't like that we need to add a dependency on FreeType. I'd like to keep dependencies at a minimum. Also, it seems we need to bundle a font to make it work, which is not super great either.

My idea was just to use a system provided widget and just output the profiling results using a simple `setText` call. I tried looking for a widget that I could overlay on the emulator's output window on Qt, but I didn't find anything. I'm pretty sure there's a way to do it though, but I was ok with having the thing on the side since it's just pretty much only used by me ATM.

As for enabling it all the time, it might be a good idea. It might be interesting to compare frame rate with and without just to be sure it doesn't add overhead.

Zer0xFF:
It depends how you view this, maintain per render implementation (1 now, 2 when vulkan comes around) vs per ui (3 + retroarch, retro can't be bench without a render based overlay stats)

For the dependency, not much that can be done as per this PR (not sure how plausible it is to write our own font to texture implementation, as that will probably more work maintain than an external dependency) , as for the font, I'm using monospaced font, so this is the only consistent option across platforms. 


Iooked at qt, qt by default don't allow widget stacking, you might be able to achieve this by creating your own qt layout.

PR #791, opened 2019-08-29
Update Android Gradle wrapper.

TacoTheDank:
Updated using ```./gradlew wrapper --gradle-version 5.4.1 --distribution-type all``` in Git Bash.


jpd002:
Thanks!

PR #792, opened 2019-08-30
add an IO Port to get the host clock value

bigianb:
Something like this would allow us to profile test scripts without needing to use the CD clock hack (that is a bit useless in any case from the EE as it goes via an RPC channel which takes an indeterminate amount of time)


bigianb:
I've pushed the test code up to a repo:

https://github.com/bigianb/ps2-speedtests

PR #797, opened 2019-09-12
Dark cloud assert fixes

### LINKED ISSUE #795: Dark Cloud trips assert in debug mode on startup
Dark Cloud (USA) is causing the assert on line 75 of Source/ee/VUExecutor.cpp to fail.  This is happening just before the main menu would start after the level-5 logo fades to black.

This does not happen in release mode (I assume because asserts are disabled in release mode).

GregoryLand:
#795 I think I created the warning correctly.  Let me know if there are any issues.


jpd002:
Thanks a bunch!

PR #807, opened 2019-10-02
Dynamically generate changelog from git history

Zer0xFF:


Zer0xFF:
since the html file was removed from repo/windows packaging, this is no longer needed

Issue #806, opened 2019-10-01
Can't detect disc in disc drive


happygreenfrog:
Emulator doesn't seem to acknowledge the existence of the drive at all, even when a disc is inserted. In significantly older versions (I think the initial "stable" 0.30 version is one such example, but I could be remembering wrong?), this was not an issue and the drive worked correctly.
![image](https://user-images.githubusercontent.com/41273098/66003929-08b29400-e46d-11e9-85a4-5c9342469ebb.png)

Windows version is Windows 10 Home, Version 10.0.18362 Build 18362. The Blu-Ray drive is an external LG-CP40NG10 Slim Portable Blu-ray Multi Drive. Game disc pictured in the drive in the screenshot is Mega Man X Command Mission, but I also tested with Mega Man X Collection and Midway Arcade Treasures to no avail.

Curiously, when I pressed "refresh" while the computer itself was still actively loading the disc, it did pause, so maybe the problem is that it can't detect that the disc is inserted, but can see the drive itself...?

jpd002:
@Zer0xFF's PR fixed the UI. Made changes to make physical disc support work again: https://github.com/jpd002/Play-/tree/win32_volume_fix.

happygreenfrog:
The good news is, I tested the automated builds, and now it is indeed detecting the disc drive.

The bad news is... apparently I need to open a new issue related to the disc drive.

...actually, upon further inspection, that issue might have to get split into two separate issues, I can't tell if they're both one larger issue or two unrelated smaller issues... hmm...

...but I'm getting ahead of myself, I think I can mark _this_ issue as closed pretty safely now...

...but before I do that, I do just want to make sure that the automated builds would indeed have the _complete_ disc support fix, or if for one reason or another there's some elements of it that still aren't on the main branch, just in case the new issues are because of something like that instead of something that actually needs to be reported.

jpd002:
Could you try this build: https://s3.us-east-2.amazonaws.com/playbuilds/50fc4e5d/Play-0.30-64.exe? That's the one with the changes from the branch I linked in my previous post.

happygreenfrog:
Have done so now. There are indeed new problems as I thought earlier, the issue is no longer that the drive itself can't be detected so much as it can't tell the disc actually has SYSTEM.CNF and also seems to revert the selection from "disc drive" to "disk image" no matter what you do the moment you open the window.

[...]

PR #808, opened 2019-10-02
allow selecting cdfs as disc type

Zer0xFF:
related to #806 
note, this still doesnt allow games to start from disc on windows, since i think `COpticalMedia` can't create a valid stream from cdvd physical address? (atleast for the games i've tested)


jpd002:
Thanks for this. Fix incoming for reading directly from Windows volumes.

PR #812, opened 2019-10-10
Spublock config

Zer0xFF:


Source/PS2VM.h
```diff
@@ -163,6 +164,7 @@ class CPS2VM : public CVirtualMachine
 
 	int16 m_samples[BLOCK_SIZE * BLOCK_COUNT];
 	int m_currentSpuBlock = 0;
+	int m_SpuBlockCount = BLOCK_COUNT;
```
jpd002: This should be ‘m_spuBlockCount’ to match the casing of other variables.

jpd002:
This is pretty cool!

The setting should be managed by ‘CPS2VM’ to avoid setting the block count everytime we init the SPU handler. In other words, ‘m_spuBlockCount’ should be initialized in ‘CPS2VM’ ctor.

Source/ui_qt/Qt_ui/settingsdialog.ui
```diff
@@ -260,6 +260,41 @@
...
+       <property name="geometry">
+        <rect>
+         <x>20</x>
+         <y>50</y>
+         <width>171</width>
+         <height>16</height>
+        </rect>
+       </property>
+       <property name="text">
+        <string>SPU Block Count*:</string>
```
jpd002: Why is there an asterisk here? Is it a typo?

Zer0xFF: I was going to add an explanation to it, but I wasn't sure what to write.

If you have a suggestion, else I'll remove it

jpd002:
Tested it myself, it's pretty cool.

Instead of calling the setting "SPU Block Count", I think it might be better to call it "Buffer Size" (only in the UI, no need to rename the other stuff). I think it might be a good occasion to give a sensible default to that setting too, something like 100 seems to be good.

Zer0xFF:
ya that "Buffer Size" probably works better, without much explanation.
Changed reduced default to 100 as well.

jpd002:
Thanks a bunch!

Zer0xFF:
👍 sure thing :)

PR #814, opened 2019-10-15
Unix: install icon before .desktop file

guoyunhe:
In Fedora and openSUSE's package build system, they will check icon availability before installing .desktop files. So icons must be installed before .desktop files to avoid errors. So I swapped the order of these two lines.


Zer0xFF:
You did a bit more than swap the order, completely removed it from Unix build XD

You moved it down 2 too many lines.

guoyunhe:
Oops, sorry. I corrected it.

PR #821, opened 2019-10-23
Arm64 travis

Zer0xFF:
this will only build Qt GLES Arm64 version on travis, it will not deploy it.
to deploy it, we need to build `linuxdeployqt` from source, since no prebuilt version currently exist, however thats currently causing me issues, since it builds, but refused to package due to the current glibc version (since they want to "encourage" use of older version for compatibility reasons).

edit:
the header problem stems from clash of GLEW/Qt vs GLES/Qt so there was no configuration that would work for all of them

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -3,10 +3,15 @@
 travis_before_install() 
 {
     if [ "$TARGET_OS" = "Linux" ]; then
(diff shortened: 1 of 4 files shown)


jpd002:
Very cool! Thank you very much!

PR #823, opened 2019-10-30
Use system nlohmann_json library if available

guoyunhe:
This makes packaging on *nix easier.


guoyunhe:
Linux packaging has a minimum size principle. All packages should use system library if possible. It saves resources of the build farm and users' disk space.

q3cpma:
It doesn't add any step once your software is packaged by the distribution. https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies explains it better than we can.

jpd002:
Sounds good, thanks!

PR #829, opened 2019-11-25
IOP memory patch

Zer0xFF:
~~note the 6be36e commit is not needed, its been only added for illustration.~~

this allows uLaunchELF and PSMS-Reloaded to fully boot and likely other home-brews as well, that rely on this code https://github.com/ps2dev/ps2sdk/blob/8b7579979db87ace4b0aa5693a8a560d15224a96/iop/dev9/poweroff/src/poweroff.c#L240


jpd002:
To answer yesterday's question regarding memory accesses below 0x400 in IOP mem:

The HLE kernel adds these in IOP memory. They are handlers used for different purposes such as providing a proper final function when a thread returns from its entry point. They look like this:

```
00000200    24020666    ADDIU          V0, R0, $0666
00000204    0000000C    SYSCALL        
00000208    27BD0010    ADDIU          SP, SP, $0010
0000020C    24020667    ADDIU          V0, R0, $0667
00000210    0000000C    SYSCALL        
00000214    24020668    ADDIU          V0, R0, $0668
00000218    0000000C    SYSCALL        
0000021C    24020669    ADDIU          V0, R0, $0669
00000220    0000000C    SYSCALL        
00000224    2402066A    ADDIU          V0, R0, $066A
00000228    0000000C    SYSCALL        
```

When servicing a SYSCALL, that function checks reads the instruction at EPC:
https://github.com/jpd002/Play-/blob/0363398aad50976fe9f91bd10f27a8aeb50044e2/Source/iop/IopBios.cpp#L2846

This is causing your logs to be triggered.

Regarding the fix for homebrew directly looking at INTC handlers, I think we should reserve a spot in the HLE kernel memory map (https://github.com/jpd002/Play-/blob/0363398aad50976fe9f91bd10f27a8aeb50044e2/Source/iop/IopBios.cpp#L44) and add the needed space for the INTC handler there. It should allow the homebrew to start without breaking anything else. I don't care about poweroff not working properly, so adding a dummy space in there is fine.

Zer0xFF:
does this revision commit seem correct?

Source/iop/IopBios.cpp
```diff
@@ -50,8 +50,11 @@
 #define BIOS_MODULESTARTREQUEST_HEAD_BASE (CIopBios::CONTROL_BLOCK_START + 0x0018)
 #define BIOS_MODULESTARTREQUEST_FREE_BASE (CIopBios::CONTROL_BLOCK_START + 0x0020)
 #define BIOS_HANDLERS_BASE (CIopBios::CONTROL_BLOCK_START + 0x0100)
-#define BIOS_HANDLERS_END (BIOS_THREADS_BASE - 1)
-#define BIOS_THREADS_BASE (CIopBios::CONTROL_BLOCK_START + 0x0200)
+#define BIOS_HANDLERS_END (BIOS_PSEUDO_INTRHANDLER_TABLE_BASE - 1)
+#define BIOS_PSEUDO_INTRHANDLER_TABLE_BASE (0x480)
```
jpd002: Please call these `BIOS_SYSTEM_INTRHANDLER_TABLE_*`.
(ie.: `BIOS_SYSTEM_INTRHANDLER_TABLE_BASE` and `BIOS_SYSTEM_INTRHANDLER_TABLE_SIZE`)

Also, can you add a `static_assert` that makes sure that `BIOS_SYSTEM_INTRHANDLER_TABLE_BASE` is greater than `CIopBios::CONTROL_BLOCK_START`. This is just to make sure we don't forget to take this area in consideration if we ever change the start address of the control block.

Source/iop/IopBios.cpp
```diff
@@ -120,6 +123,8 @@ CIopBios::~CIopBios()
 
 void CIopBios::Reset(const Iop::SifManPtr& sifMan)
 {
+	auto cdvdHandler = reinterpret_cast<uint32*>(&m_ram[BIOS_PSEUDO_INTRHANDLER_TABLE_BASE + (BIOS_PSEUDO_INTRHANDLER_ENTRY_SIZE * Iop::CIntc::LINES::LINE_CDROM)]);
```
jpd002: Please move this to a function. Something like `PopulateSystemIntcHandlers`.

Also, please use the `SYSTEM_INTRHANDLER` structure to set the `handler` member to 1. 

Would also be good to add the link to the poweroff.c line on PS2SDK's GitHub repo that shows why we need that.

Source/iop/IopBios.cpp
```diff
@@ -50,8 +50,11 @@
 #define BIOS_MODULESTARTREQUEST_HEAD_BASE (CIopBios::CONTROL_BLOCK_START + 0x0018)
 #define BIOS_MODULESTARTREQUEST_FREE_BASE (CIopBios::CONTROL_BLOCK_START + 0x0020)
 #define BIOS_HANDLERS_BASE (CIopBios::CONTROL_BLOCK_START + 0x0100)
-#define BIOS_HANDLERS_END (BIOS_THREADS_BASE - 1)
-#define BIOS_THREADS_BASE (CIopBios::CONTROL_BLOCK_START + 0x0200)
+#define BIOS_HANDLERS_END (BIOS_PSEUDO_INTRHANDLER_TABLE_BASE - 1)
+#define BIOS_PSEUDO_INTRHANDLER_TABLE_BASE (0x480)
+#define BIOS_PSEUDO_INTRHANDLER_ENTRY_SIZE (4 * 2)
```
jpd002: Please add a structure named `SYSTEM_INTRHANDLER` with the two members instead of defining the size like that.

jpd002:
I've submitted a few change requests, but this seems to do the job. 

Let's hope there's not too many of these out there.

jpd002:
Added a few other review comments. Should be good to merge after that!

Thanks!

Zer0xFF:
ill get on it 👍 

should i also add a static_assert for `sizeof(SYSTEM_INTRHANDLER)` since its a fix PS2 structure.

jpd002:
Yep, the static_assert on the size of the struct is a good idea!

jpd002:
Tested:
Castlevania: Curse of Darkness - OK
FF12 - OK
Monster World Collection - OK
Odin Sphere - OK
Marvel vs. Capcom 2 - OK
Capcom vs. SNK 2 - Doesn't boot anymore

I'll dig to see what's up with Capcom vs. SNK 2.

jpd002:
Commenting out `PopulateSystemIntcHandlers` doesn't fix it. Suspecting the game may not like the additional memory we took to place the thing (control block end is larger).

Zer0xFF:
that sucks, i avoid making the list fuller to avoid something like that :/

the issue is just the extra space we're taking right.
I imagine things getting moved around should be ok, so might look into that and see

Zer0xFF:
we can get back 0x200 from "BIOS_THREADS_BASE ", since this shouldn't be here anyway.

BIOS_VPL_SIZE 0x180
BIOS_FPL_SIZE 0x180
BIOS_MESSAGEBOX_SIZE 0x180
any of these 3 can perfectly slot between BIOS_HANDLERS_END and BIOS_SYSTEM_INTRHANDLER_TABLE_BASE
with that we gain back 0x180 +0x200 which brings us back to below 0x10000

Zer0xFF:
On my side tested:
Future Tactics
MegaMan X8
FFXII
Champions of Norrah Realms Of EverQuests
Rogue Galaxy
all boot to in game

jpd002:
Castlevania: Curse of Darkness - OK
Monster World Collection - OK
Odin Sphere - OK
Marvel vs. Capcom 2 - OK
Capcom vs. SNK 2 - OK
FFX - OK
Shadow Hearts 2 - OK
Kingdom Hearts - OK
PGEN - OK
InfoNES - OK

Everything looks good. Merging!

PR #831, opened 2019-11-25
remove libc from AppImage

### LINKED ISSUE #830: The newest version doesnt start up anymore - with the new merge
Running Play! in the terminal produces this error

Play: /tmp/.mount_Play!-qY7Wfc/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.26' not found (required by /usr/lib/libX11.so.6)
Play: /tmp/.mount_Play!-qY7Wfc/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /usr/lib/libglib-2.0.so.0)
Play: /tmp/.mount_Play!-qY7Wfc/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /usr/lib/libp11-kit.so.0)
Play: /tmp/.mount_Play!-qY7Wfc/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.26' not found (required by /usr/lib/libp11-kit.so.0)

This to me seems that play! isn't using the latest version 2.30-2

System OS: Manjaro KDE Arch

Zer0xFF:
closes #830 

`linuxdeployqt` doesnt like xenial glibc version, so we must use `-unsupported-allow-new-glibc`,
however, that param still doesnt allow us to package an image, as it expect us to package `glibc`,
but it only checks for `appdir/usr/share/doc/libc6/copyright`.

Note:
we could build on trusty to avoid the work around, but that would net us nothing.
as `-unsupported-allow-new-glibc` does nothing more than skip a check (nothing more or less is packaged)
https://github.com/jpd002/Play-/compare/e4be67948255e207a299a6b3881c77cd546c290c...84c6697b437fdd122d8769440b960a91d311ce51?expand=1

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -17,7 +17,6 @@ travis_before_install()
             sudo add-apt-repository --yes ppa:beineri/opt-qt-5.12.3-xenial
             sudo apt update -qq
             sudo apt install -qq qt512base gcc-9 g++-9 libgl1-mesa-dev libglu1-mesa-dev libalut-dev libevdev-dev
-            apt download libc6
         fi
     elif [ "$TARGET_OS" = "Linux_Clang_Format" ]; then
         wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


jpd002:
Sounds good! Thanks!

PR #833, opened 2019-11-28
Fix path builder

Zer0xFF:
as far as i can tell, these seem to be the only problematic lines, ill do another sweep later.

if device path starts with "/" the combined path truncates all of basePath except for partition part...

on windows
```cpp
std::filesystem::path basePath = "C:\blah\blash\blah";
char* devicePath = "/evil_access";
basePath / devicePath == "C:\evil_access";
```

Linux:
https://coliru.stacked-crooked.com/a/76ffae7f68e39128

~~Note: need to test this fix on Linux/OSX~~

Note to self:
fixes in `IOP_McServ` can probably just use `MakeHostPath()` instead

### DIFF
--- a/Source/iop/Iop_PathUtils.cpp
+++ b/Source/iop/Iop_PathUtils.cpp
@@ -1,0 +1,17 @@
+#include <cstring>
+#include "Iop_PathUtils.h"
+
+fs::path Iop::PathUtils::MakeHostPath(const fs::path& baseHostPath, const char* guestPath)
+{
+	if(strlen(guestPath) == 0)
+	{
+		//If we're not adding anything, just return whatever we had
+		//We don't want to introduce a trailing slash since it will
+		//break other stuff
+		return baseHostPath;
+	}
+	auto result = baseHostPath;
+	result.concat("/");
+	result.concat(guestPath);
+	return result;
+}
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -260,6 +260,8 @@ set(COMMON_SRC_FILES
 	iop/Iop_MtapMan.h
 	iop/Iop_PadMan.cpp
 	iop/Iop_PadMan.h
+	iop/Iop_PathUtils.cpp
+	iop/Iop_PathUtils.h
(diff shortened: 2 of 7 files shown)


Zer0xFF:
small update regarding this, this seems unfortunate but this is actually the intended behaviour as pre cppreference https://en.cppreference.com/w/cpp/filesystem/path/append

`GetEntSpace()` is used by Drakengard
`GetFile()` is used by PSMS Reloaded
can't find a game that uses `GetDirectory()`

jpd002:
Yeah, that fs::path behavior really sucks for us.

Would it work to move `MakeHostPath` to another file (ex.: `Iop_PathUtils`) and have `CDirectoryDevice` use that?

I think InfoNES uses `GetDirectory`.

Zer0xFF:
Ya, I can do this.

I'll check InfoNES to be sure this is working as intended.

Zer0xFF:
I cant seem to find a link for InfoNES to test.

also, it seems `MakeHostPath()` was a static function so it could have been used without moving it to its own function, so do you have a preference in that case?

jpd002:
I don't want to add dependencies between `McServ` and `Ioman`, that's why I want a separate place for that function to live.

And yeah, the `C` should be dropped in the namespace name.

Zer0xFF:
ah makes sense.

I've tested `ps2psxe` and `GetDirectory()` seems to work fine

Source/iop/Iop_PathUtils.cpp
```diff
@@ -0,0 +1,17 @@
+#include <cstring>
+#include "Iop_PathUtils.h"
+
+fs::path Iop::PathUtils::MakeHostPath(const fs::path& baseHostPath, const char* mcPath)
```
jpd002: Could you rename `mcPath` to `guestPath`?

jpd002:
Looking good, will merge after the cosmetic change is done.

jpd002:
Awesome! Thanks for this!

seisdr:
Could you specific directory to search for games on for android version

PR #835, opened 2019-12-10
Remove boost dependency

xsacha:
Boost has already been removed from the project, but this include somehow lasted and was causing an error if boost was not available.
This file could probably be removed anyway since it hasn't been updated.


jpd002:
Thanks!

PR #836, opened 2019-12-19
Qt debugger

Zer0xFF:
TODO: (for the scope of this PR)

- [x] `CFunctionsView` : groups
- [x] set proper disasm size
- [x] segfault on close
- [x] test effect of DPI (especially with breakpoint/arrow icons)
- [x] double check RegView, from what i can tell, its all complete
- [x] layout saving, (issue with Qt and getting visibility value)
- [x] Elf Info, need to test `SHT_DYNAMIC`

- [ ] move Qt UI creation from c++ to Qt Creator (had a look, i wouldnt recommend it, `QMdiSubWindow` isn't supported for preview in Qt Creator, and most windows contain 1 widget anyway)
- [x] use slot/signals to ensure UI access through UI thread

- [x] move framedebugger from DirectX to OpenGL
- [x] move framedebugger from Win32 UI to Qt

- [x] check previous kick qt logic
- [x] test on OSX
- [x]  `isExposed` surface doesnt contain this method... ~~template GS?~~

- Disassembly
  - [x] remove column headers
  - [x] make font smaller (Maybe use monospace font like Courier).
  - [x] Columns are initially very tiny. Would be nice to have sensible default widths for I-Mn and I-Op columns
  - [x] When stepping, the view range is updated to keep the PC arrow at the center of the view. This makes it very hard to follow what's going on. The view range should be updated only if the PC goes out of range of the view.

- Registers
  - [x] Remove the row and column headers
- Memory View
  - [x] Fix: memory view display (needs a 2nd look)
- Functions View
  - [x] Fix: Exception when double-clicking "Global"

Notes:
- as of yet, this has been tested on Win32/linux
~~- The Frame debugger is currently disabled even on windows (should you want to merge this, I'll push a commit to enable on Win32, until i get to porting it over as well)~~
~~- Elf Info hasn't been ported, and I think it'll be the last thing for the scope of this PR (the frame debugger would likely come in a separate PR)~~

~~Please take a look at this question: https://github.com/jpd002/Play-/pull/836#discussion_r360131758~~


Source/ui_qt/DebugSupport/QtDebugger.cpp
```diff
@@ -0,0 +1,939 @@
...
+}
+
+void QtDebugger::OnExecutableChange()
+{
+	//SendMessage(m_hWnd, WM_EXECCHANGE, 0, 0);
+}
+
+void QtDebugger::OnExecutableUnloading()
+{
+	//SendMessage(m_hWnd, WM_EXECUNLOAD, 0, 0);
```
Zer0xFF: do `SendMessage()` sole function is to trigger `OnWndProc()`? so we can just map these calls directly to their equivalent call?

jpd002:
The purpose of using SendMessage is to make sure the UI update code is called in the UI thread. There was some instances where we could get caught in a deadlock if we tried to update the UI stuff on the VM thread. This is needed because of the way Win32 handles things, maybe it’s not needed for Qt.

Source/ui_qt/DebugSupport/QtDebugger.cpp
```diff
@@ -0,0 +1,854 @@
+#include "QtDebugger.h"
+#include "ui_QtDebugger.h"
+
+#include <QInputDialog>
+#include <QMessageBox>
+
+#include "iop/IopBios.h"
+#include <stdio.h>
```
jpd002: I think these 3 includes can be removed.

Source/ui_qt/DebugSupport/RegViewFPU.cpp
```diff
@@ -0,0 +1,106 @@
+#include <stdio.h>
+#include <string.h>
+#include <QHeaderView>
+#include <QMenu>
+#include "RegViewFPU.h"
+
+#define MENUCMD_BASE 40000
```
jpd002: This is probably not needed.

Source/ui_qt/DebugSupport/FunctionsView.cpp
```diff
@@ -0,0 +1,274 @@
...
+CFunctionsView::CFunctionsView(QMdiArea* parent)
+    : QMdiSubWindow(parent)
+{
+
+	resize(300, 700);
+
+	parent->addSubWindow(this);
+	setWindowTitle("Functions");
+
+	m_treeWidget = new QTreeWidget(this);
```
jpd002: Could we convert the UI creation code to a Qt UI file?

jpd002:
Did a first pass on the code, looking good overall.

Couldn’t flag all instances because I’m on mobile, but please remove the code you’ve commented.

Also, one recurring thing I saw was empty destructors in cpp files. They can be replaced with = default; in the class definition.

Gonna give the debugger a spin once I get the chance. I’ll probably have more comments to give then :)

Thanks!

Source/ui_qt/DebugSupport/CallStackWnd.cpp
```diff
@@ -0,0 +1,73 @@
+#include "CallStackWnd.h"
+#include "string_cast.h"
+#include "lexical_cast_ex.h"
+#include "DebugUtils.h"
+#include "MIPS.h"
+#include "BiosDebugInfoProvider.h"
+
+#define CLSNAME _T("CallStackWnd")
```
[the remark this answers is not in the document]
Zer0xFF: 👍  there are few preprocessors that I should remove.

Source/ui_qt/DebugSupport/ELFSectionView.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+void CELFSectionView::ResizeEvent()
+{
+	if(!m_model || m_bytesPerLine)
+		return;
+
+	AutoColumn();
+}
+
+void CELFSectionView::ShowContextMenu(const QPoint& pos)
```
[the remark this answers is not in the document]
Zer0xFF: I think this can be made more generic.

Source/ui_qt/DebugSupport/FunctionsView.cpp
Zer0xFF: for all of them or just this view?

Zer0xFF:
> The purpose of using SendMessage is to make sure the UI update code is called in the UI thread. There was some instances where we could get caught in a deadlock if we tried to update the UI stuff on the VM thread. This is needed because of the way Win32 handles things, maybe it’s not needed for Qt.

ah, well that should also be required in Qt, though it didn't complain about it yet, so ill make few changes. (Qt signal/slot should handle that)

> but please remove the code you’ve commented.

there are 2 particular lines that are currently commented out in the master branch, should those go as well?

https://github.com/jpd002/Play-/pull/836/files/0bb655de9a1ee7e4a8e737b4afb804c189cf5ff1#diff-88f86fbf6e45bb4cd1e9d8784e419d44R675
https://github.com/jpd002/Play-/pull/836/files/0bb655de9a1ee7e4a8e737b4afb804c189cf5ff1#diff-88f86fbf6e45bb4cd1e9d8784e419d44R688
and note, I will re-enable `MessageBeep()` in windows with preprocessors (unless you want to just remove it)

> Also, one recurring thing I saw was empty destructors in cpp files. They can be replaced with = default; in the class definition.

 👍

jpd002:
Yeah, these two lines can be removed too.

Can we use this to replace MessageBeep? https://doc.qt.io/qt-5/qapplication.html#beep

Source/ui_qt/DebugSupport/FunctionsView.cpp
jpd002: All of them if possible.

Zer0xFF:
> Yeah, these two lines can be removed too.
> 
> Can we use this to replace MessageBeep? https://doc.qt.io/qt-5/qapplication.html#beep

interesting, that might just do it.

Zer0xFF:
I just pushed 87ed255 to get your opinion, from what I can tell, `QMdiSubWindow` wasn't made to be used directly in Qt UI file, in that it's children widget aren't displayed correctly, to fix that we have to call `setWidget(...)` after setting up the UI https://github.com/jpd002/Play-/commit/87ed2551d7c315a38f216fb1f711a989830ba5fd#diff-dc18fa66872c223800e76ef0fb99bd5dR14
because of that, the widget is set from that call, so there is a small disconnect between UI file and what's on display, https://github.com/jpd002/Play-/commit/87ed2551d7c315a38f216fb1f711a989830ba5fd#diff-fc7ffbf0cae1d7fe3826b2e489a2a047R16 <-- this grid layout is purely cosmetic for Qt Creator(without the grid, the widget has no anchor point, but when the actual view, its anchored to its parent), but has no part in the actual view, since the view starts from `setWidget(ui->addressListView)` aka 2 lines below
 it's not that big of a deal, but I want to get your option before continuing ahead.

Zer0xFF:
Alright, as of 600daed frame debugger runs OpenGL in Qt. (I've yet to test on any platform beside Win, yet)

however, I've made 2 observation

1) simplifying widget inheritance from `QMdiSubWindow` -> whatever the base widget is, is probably the "correct" thing to do, to make them more usable in other part of the UI, however, as of now, this has been useful in for only 3 views, `CMemoryViewMIPSWnd` since thats used in the debugger and frame debugger, and `CELFView` IF we're to make it work as a standalone, frame debugger since its a complex layout (and i do think these 3 should be converted to qt.ui)

2) converting this to Qt UI files is a pointless task, as most of those widgets are single widget in a window, so having ui file does nothing to give us more control, but instead its just an additional file in the build system.

3) by moving from "`QMdiSubWindow` -> whatever the base widget is"
we now have to keep reference to widget class and window class, or keep reference to parent and call for the widget class every time we need it, resulting in needless casting everywhere
as per this commit
https://github.com/jpd002/Play-/pull/836/commits/86c67b5396565f1a31bf6685435894a9f57703ec#diff-88f86fbf6e45bb4cd1e9d8784e419d44R405-R406
a pattern repeated for every widget

this is what all widget in qt ui file would look like https://gist.github.com/Zer0xFF/63842060c178b17cf263718876799307

thoughts?

jpd002:
So, I've tried this a while ago and had written a first review:

UI
- Disassembly
  - I think we should let go of the column headers
  - Can we make the font smaller to see more instructions? Maybe use monospace font like Courier.
  - Columns are initially very tiny. Would be nice to have sensible default widths for I-Mn and I-Op columns
  - When stepping, the view range is updated to keep the PC arrow at the center of the view. This makes it very hard to follow what's going on. The view range should be updated only if the PC goes out of range of the view.
- Registers
  - Can we remove the row and column headers?
- Memory View
  - Bytes per line isn't respected for 32-bit values (16-bytes per line with 32 bit integers should show 4 values per row)
- Functions View
  - Exception when double-clicking "Global"

I'll check it out again, maybe these don't really apply anymore. Just to let you know what I've been using my debugger for years and I'll have a lot of things to bring up due to the way I use it that might not be really obvious, so, sorry in advance for that.

I think this is the future and I also want the GS debugger (hoping that we can make it work with Vulkan at some point), so, let's make this work :)

Thanks!

Zer0xFF:
> sorry in advance for that.

haha, don't worry about it.

my exams are coming up soon, so ill have a look into these once I'm done with them 🤞

krieghof:
Hello, I think it is very important to get the debugger running on Linux and Mac too. So my question is, are the UI files the only blocker for this MR? Because I think a working debugger on the other platforms is way more important than those files translated.

I'm also trying to make the framedebugger work with vulkan, and I got a friend on board who is helping me with the UI.
But we don't really want to create the .ui files right now, because we are thinking about an UI overhaul and fixing the problems @jpd002 mentioned before. It would also help me and my friend if we could work on the master branch, instead of an already existing MR.


@jpd002 I would also add to your list:
- Disassembly
  - Get rid of the PC arrow, highlight the current line in blue.
- Main Window
  - Hide "Dump Next Frame" when Vulkan is used. (At least for now)
- Setting Dialog
  - Add option to change default directories (framedumps, ...)

Zer0xFF:
I do think creating UI files for these widgets isn't advised, as i noted before, each UI is a single widget, custom widget and are wrapped in QMdiSubWindow, QMdiSubWindow dont have a valid represenation for them in Qt creator, so you just see blank view.

I do also need to double check 5b8eff4, but other than that, i do believe i addressed everything(/most) of what has been mentioned here https://github.com/jpd002/Play-/pull/836#issuecomment-620846660, so ill try to make time over the next few days to double check this.

krieghof:
@Zer0xFF 

> I do think creating UI files for these widgets isn't advised, as i noted before, each UI is a single widget, custom widget and are wrapped in QMdiSubWindow, QMdiSubWindow dont have a valid represenation for them in Qt creator, so you just see blank view.

Ah okay, now I got it, sounds like the path is clear now.
Those are great news, thanks for your contribution.

Zer0xFF:
I do believe as of that last commit, everything in the TODO list should have been addressed

jpd002:
I've been using the new debugger since I can't boot on Windows anymore with my M1 MacBook and I want to be able to debug games on the go whenever we'll be allowed to go out.

I think we're very close to the point where we can merge this in the main line. I still have a few deal breakers though:

- Maybe this was mentioned earlier, but it doesn't seem to be possible to see comments in the disassembly view. Like, if I add a comment on an instruction using the contextual menu, nothing appears.
- There's still something really wonky about the memory view when using 32-bit units (words or fp32). If I switch to that mode and try to go to a specific address, it doesn't seem to work consistently.
- Seems like there's a small bug in the effective address computation in the disassembly view:
image

That's it for now, I'll keep using it and see if anything else stands out.

jpd002 (dismissed):

Zer0xFF:
those issues should now be fixed.

do let me know if you notice anything else, as it can be tough spotting those issues, without being a user of the debugger myself.

jpd002:
Thanks for the update.

Tested a bit more today, found some additional problems:

- Seems it's not possible to see kernel areas in the disassembly view:
スクリーンショット 2021-01-02 午前10 00 05

Double clicking there, entering the address manually or doing "Goto PC" doesn't do anything.

- There's still a bug in the memory view:
image
The memory view shows all zeroes for the same area that we see in the disassembly view.

- Could we have the comments and target in the same column in the disassembly view? It was working that way on the Win32 debugger, with the target function name having priority over a comment. If it's possible to have different colours for them, it would be nice (blue for target functions and green for comments).

- It doesn't seem possible to remove a comment by going in "Edit Comment" and clearing out whatever is there and clicking ok.

Thanks!

Zer0xFF:
> Double clicking in the call stack, entering the address manually or doing "Goto PC" doesn't do anything.

ummm... I cant replicate any of these? do they happen all the time, or "randomly"?
also, is the emu paused when you try this? as per the current implementation, which the [Qt implementation](https://github.com/jpd002/Play-/pull/836/files#diff-4418cf02f8b0b388e0f96664e9fa8123038b7dff2f2b107efa255df571ce8237R304) follows, its set to not trigger if the emu is running, though do note that double clicking the call stack doesnt follow the same restriction

> It doesn't seem possible to remove a comment by going in "Edit Comment" and clearing out whatever is there and clicking ok.

fixed, was caused by over restrictive guarding

> The memory view shows all zeroes for the same area that we see in the disassembly view.

issue with address calculation, hopefully the final fix for this :/

> Seems it's not possible to see kernel areas in the disassembly view:

I was using [PS2::EE_RAM_SIZE](https://github.com/jpd002/Play-/pull/836/files#diff-d72f94bacb8ffbdd4392078ab6df54db466f311417cb6f3e02c1b6e08699b963R61) to set the views limit, which seems to be incorrect for this, I've changed it to set the limit at the end of the bios address, but which got me questioning what size should we be using for VU0/1, just their sizes? and we're ignoring IO avoid accessing IO regs?and for IOP its PS2::IOP_RAM_SIZE (as the rest are just remaps right)

this also highlighted how badly handles list/tables... Qt allocate all the memory it needs for the view, and then just dies when I set the view size to the end of the bios address.

> Could we have the comments and target in the same column in the disassembly view? It was working that way on the Win32 debugger, with the target function name having priority over a comment.

sure

> If it's possible to have different colours for them, it would be nice (blue for target functions and green for comments).

should be possible with custom paint method

jpd002:
> Double clicking in the call stack, entering the address manually or doing "Goto PC" doesn't do anything.

It's the "cannot see kernel areas in disassembly view" problem. Other addresses work fine.

> I was using PS2::EE_RAM_SIZE to set the views limit, which seems to be incorrect for this, I've changed it to set the limit at the end of the bios address, but which got me questioning what size should we be using for VU0/1, just their sizes? and we're ignoring IO avoid accessing IO regs?and for IOP its PS2::IOP_RAM_SIZE (as the rest are just remaps right)

For now, we can only make this work on RAM/ROM areas. For VUs, just their micro memories. 

There's a larger problem with virtual memory, but that's not handled correctly by the emulator anyways at the moment, but in the future, we would probably want to be able to see any address in a 32-bits range for the EE and the IOP. So, I dunno if there would be a possibility of having some kind of virtual list view in Qt to make this possible (maybe for a future PR).

Also, like you mentioned the disassembly view is currently broken for the EE, looking forward to your solution 😃 

The rest of the issues seem covered. Thanks!

Zer0xFF:
> It's the "cannot see kernel areas in disassembly view" problem. Other addresses work fine.

ah, good, so this should be fixed with the current fix i have.

> So, I dunno if there would be a possibility of having some kind of virtual list view in Qt to make this possible

not sure if i fully understand, but my current fix (PoC) only handles EE case, however, I figured a better fix would be to retrieve the instruction mapping list, and display those areas, though i do make the assumption, you only want to see instruction memory space, as anything else will be garabage and meaningless to the disasm. (nearly done with that actually, so ill try cleaning it up, testing and psuh it out asap)
aka, as you can see the transition between EE ram and bios space
image

Source/ui_qt/DebugSupport/RegViewVU.cpp
```diff
@@ -0,0 +1,189 @@
...
+	WriteTableLabel(38, "STKF");
+	WriteTableLabel(39, "CLIP");
+	WriteTableLabel(40, "PIPE");
+	WriteTableLabel(41, "PIPEQ");
+	WriteTableLabel(42, "PIPEP");
+	WriteTableLabel(43, "PIPEM");
+	WriteTableLabel(44, "PIPEC");
+	for(unsigned int x = 0; x < 16; x++)
+	{
+		if(x < 10)
```
jpd002, line 38: Wouldn't it be possible to use a better formatting string to avoid a `if` here? I think `%02d` would do the trick.

Source/ui_qt/DebugSupport/RegViewVU.cpp
```diff
@@ -0,0 +1,189 @@
...
+CRegViewVU::CRegViewVU(QWidget* parent, CMIPS* ctx)
+    : CRegViewPage(parent)
+    , m_ctx(ctx)
+{
+	AllocateTableEntries(2, 45 + 16);
+	setColumnWidth(0, 46);
+	horizontalHeader()->setStretchLastSection(true);
+	for(unsigned int x = 0; x < 32; x++)
+	{
+		if(x < 10)
```
jpd002, line 18: Same comment as below, there's probably a better formatting string that can be used here.

Source/ui_qt/DebugSupport/RegViewFPU.cpp
```diff
@@ -0,0 +1,104 @@
...
+    , m_nViewMode(VIEWMODE_SINGLE)
+{
+	// Setup the register names
+	AllocateTableEntries(2, 35);
+	setColumnWidth(0, 40);
+	horizontalHeader()->setStretchLastSection(true);
+	for(unsigned int x = 0; x < 32; x++)
+	{
+		setRowHeight(x, 16);
+		if(x < 10)
```
jpd002, line 19: Can we use a better format specifier here? Like `%02d`?

Source/ui_qt/main.cpp
```diff
@@ -38,9 +40,9 @@ int main(int argc, char* argv[])
 	parser.process(a);
 
 	MainWindow w;
-#ifdef DEBUGGER_INCLUDED
-	a.installNativeEventFilter(&w);
-#endif
+	//#ifdef DEBUGGER_INCLUDED
```
jpd002, line 43: Can we remove this dead code?

Source/ui_qt/mainwindow.cpp
```diff
@@ -780,33 +786,33 @@ void MainWindow::resizeWindow(unsigned int width, unsigned int height)
...
-		return true;
-	}
-	if(m_frameDebugger && (activeWnd == m_frameDebugger->m_hWnd) &&
-	   TranslateAccelerator(m_frameDebugger->m_hWnd, m_frameDebugger->GetAccelerators(), msg))
-	{
-		return true;
-	}
-	return false;
-}
+//bool MainWindow::nativeEventFilter(const QByteArray&, void* message, long* result)
```
jpd002, line 789: Can we remove this dead code?

Source/ui_qt/mainwindow.cpp
```diff
@@ -780,33 +786,33 @@ void MainWindow::resizeWindow(unsigned int width, unsigned int height)
...
+//	}
+//	return false;
+//}
 
 void MainWindow::ShowDebugger()
 {
-	m_debugger->Show(SW_SHOWMAXIMIZED);
-	SetForegroundWindow(*m_debugger);
+	m_debugger->showMaximized();
+	//SetForegroundWindow(*m_debugger);
```
jpd002, line 809: Remove dead code.

Source/ui_qt/mainwindow.cpp
```diff
@@ -780,33 +786,33 @@ void MainWindow::resizeWindow(unsigned int width, unsigned int height)
...
 }
 
 void MainWindow::ShowFrameDebugger()
 {
-	m_frameDebugger->Show(SW_SHOWMAXIMIZED);
-	SetForegroundWindow(*m_frameDebugger);
+	m_frameDebugger->show();
+	//SetForegroundWindow(*m_frameDebugger);
```
jpd002, line 815: Remove dead code.

Source/ui_qt/mainwindow.h
```diff
@@ -33,10 +33,10 @@ namespace Ui
 #endif
 
 class MainWindow : public QMainWindow
-#ifdef DEBUGGER_INCLUDED
-    ,
-                   public QAbstractNativeEventFilter
-#endif
+//#ifdef DEBUGGER_INCLUDED
```
jpd002, line 36: Remove dead code.

Source/ui_qt/mainwindow.h
```diff
@@ -93,7 +93,7 @@ class MainWindow : public QMainWindow
 	void resizeWindow(unsigned int, unsigned int);
 	void UpdateGSHandlerLabel(int);
 #ifdef DEBUGGER_INCLUDED
-	bool nativeEventFilter(const QByteArray&, void*, long*) Q_DECL_OVERRIDE;
+	//bool nativeEventFilter(const QByteArray&, void*, long*) Q_DECL_OVERRIDE;
```
jpd002, line 96: Remove dead code.

jpd002 (changes requested):

jpd002:
I've tested your solution and it works for me. We can see how we can have a view of the whole virtual memory space in there later.

I've reviewed the code and added a few comments. Also, I guess that the Win32 debugger is going to be dead once this is merged, right? I see that a part of the Win32 code is still there, would it be safe to remove it completely?

Source/MemoryMap.h
```diff
@@ -22,6 +22,12 @@ class CMemoryMap
 		MEMORYMAP_TYPE_FUNCTION
 	};
 
+	struct RANGE
```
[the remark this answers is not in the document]
Zer0xFF: I wanted to avoid returning pointer and handler, other than that, no, we can return `m_instructionMap` directly

Zer0xFF:
> I've tested your solution and it works for me. We can see how we can have a view of the whole virtual memory space in there later.

we'd likely have to create our own widget to make it work, why is there a need to show the entire virtual memory space here, do you expect to find executable code in the non mapped region?

> I've reviewed the code and added a few comments.

I'll try to address them later tonight if i get a chance.


> Also, I guess that the Win32 debugger is going to be dead once this is merged, right? I see that a part of the Win32 code is still there, would it be safe to remove it completely?

yes, it's safe to remove all of it, i mainly kept it until now, since I was refering back to it as i worked on the Qt stuff, so it was easier to keep it in place, since everything has been moved over, it can now be removed.

Source/MemoryMap.h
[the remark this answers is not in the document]
Zer0xFF: `GetInstructionMaps` should return a `const ref`, right, but we dont want to keep a ref in `CQtDisAsmTableModel`, to avoid life cycle dependencies?

jpd002:
> we'd likely have to create our own widget to make it work, why is there a need to show the entire virtual memory space here, do you expect to find executable code in the non mapped region?

There can be executable pages mapped anywhere using the TLB. The emulator doesn't support this ATM and only one game uses this from what I've seen (Goemon).

Alright, this is all good for me! Merging this 😄 Thanks a bunch for your hard work!

jpd002 (approved):

Zer0xFF:
> Alright, this is all good for me! Merging this 😄 Thanks a bunch for your hard work!

Glad to be of service :)

giphy


> There can be executable pages mapped anywhere using the TLB. The emulator doesn't support this ATM and only one game uses this from what I've seen (Goemon).

ah, i see, I'll try to have a look at that soon.
though does the memory have to be marked by emu to be used? aka is there a syscall to set this up?
if so, it can just be added to the map then (on debugger build) , and it will appear in the dism, else a custom widget is the only real way, as I've already checked the Qt bug tracker, they're refusing to fix this, as "the user will never be able to see this amount of rows anyway".

jpd002:
> though does the memory have to be marked by emu to be used? aka is there a syscall to set this up?

It's done through writes to the TLB (`TLBWI`/`HandleTLBWrite`). But that doesn't change the memory map, it only changes the operation of `TranslateAddressTLB` which handles the translation of virtual addresses to physical addresses. 

In any case, it's not crucial that we look at this right now since it's used by only 1 game.

PR #840, opened 2020-01-03
[WIP] Move the Framedebugger to OpenGL

Zer0xFF:
this PR is phase 1 of porting the frame debugger, aka moving frame debugger to OpenGL (this PR)
phase 2, moving it to Qt to make it multi-platform (will either be it's own PR or part of #836)

TODO:

- [x] Texture coloring, currently displaying as shades of red
- [ ] Switching context renders a blank texture/framebuffer, but changing packet fixes the render (will ignore this for now, since this is likely a ui issue, which will be removed anyway )

~~I'd need to focus on my exam that start next week, so i though I'd post my progress so far~~


Zer0xFF:
now part of #836

PR #845, opened 2020-01-24
capture screenshot directly from display framebuffer, at rendering resolution

Zer0xFF:
I've noticed that this PR, this code block below is now duplicated between `GetScreenshot()`, and both `FlipImpl()` in OpenGL and Vulkan, perhaps its best to abstract it into the `CGSHandler`, as these are they're graphics api independent (function name suggestions? `PopulatePS2DisplayInfo(DISPFB& fb, DISPLAY& d)`)

```cpp
	DISPFB fb;
	DISPLAY d;
	std::lock_guard<std::recursive_mutex> registerMutexLock(m_registerMutex);
	unsigned int readCircuit = GetCurrentReadCircuit();
	switch(readCircuit)
	{
	case 0:
		fb <<= m_nDISPFB1.value.q;
		d <<= m_nDISPLAY1.value.q;
		break;
	case 1:
		fb <<= m_nDISPFB2.value.q;
		d <<= m_nDISPLAY2.value.q;
		break;
	}
```

### DIFF
--- a/Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
+++ b/Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
@@ -2185,11 +2185,50 @@ void CGSH_OpenGL::ReadFramebuffer(uint32 width, uint32 height, void* buffer)
 
 Framework::CBitmap CGSH_OpenGL::GetScreenshot()
 {
-	GLint viewport[4];
-	glGetIntegerv(GL_VIEWPORT, viewport);
-	auto imgbuffer = Framework::CBitmap(viewport[2], viewport[3], 32);
-	glReadPixels(viewport[0], viewport[1], viewport[2], viewport[3], GL_RGBA, GL_UNSIGNED_BYTE, imgbuffer.GetPixels());
-	return imgbuffer.FlipVertical();
+	DISPFB fb;
+	DISPLAY d;
+	std::lock_guard<std::recursive_mutex> registerMutexLock(m_registerMutex);
+	unsigned int readCircuit = GetCurrentReadCircuit();
+	switch(readCircuit)
+	{
+	case 0:
+		fb <<= m_nDISPFB1.value.q;
+		d <<= m_nDISPLAY1.value.q;
+		break;
+	case 1:
+		fb <<= m_nDISPFB2.value.q;
+		d <<= m_nDISPLAY2.value.q;
+		break;
+	}
+	unsigned int dispWidth = (d.nW + 1) / (d.nMagX + 1);
+	unsigned int dispHeight = (d.nH + 1);
+
+	bool halfHeight = GetCrtIsInterlaced() && GetCrtIsFrameMode();
+	if(halfHeight) dispHeight /= 2;
+
+	FramebufferPtr framebuffer;
+	for(const auto& candidateFramebuffer : m_framebuffers)
+	{
+		if(
+		    (candidateFramebuffer->m_basePtr == fb.GetBufPtr()) &&
+		    (GetFramebufferBitDepth(candidateFramebuffer->m_psm) == GetFramebufferBitDepth(fb.nPSM)) &&
+		    (candidateFramebuffer->m_width == fb.GetBufWidth()))
+		{
+			//We have a winner
+			framebuffer = candidateFramebuffer;
+			break;
+		}
+	}
+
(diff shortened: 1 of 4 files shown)


jpd002:
Agreed, let's make that `std::pair<uint64, uint64> GetCurrentDisplayInfo()`. `GetCurrentReadCircuit` can also be moved in `CGSHandler`. I'll update the Vulkan GS handler to use these.

Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
```diff
@@ -128,23 +128,9 @@ void CGSH_OpenGL::FlipImpl()
...
-			fb <<= m_nDISPFB1.value.q;
-			break;
-		case 1:
-			d <<= m_nDISPLAY2.value.q;
-			fb <<= m_nDISPFB2.value.q;
-			break;
-		}
-	}
+	auto dispInfo = GetCurrentDisplayInfo();
+	DISPFB fb = make_convertible<DISPFB>(dispInfo.first);
```
jpd002: Please use auto.

Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
```diff
@@ -128,23 +128,9 @@ void CGSH_OpenGL::FlipImpl()
...
+	DISPLAY d = make_convertible<DISPLAY>(dispInfo.second);
```
jpd002: Please use auto.

Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
```diff
@@ -2185,11 +2137,39 @@ void CGSH_OpenGL::ReadFramebuffer(uint32 width, uint32 height, void* buffer)
 
 Framework::CBitmap CGSH_OpenGL::GetScreenshot()
 {
-	GLint viewport[4];
-	glGetIntegerv(GL_VIEWPORT, viewport);
-	auto imgbuffer = Framework::CBitmap(viewport[2], viewport[3], 32);
-	glReadPixels(viewport[0], viewport[1], viewport[2], viewport[3], GL_RGBA, GL_UNSIGNED_BYTE, imgbuffer.GetPixels());
-	return imgbuffer.FlipVertical();
+	auto dispInfo = GetCurrentDisplayInfo();
+	DISPFB fb = make_convertible<DISPFB>(dispInfo.first);
```
jpd002: Please use auto here too.

Source/gs/GSHandler.cpp
```diff
@@ -995,6 +995,60 @@ bool CGSHandler::GetCrtIsFrameMode() const
...
+std::pair<uint64, uint64> CGSHandler::GetCurrentDisplayInfo()
+{
+	std::lock_guard<std::recursive_mutex> registerMutexLock(m_registerMutex);
+	unsigned int readCircuit = GetCurrentReadCircuit();
+	DISPFB fb;
+	DISPLAY d;
+	switch(readCircuit)
+	{
+	case 0:
+		fb <<= m_nDISPFB1.value.q;
```
jpd002: No need to set the fb and d variables here, we can just return the registers values directly.
Also, we can add default above case 0 to make sure we always return something and avoid a compiler warning.

Zer0xFF: haha, i actually spent way too long trying to understand why this left shift assignment doesnt always return zero.... until i realized `convertible` is not part of the standard.

Source/gs/GSHandler.cpp
```diff
@@ -995,6 +995,57 @@ bool CGSHandler::GetCrtIsFrameMode() const
 	return smode2.ffmd;
 }
 
+std::pair<uint64, uint64> CGSHandler::GetCurrentDisplayInfo()
+{
+	std::lock_guard<std::recursive_mutex> registerMutexLock(m_registerMutex);
+	unsigned int readCircuit = GetCurrentReadCircuit();
+	DISPFB fb;
```
jpd002: These two variables are not used anymore.

Source/gs/GSHandler.cpp
jpd002: Yeah, sorry, kinda abusing C++ operators overloading here 😄

Zer0xFF: why not just overload assignment operator?

jpd002: Don’t remember why. Maybe I wanted to prevent unwanted assignments?

jpd002:
Thanks!

PR #848, opened 2020-02-09
dont fail to load game info, if it has no overview

Zer0xFF:
this is currently throwing an exception causing the entire process to fail


jpd002:
Thanks for this!

PR #849, opened 2020-02-10
Vulkan toggle

Zer0xFF:
Note:
```I think ideally we'd have a function `CGSHandler::Swap(gs)` that would copy memory/regs over, and swap the reset of the states, however, right now, all of these are protected, so this was the easiest way, without moving everything around.```
edit: it seems i misunderstood how protected members work, they can actually be accessed just fine by another instance of the same class.

Tested:Win32
Games: KH, FF12, MX8
MX8 produces weird artifacts at the bottom when swapped during dialog
videos on discord

overlay is MSI afterburn

### DIFF
--- a/Source/gs/GSHandler.h
+++ b/Source/gs/GSHandler.h
@@ -771,6 +771,7 @@ public:
 
 	virtual void SaveState(Framework::CZipArchiveWriter&);
 	virtual void LoadState(Framework::CZipArchiveReader&);
+	void Copy(CGSHandler*);
 
 	void SetFrameDump(CFrameDump*);
 
--- a/Source/ui_qt/PreferenceDefs.h
+++ b/Source/ui_qt/PreferenceDefs.h
@@ -2,4 +2,4 @@
 
 #define PREFERENCE_AUDIO_ENABLEOUTPUT "audio.enableoutput"
 #define PREF_UI_PAUSEWHENFOCUSLOST "ui.pausewhenfocuslost"
-#define PREF_VIDEO_USEVULKAN "video.usevulkan"
+#define PREF_VIDEO_GS_HANDLER "video.gshandler"
--- a/Source/ui_qt/Qt_ui/settingsdialog.ui
+++ b/Source/ui_qt/Qt_ui/settingsdialog.ui
@@ -205,42 +205,60 @@
         </widget>
        </item>
        <item>
-        <widget class="QLabel" name="label_6">
-         <property name="text">
-          <string>GS Handler:</string>
-         </property>
-        </widget>
-       </item>
-       <item>
-        <widget class="QComboBox" name="comboBox_gs_selection">
-         <item>
-          <property name="text">
-           <string>OpenGL</string>
+        <widget class="QWidget" name="gs_option_widget" native="true">
+         <layout class="QGridLayout" name="gridLayout_2">
+          <property name="leftMargin">
+           <number>0</number>
           </property>
-         </item>
-         <item>
-          <property name="text">
-           <string>Vulkan</string>
+          <property name="topMargin">
+           <number>0</number>
           </property>
-         </item>
-        </widget>
-       </item>
-       <item>
-        <widget class="QLabel" name="label_5">
-         <property name="text">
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: Source/ui_qt/PreferenceDefs.h [hunk 1]
```diff
@@ -2,4 +2,4 @@
 
 #define PREFERENCE_AUDIO_ENABLEOUTPUT "audio.enableoutput"
 #define PREF_UI_PAUSEWHENFOCUSLOST "ui.pausewhenfocuslost"
-#define PREF_VIDEO_USEVULKAN "video.usevulkan"
+#define PREF_VIDEO_GS "video.gs"
```
jpd002: Please rename this to `PREF_VIDEO_GS_HANDLER` and `video.gshandler`.

Source/ui_qt/Qt_ui/settingsdialog.ui
```diff
@@ -205,28 +205,60 @@
...
+          <property name="rightMargin">
+           <number>0</number>
+          </property>
+          <property name="bottomMargin">
+           <number>0</number>
+          </property>
+          <item row="0" column="0">
+           <widget class="QLabel" name="label_6">
+            <property name="text">
+             <string>GS Render:</string>
```
jpd002: Please rename this to "GS Handler:".

Source/ui_qt/PreferenceDefs.h
jpd002: Also, we should have an `enum` to define the possible values for that preference (currently OpenGL or Vulkan) instead of using 0 and 1 in the code.

Source/ui_qt/Qt_ui/settingsdialog.ui
```diff
@@ -205,28 +205,60 @@
...
+            </property>
+           </widget>
+          </item>
+          <item row="1" column="0">
+           <widget class="QComboBox" name="comboBox_gs_selection">
+            <item>
+             <property name="text">
```
jpd002: Might be interesting to consider to populate that from the code and reuse the same logic that is used in the status bar code. We might only have 2 GS handlers for a long time, but let's make this easy to add a 3rd one if we need.

jpd002:
Quite clever and seems to work very nicely (I need to test on macOS). Will be pretty useful for debugging and comparing between OGL and Vulkan!

Source/ui_qt/PreferenceDefs.h
Zer0xFF: should the enum go in `CGSHandler`

Zer0xFF:
I've just tested it on OSX and it works just the same... as a matter of fact, the artifcats that show up in MX8 dont show up on OSX

Source/gs/GSHandler.h
```diff
@@ -234,6 +234,15 @@ class CGSHandler
 		REGISTER_MAX = 0x80
 	};
 
+	enum GS_HANDLERS
```
jpd002: Enum is good, but it should reside in the Qt UI module. CGSHandler doesn't need to know about its implementations.

Zer0xFF: sure, but i presume you'd want android to get vulkan as well, so if it went into lets say `SettingsDialog`, you'd need another enum on android side

jpd002: True. When we get there, it would probably make sense to have it in `ui_shared`, but let's keep it in `ui_qt` for now.

jpd002:
Thanks! Gonna furiously play with the GS handler switcher now! :laughing:

Zer0xFF:
Technician disclaimer: over use of toggle can results in side effects XD

jpd002:
I think I just noticed a small bug: https://github.com/jpd002/Play-/issues/853#event-3027749938.

In the second screen shot, we see that Vulkan is used, but it still shows "OpenGL" in the status bar. I guess he went in the settings menu and changed the GS handler there but that it didn't update the status bar.

Zer0xFF:
I'll try to get to this before I head into bed

PR #855, opened 2020-02-12
Libretro fix

Zer0xFF:
to elaborate a bit on this fix,
so when running in single thread gs mode, during VM's handler destruction, the VM uses its mailbox to queue the handler destruction action, however, since the GS is not threaded, the VM mailbox can already be blocked, by a blocking call to the GS, leading to a deadlock.

so the initial solution was, to destroy the GS 1st, then start VM destruction, which worked fine for the most part, but that ran into issues, (1) between calling `gs->Release()` -> `delete gs;` a call could have already been queued up, and blocked VM mailbox (thus we needed 884ccc2)
2) we successfully delete gs, but segfault instead, since between `delete gs;` and `vm->->Pause()`, the DMA/GIF tried to send cmd to a destroyed GS...

so the easier solution was, to swap the GS on the fly, with a non-threaded null gs, that would just discard all mailbox requests,  then delete the old gs, then "regularly" destroy the VM

### DIFF
--- a/Source/PS2VM.cpp
+++ b/Source/PS2VM.cpp
@@ -207,6 +207,17 @@ void CPS2VM::Pause()
 	OnRunningStateChange();
 }
 
+void CPS2VM::PauseAsync()
+{
+	if(m_nStatus == PAUSED) return;
+	m_mailBox.SendCall([this]()
+	{
+		PauseImpl();
+		OnMachineStateChange();
+		OnRunningStateChange();
+	});
+}
+
 void CPS2VM::Reset()
 {
 	assert(m_nStatus == PAUSED);
--- a/Source/PS2VM.cpp
+++ b/Source/PS2VM.cpp
@@ -210,8 +210,7 @@ void CPS2VM::Pause()
 void CPS2VM::PauseAsync()
 {
 	if(m_nStatus == PAUSED) return;
-	m_mailBox.SendCall([this]()
-	{
+	m_mailBox.SendCall([this]() {
 		PauseImpl();
 		OnMachineStateChange();
 		OnRunningStateChange();
--- a/Source/PS2VM.h
+++ b/Source/PS2VM.h
@@ -45,6 +45,7 @@ public:
 
 	void Resume() override;
 	void Pause() override;
+	void PauseAsync();
 	void Reset();
 
 	STATUS GetStatus() const override;
--- a/Source/ui_libretro/GSH_OpenGL_Libretro.cpp
+++ b/Source/ui_libretro/GSH_OpenGL_Libretro.cpp
@@ -69,6 +69,7 @@ void CGSH_OpenGL_Libretro::FlushMailBox()
 	bool isFlushed = false;
 	SendGSCall([&]() {
 		isFlushed = true;
+		m_flipped  = true;
 	},
 	           true);
 	while(!isFlushed)
--- a/Source/ui_libretro/GSH_OpenGL_Libretro.cpp
+++ b/Source/ui_libretro/GSH_OpenGL_Libretro.cpp
@@ -69,7 +69,7 @@ void CGSH_OpenGL_Libretro::FlushMailBox()
 	bool isFlushed = false;
 	SendGSCall([&]() {
 		isFlushed = true;
-		m_flipped  = true;
+		m_flipped = true;
 	},
 	           true);
 	while(!isFlushed)
--- a/Source/ui_libretro/main_libretro.cpp
+++ b/Source/ui_libretro/main_libretro.cpp
@@ -573,11 +573,11 @@ void retro_deinit()
 	{
 		// Note: since we're forced GS into running on this thread
 		// we need to clear its queue, to prevent it freezing the reset of the system during delete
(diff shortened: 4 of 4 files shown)


i30817:
Sounds like a rust solution ( [mem_replace](https://github.com/rust-unofficial/patterns/blob/master/idioms/mem-replace.md) use in rust abstractions always looks a bit insane to me)

Zer0xFF:
I've been looking at another approach to this, the issue is `GSHandler` getting locked up due to a mailbox block, so if we make the mailbox a pointer, we can delete it when `GSHandler ::Release()` is called preventing any more msgs from being accepted.

jpd002:
I've been playing with the libretro integration and I noticed the deadlock you are mentioning here.  

It seems like the call to `Pause `is in a strange place: https://github.com/jpd002/Play-/blob/master/Source/ui_libretro/main_libretro.cpp#L580. 

I guess it's there because it's blocking. I was thinking that maybe making the `Pause `call async would help (maybe by adding a new `PauseAsync` function in PS2VM?). Then, you would probably need to pump the messages in the GS mail box (I guess through calling `ProcessSingleFrame`) till the VM is in paused state. After that, it should be safe to delete everything else.

I think it would be better to handle that this way.

Zer0xFF:
I think`PauseAsync` will avoid deadlocking, though it can't really be used in conjunction with `ProcessSingleFrame` as is, as `ProcessSingleFrame` effectively blocks util we flip the buffer, so if we pause before flipping, it'd be another deadlock and vice versa (we can maybe add a flag to `ProcessSingleFrame` to return if we pause/process all currenly pending mailbox), ill try to look into it in the next few days.

just so, if pause is async, we can't delete the VM

Zer0xFF:
Okay, `PauseAsync` did the trick, it seems my confusion in my previous comment came from incorrect assumption that `GS` has a dependency on `PS2VM`, which isn't true, so we cant deadlock `GS` by pausing the `PS2VM`, (dependency is only on `PS2VM`side, if we lock `GS`, we could end up deadlocking the `PS2VM`).

jpd002:
I have a feeling that we're missing a another part to this. 

I think we need to process messages inside the GS's mailbox until the VM is in a `PAUSED ` state. I think there could be a possibility of having a left over `Flip` call that could stay in the GS's mailbox after the `Release ` call is done. I'll try to test on my end later, but I'm not 100% sure this fixes everything.

Zer0xFF:
I had some print statements when i was testing this, and they showed that the VM is always paused before we're return from Release. (ofc, no guarantee there isn't a race condition still). 

I also had few other changes which I didn't think were needed (I've push it out now), aka I used condition variable, to wait until pause is triggered (though I don't know if pause guarantees empty GS inbox? but i guess that is the expected behaviour on the other UI anyway).

Zer0xFF:
alright on 2nd look, i think I understand this better,

so since we're calling `PauseAsync` before `Release` we can guarentee to process enough messages in VM mailbox for it to be paused, from here, everything will be the same as any other platforms implementation. not only that, but we call both `Destroy` and `DestroyGSHandler` and both of them will call `Release` again (well `Destroy` wont, since GS would have been destroyed by the time it tries to release it), granted this will now be done in the VM thread which doesnt have OpenGL context(aka if we encounter a flip), however that shouldnt matter, as we're attempting to destroy the GS, not draw more stuff to the screen.

jpd002:
> so since we're calling PauseAsync before Release we can guarentee to process enough messages in GS mailbox to allow enough VM to processes enough for it to be paused

This is a bad assumption. `PauseAsync` and `Release` can both complete while the VM thread is still waiting to be the active thread. If the VM thread takes the CPU after both `PauseAsync` and `Release` are called on the main/GS thread, it could call `SetVblank`/`Flip` (which needs acknowledgement from the main/GS thread) before processing `PauseAsync`.

That's why we need to pump messages inside the main/GS thread (to acknowledge the `Flip` call) until we know the VM thread is in the `PAUSED` state. No need for the condition variable if we're reading the VM's state in a loop. We'd probably need to add a `this_thread::yield` in there to prevent the CPU from being busy all the time.

This is very annoying thread logic, it's really sad that RetroArch doesn't allow executing OpenGL commands on another thread :sad:.

Zer0xFF:
> This is very annoying thread logic

We can both agree on that XD

are you suggesting something like this then?
```cpp
		m_virtualMachine->PauseAsync();
		auto gsHandler = m_virtualMachine->GetGSHandler();
		if(gsHandler)
		{
			// Note: since we've forced GS into running on this thread
			// we need to clear its queue, to prevent it freezing the reset of the system during delete
			while(m_virtualMachine->GetStatus() != CVirtualMachine::PAUSED)
			{
				std::this_thread::yield();
				gsHandler->Release();
			}
		}
```

jpd002:
Yep, that's what I'm suggesting. 😃

jpd002:
Nice! Thanks!

Zer0xFF:
No problem.

just to mention, in my test, atleast once, `Release` was called multiple times before the VM paused.

PR #857, opened 2020-02-18
Qt List S3 bootables

Zer0xFF:
this other PR is not required, to merge this PR anymore, but should be once done, to ensure all games can be list ~~requires https://github.com/jpd002/Play--Framework/pull/27 to correctly list/load all games~~

side note: I tried to get this to work on Android, however, that requires `Framework::HashUtils::ComputeHmacSha256` which in turns requires openssl.
alternatively, we could call android java methods from cpp to get the same results.

### DIFF
--- a/Source/s3stream/AmazonS3Client.cpp
+++ b/Source/s3stream/AmazonS3Client.cpp
@@ -172,7 +172,7 @@ GetObjectResult CAmazonS3Client::GetObject(const GetObjectRequest& request)
 	Request rq;
 	rq.method = Framework::Http::HTTP_VERB::GET;
 	rq.uri = "/" + Framework::Http::CHttpClient::UrlEncode(request.object);
-	rq.host = string_format("%s." S3_HOSTNAME, request.bucket.c_str());
+	rq.host = string_format("%s.s3-%s.amazonaws.com", request.bucket.c_str(), m_region.c_str());
 	rq.urlHost = rq.host;
 
 	if(request.range.first != request.range.second)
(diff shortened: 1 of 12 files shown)


Source/s3stream/AmazonS3Client.cpp
```diff
@@ -322,3 +322,30 @@ Framework::Http::RequestResult CAmazonS3Client::ExecuteRequest(const Request& re
 #endif
 	return response;
 }
+
+ListObjectsResult CAmazonS3Client::GetListObjects(std::string accessKeyId, std::string secretAccessKey, std::string bucketName)
```
jpd002: This should be moved in `ui_shared`. I'd like `AmazonS3Client` to stay a thin layer over S3's REST API (https://docs.aws.amazon.com/AmazonS3/latest/API/API_Operations_Amazon_Simple_Storage_Service.html). We could create a `CAmazonS3Utils` class in the ui_shared module to host that function.

jpd002:
image

Can we give these two text boxes a "read-only" style to show they are not editable.

jpd002:
Looking good, except this will need some kind of asynchronicity at some point because scanning S3 might make the UI hang for a long time if your connection is slow. Will merge after the two changes above are done.

Zer0xFF:
about read only style, if we disable the widget, then it will be greyed out, but we can't select and copy path or Serial, but we can set the background to light-gray and text color to grey seems to replicate it? any preference?

jpd002:
Humm, kinda sucks that `setReadOnly` doesn't change the color of the widget with Qt. I don't want to fiddle with custom colors because that can get messy quickly, so let's revert it to what you had before. I prefer being able to select and copy the text even if it text box looks editable. 

Sorry about that!

Zer0xFF:
no worries.
 
there is another option, we can add a keyboard shortcut to copy path/serial to clipboard, though we either set 2 different shortcuts (1 for each [path, serial], ctrl+x, ctrl+c) or 1 shortcut the copies them both, e.g "/path/to/game - SERIAL-12344". (having said that, I haven't tested shortcuts on dialogs, so small chance this doesnt work).

Zer0xFF:
>  except this will need some kind of asynchronicity at some point because scanning S3 might make the UI hang for a long time if your connection is slow

I've a commit to fix this now,
alot of the hang is due to the processing(registering bootables) we do in the callback, all of that can be moved to `std::async` above it.

however, since you're ready to merge this, I haven't pushed the changed to this PR, but will open a new one, once this is merged. (if you want to delay this PR to get those changes, I can push them here, for now, you can find the changes in this [branch](https://github.com/Zer0xFF/Play-/tree/s3_bootables_3))

jpd002:
Gonna merge this one now, please open a new PR with the async stuff. Thanks!

PR #866, opened 2020-03-09
Qt S3 bootables async processing

Zer0xFF:


Zer0xFF:
not sure why build broke there... even broke on my travis job, (recommited to force rebuild), so hopefully, it was a once off behaviour.

Source/ui_qt/bootablelistdialog.cpp
```diff
@@ -57,7 +60,16 @@ BootableListDialog::BootableListDialog(QWidget* parent)
...
 		        QModelIndex index = ui->listView->selectionModel()->selectedIndexes().at(0);
 		        bootable = model->GetBootable(index);
-		        accept();
+		        if(!m_s3_processing)
+		        {
+			        accept();
+		        }
+		        else
+		        {
+			        QMessageBox::warning(this, "Warning Message",
```
jpd002, line 69: Could we move this to a function? Like `displayWarningMessage` or something. I see the code appears 3 times.

Source/ui_qt/bootablelistdialog.cpp
```diff
@@ -194,26 +219,43 @@ void BootableListDialog::on_awsS3Button_clicked()
...
 		{
 			auto path = string_format("//s3/%s/%s", bucketName.c_str(), item.key.c_str());
 			try
 			{
-				TryRegisteringBootable(path);
+				std::string msg = string_format("Processing: %s (%d/%d)", path.c_str(), i, size);
+				AsyncUpdateStatus(msg);
+				if(TryRegisteringBootable(path))
+				{
+					m_bootables = BootablesDb::CClient::GetInstance().GetBootables(m_sortingMethod);
```
jpd002, line 240: Instead of potentially updating all the models on every single object, could we save a flag that would say that there was a change and call the ResetModel at the end of that lambda if something changed?

jpd002:
Tested it locally and I'm getting a crash that seems to be due to `m_bootables` being modified by the async task while another thread is accessing it (the `cover_loader` thread).

I think fixing [this](https://github.com/jpd002/Play-/pull/866/files/9dbedfd2ef75c67f2cab6fc5d749d204520381cd#diff-6453984fce223492100711ca06c0b6bfR240) will prevent the crash from happening.

Source/ui_qt/bootablelistdialog.cpp
Zer0xFF: we could, just to note, this was done like this to avoid the UI looking stuck with no new entries, im just going to rewrite it so that it updates once at the end.

but we could also add a fake entry to m_bootables, everytime we find a new entry? to keep the UI ticking?

jpd002: Yes, adding entries in the UI as they are being found by the S3 scanner would be nice, but I think it would be better to do it using a different approach. Something like firing an event when a new bootable has been registered. The UI code would be able to catch the event and add a new entry in the list view (or something else). Ideally, the S3 scanning logic should be completely UI agnostic.

jpd002:
Going to merge this as this seems to be working well (not crashing anymore). Improvements can be made in another PR :)

Thanks!

PR #882, opened 2020-04-22
document CCdvdfsv::Invoke592 method 0 structure

Zer0xFF:
I found this through logs, if you try to run OSDSYS from bios, it will print these details.

`sceCdInit: cdvdfsv Ver 0000 cdvdman Ver 0000`


jpd002:
Thanks! Might be useful in the future! 😃

PR #883, opened 2020-04-23
Implement missing blendmode 1002

kernle32dll:
Fixes loading screen not rendering in Burnout 3 (its still stuck tho)

Vulkan implementation is missing as I lack required hardware.

before_patch
before

after_patch
after


jpd002:
Thanks for this!

Vulkan doesn't need to support blending modes explicitly since it's all done in the fragment shader.

PR #885, opened 2020-04-24
Small mc improvements

kernle32dll:
Implements some small memory card related improvements.

The 0 / -1 switch on CMcServ::GetInfo is just niceness - I don't think any game or homebrew actually relies on this behavior.

The Mkdir call is used by the PS2SDK memory card example, and partially fixes that example.

### DIFF
--- a/Source/iop/DirectoryDevice.cpp
+++ b/Source/iop/DirectoryDevice.cpp
@@ -75,3 +75,13 @@ Directory CDirectoryDevice::GetDirectory(const char* devicePath)
 	}
 	return fs::directory_iterator(path);
 }
+
+void CDirectoryDevice::CreateDirectory(const char* devicePath)
+{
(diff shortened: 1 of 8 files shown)


carbot1406:
Hey, I am very happy that someone is pushing this emulator forward! Thank you so much for giving me hope to have a PS 2 emulator on my iPhone one day. If you get a chance I would like to know if it would be possible to emulate Crash Twinsanity at any point on an iPhone 11 Pro Max that is jailbroken.

jpd002:
I'm not too confident about `GetInfo` returning 0/-1 change. I'm not sure how games will react to that and I don't think it brings much to report that a new memory card was inserted since the last time (-1) `GetInfo` was called. Can we leave this one aside?

The rest is good :) I'll merge this once the `GetInfo` commit is removed.

kernle32dll:
Done :)

jpd002:
Awesome! Thank you!

PR #886, opened 2020-04-24
Implement alternative GS CSR register

kernle32dll:
This fix allows the [funslower demo](https://files.scene.org/view/parties/2001/trsac01/wild/soopadoopa-funslower.zip) to run completely (note: this demo does **not** run on PCSX2).

This is a mind boggling one: The demo uses both 0x12001000 and 0x12000400 for the GS CSR register.

I'm puzzled, as 0x12000400 seems to be the correct address indicated in the GS User's Manual (both 3.1 and 6.0). Still, 0x12001000 is also used (for some vsync logic), and 0x12001000 is widely referenced in documentation on the net.

Screenshot from 2020-04-24 12-11-27
Screenshot from 2020-04-24 12-11-37
Screenshot from 2020-04-24 12-12-21


Source/ee/Ee_SubSystem.cpp
```diff
@@ -450,7 +450,7 @@ uint32 CSubSystem::IOPortReadHandler(uint32 nAddress)
 		                         nAddress, m_EE.m_State.nPC);
 	}
 
-	if((nAddress == CINTC::INTC_STAT) || (nAddress == CGSHandler::GS_CSR))
+	if((nAddress == CINTC::INTC_STAT) || (nAddress == CGSHandler::GS_CSR) || (nAddress == CGSHandler::GS_CSR_ALT))
```
jpd002, line 453: This check is for idle loop skipping purposes. 

This demo doesn't seem to check `GS_CSR_ALT (0x12000400)` to wait a frame's end, so we should remove this additional check.

Source/gs/GSHandler.h
```diff
@@ -103,6 +103,11 @@ class CGSHandler
 		GS_CSR = 0x12001000,
 		GS_IMR = 0x12001010,
 		GS_SIGLBLID = 0x12001080,
+
+		// Used by funslower demo
+		// Are the 0x120010XX Addresses above correct?
+		// GS User Manual list them all under 0x120004XX.
+		GS_CSR_ALT = 0x12000400,
```
jpd002, line 110: Please place `GS_CSR_ALT` with the others above and make sure it's sorted by address. Add the comment about it being used in funslower demo on the same line. No need for the comments that asks if the ones above are correct, they are most likely are since all the games are using it.

jpd002:
Nice find :) That demo runs very well in the emulator.

Added a few things that need to be addressed, I'll merge the PR after that.

Thanks!
JP

kernle32dll:
> Added a few things that need to be addressed, I'll merge the PR after that.

Done

> Thanks!

You are welcome :)

Issue #887, opened 2020-04-24
GS receiving junk in gsKit example "fontm"


kernle32dll:
I'm opening this as a separate issue to the [compatibility report](https://github.com/jpd002/Play-Compatibility/issues/1097), because I think this requires some discussion.

Ever since testing the `fontm` example of [gsKit](https://github.com/ps2dev/gsKit), I was intrigued by it not rendering correctly. The graphical glitches vary, but in a nutshell the fonts don't render correctly.

_For the sake of discussion, the example `pixelperfect` (with some mode fixes) exhibits the same following issues. So its not strictly a "font" problem, but straight texture based._

So I began to debug, only to [fall flat on my face](https://github.com/jpd002/Play-/blob/8af8a752b43100bf9d9ad004cdcf04770b813115/Source/gs/GSHandler.cpp#L1193) with an assert first thing. The CLD given is 7 - why is illegal. But still, if removing the assert, it only triggers [the next assert](https://github.com/jpd002/Play-/blob/61b503013f53f005e4af6190222814c59cdc3076/Source/gs/GSH_OpenGL/GSH_OpenGL_Texture.cpp#L171), as it tries to allocate a `32768x32768` texture. Looking at `nData` - its `18446744073642472576` - which is pretty much all `1`s +/- the high half. So somethings clearly wrong here.

On the gsKit side of things, I traced it back to [this loc](https://github.com/ps2dev/gsKit/blob/7061b2e3fdf1bf9e6ce6693b933dde4dbd77a6a6/ee/gs/src/gsTexture.c#L467). But nothing really stands out to me.

So the big question is - what happened in between? I poked around with some breakpoints, but nothing came from that. I did all I could, now I need to inspiration how to debug this further.

Any ideas?

jpd002:
So, do you get a ~0 [here](https://github.com/jpd002/Play-/blob/master/Source/ee/GIF.cpp#L147)?

kernle32dll:
Code does not go there, here is the stack to the first assert:

```
CGSHandler::ProcessCLD GSHandler.cpp:1253
CGSHandler::SyncCLUT GSHandler.cpp:1228
CGSHandler::WriteRegisterImpl GSHandler.cpp:618
CGSH_OpenGL::WriteRegisterImpl GSH_OpenGL.cpp:1837
CGSHandler::WriteRegisterMassivelyImpl GSHandler.cpp:714
CGSHandler::<lambda()>::operator()(void) const GSHandler.cpp:601
[...]

PR #888, opened 2020-04-26
add rom0, host0 devices

Zer0xFF:
ref #887


jpd002:
Working nicely. Very useful if you want to use a `LIBSD` module for homebrew.

PR #889, opened 2020-04-27
Account for textures with 0 buffer width

kernle32dll:
The problem fixed here is the following:

DBZ Budokai Tenkaichi 2 and 3 (at least) use textures, which have a buffer width of 0. This caused the texture cache to get confused, as it effectively build a 0 pixel wide rect.

The other thing that baffled me was the fact that these games also use 0 length "host to local" data transfers. This triggered an assert downstream (since it found an dirty page, but due to the 0 transfer width, there was nothing to actually set dirty).

I don't really see how these textures even work with a 0 buffer size, but they seem to (note, the textures in the menu of Tenkaichi 2 are also kinda animated).

Unrelated: Unfortunately, Tenkaichi 2 still has the all-black texture for characters problem... But I have a hunch what the problem is.

**Screenshots (Tenkaichi 2):**

tenkaichi_2_menu_broken

tenkaichi_2_menu_broken_2

tenkaichi_2_menu_fixed

tenkaichi_2_ingame_broken

tenkaichi_2_ingame_fixed

**Screenshots (Tenkaichi 3):**

tenkaichi_3_menu_broken

tenkaichi_3_menu_fixed

tenkaichi_3_ingame_broken

tenkaichi_3_ingame_fixed

### DIFF
--- a/Source/gs/GSH_Direct3D9/GSH_Direct3D9.cpp
+++ b/Source/gs/GSH_Direct3D9/GSH_Direct3D9.cpp
@@ -83,7 +83,15 @@ void CGSH_Direct3D9::ProcessHostToLocalTransfer()
 		//Find the pages that are touched by this transfer
 		auto transferPageSize = CGsPixelFormats::GetPsmPageSize(bltBuf.nDstPsm);
 
-		uint32 pageCountX = (bltBuf.GetDstWidth() + transferPageSize.first - 1) / transferPageSize.first;
+		// DBZ Budokai Tenkaichi 2 and 3 use invalid (empty) buffer sizes
+		// Account for that, by assuming trxReg.nRRW.
+		auto width = bltBuf.GetDstWidth();
+		if(width == 0)
+		{
+			width = trxReg.nRRW;
+		}
+
+		uint32 pageCountX = (width + transferPageSize.first - 1) / transferPageSize.first;
 		uint32 pageCountY = (trxReg.nRRH + transferPageSize.second - 1) / transferPageSize.second;
 
 		uint32 pageCount = pageCountX * pageCountY;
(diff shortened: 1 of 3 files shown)


kernle32dll:
Pushed a fix - it's all good now.

The previous version did break Suikoden 5 (and probably other games), and did not catch a few edge-cases even in Budokai Tenkaichi 2 (yellow energy sprite at bottom right in title screen - see above) and 3 (arrow in wrong direction at language select - see above).

Screenshot from 2020-04-28 18-47-37
Screenshot from 2020-04-28 18-47-56

boberto5888:
This PR fixes some issues in KH1 and Rugrats that were introduced between [between commits fc7152e2 and 016a4df4](https://github.com/jpd002/Play-/compare/fc7152e2...016a4df4).

---

<p align="center"><sup>click bullet for examples</sup></p>
Kingdom Hearts 1<ul><details><summary> Fixes glitched/missing icons on the world map and in minigames.</summary></ul><p>
Rugrats Royal Ransom.<ul><Details><summary>Fixes broken/flickering textures.</summary></details><Details><summary>Fixes text/images being displayed on top of each other.</summary></details></ul>

jpd002:
So, I've been thinking about this and it might be best to use the transfer/texture width all the time and not just when the buffer width is 0. I'm gonna try that on my end and see what kind of side effect it does.

Buffer width 0 is not really invalid, it's fine for games to use it as long as they keep the transfer/texture height below a page height. Beyond that, I think some pixels might be mixed up. DBZ seems to keep that below the page height, so, it's defined. The Vulkan GS handler doesn't do any particular check for that case and doesn't exhibit this problem.

jpd002:
I've been testing some stuff on my end and I think you're handling this the right way after all! 

Merging! Thanks a lot!

jpd002:
Also, as a note, this fixes other DBZ games on OpenGL (such as Sparking Meteor).

PR #893, opened 2020-04-29
Prevent McServ from leaving memory card folder

### LINKED ISSUE #892: McServ folders not properly jailed
Need for Speed Underground 2 starts the game right of the bat with an ChDir command of `../`. This allows is to write folders outside the memory card folder.

Screenshot from 2020-04-30 00-32-41

This should not be the case ;-)

kernle32dll:
Fixes #892


kernle32dll:
Already solved - closing

Issue #891, opened 2020-04-29
Play libretro core crashes when built with gcc-8


kwyxz:
Using the libretro buildbot recipe to build the Play! libretro core:

`SINGLE_CORE=play NOCLEAN=0 FORCE=YES ./libretro-buildbot-recipe.sh recipes/linux/cores-linux-x64-generic` 

This builds fine and generates a working play_libretro.so with gcc-7 and g++-7 however the following happens with gcc-8 and g++-8 :

`retroarch: symbol lookup error: /usr/local/lib/libretro/play_libretro.so: undefined symbol: _ZNSt10filesystem7__cxx114path14_M_split_cmptsEv`

Zer0xFF:
gcc-8 is an out lier... it reports filesystem support, but the library for it is listed separately,  aka it so doesn't automatically link, that was changed with gcc-9.

if you really want to use 8, you need to add the library to the cmake script.

kwyxz:
Unfortunately gcc-8 is the default version of gcc in at least Debian stable :(

I tried adding this to CMakeLists.txt :
`set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -lstdc++fs" )`
But it looks like it's not enough.

Zer0xFF:
I'd try adding `stdc++fs` directly to the `target_link_libraries` `ui_libretro` cmake

kwyxz:
Thanks, it helped. Since it could be useful to others I created a pull request, #894

PR #894, opened 2020-04-30
Fix symbol error when building libretro core with gcc-8

kwyxz:
As mentioned in #891 a Linux build of the Play libretro core with gcc-8 will crash because of an unknown symbol

`retroarch: symbol lookup error: /usr/local/lib/libretro/play_libretro.so: undefined symbol: _ZNSt10filesystem7__cxx114path14_M_split_cmptsEv`

The issue seems to be specific to gcc-8 and the attached changes fix it.


jpd002:
Looking good! Thanks!

PR #898, opened 2020-05-03
Fix rendering issues in DBZ Budokai

kernle32dll:
~~This PR fixes two visual bugs in DBZ Budokai, distinct by commit:~~

~~Firstly, it implements a missing blend mode, required for some kinda minigame (dunno how to call this otherwise :D).~~

**Edit**: This PR was adjusted, to only contain the changes mentioned below. The blend-mode commit was removed, as a proper fix will be implemented elsewhere.

Screenshot from 2020-05-03 21-03-50

Screenshot from 2020-05-03 21-00-23

The second regards the transparent junk in the background during battle:

If we do a local to local transfer, the destination might not be a framebuffer, but a texture. This is used in the game for some "depth of field" effect during battle.

Screenshot from 2020-05-03 21-05-13

Screenshot from 2020-05-03 20-59-19

### DIFF
--- a/Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
+++ b/Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
@@ -2049,9 +2049,11 @@ void CGSH_OpenGL::ProcessLocalToLocalTransfer()
 		                                           return (framebuffer->m_basePtr == bltBuf.GetDstPtr()) &&
 		                                                  (framebuffer->m_width == bltBuf.GetDstWidth());
 	                                           });
-	if(
-	    srcFramebufferIterator != std::end(m_framebuffers) &&
-	    dstFramebufferIterator != std::end(m_framebuffers))
+
+	bool foundSrc = srcFramebufferIterator != std::end(m_framebuffers);
+	bool foundDest = dstFramebufferIterator != std::end(m_framebuffers);
+
+	if(foundSrc && foundDest)
 	{
 		FlushVertexBuffer();
 		m_renderState.isValid = false;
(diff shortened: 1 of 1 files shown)


Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
```diff
@@ -801,6 +801,13 @@ void CGSH_OpenGL::SetupBlendingFunction(uint64 alphaReg)
 		glBlendColor(0, 0, 0, static_cast<float>(alpha.nFix) / 128.0f);
 		glBlendFuncSeparate(GL_ONE, GL_CONSTANT_ALPHA, GL_ONE, GL_ZERO);
 	}
+	else if((alpha.nA == ALPHABLEND_ABD_CD) && (alpha.nB == ALPHABLEND_ABD_ZERO) && (alpha.nC == ALPHABLEND_C_FIX) && (alpha.nD == ALPHABLEND_ABD_CD))
+	{
+		//1221 -> Cd * (1 + Fix)
+		float bf = static_cast<float>(alpha.nFix) / 128.0f;
+		glBlendColor(bf, bf, bf, bf);
+		glBlendFuncSeparate(GL_DST_COLOR, GL_CONSTANT_COLOR, GL_ONE, GL_ZERO);
```
jpd002, line 809: Is this correct? Wouldn't that give this blending formulae?
`FinalColor = SourceColor*DestColor + ConstColor*DestColor?`

I don't think this blending mode can be emulated using the fixed blending functions. I think we should do something similar to what's done for 1201 temporarily.

Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
```diff
@@ -2071,6 +2080,39 @@ void CGSH_OpenGL::ProcessLocalToLocalTransfer()
 
 		glBindFramebuffer(GL_READ_FRAMEBUFFER, 0);
 	}
+	else if(foundSrc && !foundDest)
+	{
+		auto trxPos = make_convertible<TRXPOS>(m_nReg[GS_REG_TRXPOS]);
```
jpd002, line 2085: We need a call to `FlushVertexBuffer` here to make sure all rendering commands are complete before we read from the framebuffer.

kernle32dll: Fixed

Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
jpd002: Alright, I checked in the frame debugger and this (GL_DST_COLOR, GL_CONSTANT_COLOR) works in this particular case because all source pixels are 1 (white) when the cone is rendered. It effectively becomes `FinalColor = DestColor * 1 + ConstColor * DestColor` which is the same as `Cd * (1 + Fix)`.

But if the source color is anything other than 1, it's going to give a wrong result.

To make this work in the general use case, I think a good option would be to have the fragment shader output 1 in `gl_FragColor` when this blending mode is selected since we know the source color is not used in the final output.

kernle32dll: I saw you working on this on a separate branch. As I lack the time to implement this right now - should I remove the commit, and you fix the blend mode on your branch?

jpd002: Yes please! The other commits are good, so, I will be able to merge the PR.

jpd002: Thanks!

kernle32dll: Done - you are welcome. Rebased onto master for good measure.

jpd002:
Tested with a few other games, seems to be working fine :)

Thanks!

PR #902, opened 2020-05-10
travis linux parallel build

Zer0xFF:
the arm64 build seem to have 32 threads at their disposal, so why not use them :)

I checked other builds, this doesn't seem to have any effect, likely due default parallel building.
edit: didn't test android, though I'm not sure how we'd pass this to it, or if we even should, since we don't interact with cmake directly.


jpd002:
Very nice!!! Arm64 builds are blazing fast now!!

PR #908, opened 2020-05-21
improve VU MOVE/RNEXT instruction

Zer0xFF:
RNEXT behaviour now matches the hardware test results

I found the behaviour here https://web.archive.org/web/20060908233922/http://project-fao.org/doc/rnd.html which was linked to in PCSX2 source.


jpd002:
Very cool, thanks a lot! :)

Issue #906, opened 2020-05-20
Build error on Linux x86 32bit


guoyunhe:
```
[   69s] /usr/lib/gcc/i586-suse-linux/9/include/xmmintrin.h:878:1: error: inlining failed in call to always_inline 'void _MM_SET_FLUSH_ZERO_MODE(unsigned int)': target specific option mismatch
[   69s]   878 | _MM_SET_FLUSH_ZERO_MODE (unsigned int __mode)
[   69s]       | ^~~~~~~~~~~~~~~~~~~~~~~
[   69s] /home/abuild/rpmbuild/BUILD/libretro-play-0~git20200519/Source/FpUtils.cpp:32:25: note: called from here
[   69s]    32 |  _MM_SET_FLUSH_ZERO_MODE(_MM_FLUSH_ZERO_ON);
[   69s]       |  ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
[   69s] In file included from /home/abuild/rpmbuild/BUILD/libretro-play-0~git20200519/Source/FpUtils.cpp:22:
[   69s] /usr/lib/gcc/i586-suse-linux/9/include/xmmintrin.h:853:1: error: inlining failed in call to always_inline 'void _mm_setcsr(unsigned int)': target specific option mismatch
[   69s]   853 | _mm_setcsr (unsigned int __I)
[   69s]       | ^~~~~~~~~~
[   69s] In file included from /home/abuild/rpmbuild/BUILD/libretro-play-0~git20200519/Source/FpUtils.cpp:23:
[   69s] /home/abuild/rpmbuild/BUILD/libretro-play-0~git20200519/Source/FpUtils.cpp:31:2: note: called from here
[   69s]    31 |  _MM_SET_DENORMALS_ZERO_MODE(_MM_DENORMALS_ZERO_ON);
[   69s]       |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~
[   69s] In file included from /home/abuild/rpmbuild/BUILD/libretro-play-0~git20200519/Source/FpUtils.cpp:22:
[   69s] /usr/lib/gcc/i586-suse-linux/9/include/xmmintrin.h:821:1: error: inlining failed in call to always_inline 'unsigned int _mm_getcsr()': target specific option mismatch
[   69s]   821 | _mm_getcsr (void)
[   69s]       | ^~~~~~~~~~
[...]

PR #909, opened 2020-05-22
fix FreeBSD builds

Zer0xFF:
CMake `set()` seems to create a list if items are separated by space.

might be worthwhile setting up FreeBSD CI as well, I'll see what I can do when I get the chance.

reference: #906


jpd002:
Thanks for taking care of this! :)

PR #910, opened 2020-05-23
FreeBSD CI

Zer0xFF:
Muahaha a lucky break as Travis has recently added FreeBSD support (even documentation isn't up yet) 👿 BSD
few notes
- FreeBSD travis image seems to contain gcc 10, which is causing build problems, mostly due to ["Reduced header dependencies, leading to faster compilation for some code"](https://gcc.gnu.org/gcc-10/changes.html), which we will need to address eventually.
- I don't really know how the packaging system work on FreeBSD systems yet, so for now, CI will just be used to ensure no breaking changes are introduced

TODOs:
- [x] Test the changes on GhostBSD (local setup)
  - we crash when opening controller dialog, Qt 5.14 issue, caused by `QGroupBox` replacing it with `QFrame` fixes the crash (not sure if we should just drop `QGroupBox` or wait and see if its fixed upstream?)(since we're not doing releases, this is a non issue)
  -  issue has been reported already and seems to effect all OSes https://bugreports.qt.io/browse/QTBUG-82287 I think we should just wait for a fix, since `QGroupBox` and `QFrame` aren't the same, I only used it for a quick test
  - Note: GhostBSD repo seem more upto date than travis FreeBSD, because of that, travis is using older Qt which isn't effect by the issue, but we also can't build Vulkan yet since its using an older header version (so im removing 208486a)
  - Gamepad doesnt work, but it doesnt work on Firefox as well, i suspect its kernel/OS config issue, and im not familiar with FreeBSD to diagnose/fix
- [x] Test the changes on linux
- [x] Check, the status of Vulkan on FreeBSD

Side note:
it seems that clang default compiler is actually clang, not gcc.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -39,6 +39,8 @@ travis_before_install()
     elif [ "$TARGET_OS" = "Android" ]; then
         echo y | sdkmanager 'ndk;21.0.6113669'
         echo y | sdkmanager 'cmake;3.10.2.4988404'
+    elif [ "$TARGET_OS" = "FREEBSD" ]; then
+        su -m root -c 'pkg install -y cmake qt5 gcc9 evdev-proto'
     fi;
 
     git submodule update -q --init --recursive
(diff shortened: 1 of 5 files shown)


jpd002:
Very cool :) Compiling with GCC is fine. We can readjust that if needed later. Thanks!

PR #917, opened 2020-06-08
WIP: Fix GCC10 link errors

guoyunhe:
I need to test this...


Zer0xFF:
This probably won't work as is, if this is to work, you need to another one in the framework repo as well

Issue #928, opened 2020-08-12
Evdev Input Handler is broken


Hirato:
The Evdev handlers currently does not report any events.
The Key issue seems to be the usage of pselect in GamePadDeviceListener and GamePadInputEventListener
pselect consistently returns that the device the open file descriptors as non-modified (return value of 0), and so no events are ever received.

The pselect invocations also appear to be done incorrectly.
In that the number of file descriptors to check is `fd + 1`, rather than the constant 1

e.g.

     		if(pselect(fd + 1, &fds, NULL, NULL, &ts, &mask) == 0) continue;

should be

		if(pselect(1, &fds, NULL, NULL, &ts, &mask) == 0) continue;

Additionally. Axis values for evdev range from -32767 to 32767.
Also, Trigger buttons, return an `abs->flat` value of 0, but a range of 0-1023 (At least for my Xbox One controller).
This causes the input handler to incorrectly handle them as "POVHAT" buttons instead of as Axis.

I can't speak for the rest of the project, but inputeventselectiondialog at the very least appears to expect that all axis is cleanly mapped within a 0-255 range.
If so, you may want to handle axis as `(0x7FFF + value) >> 8` (-32767..32767 -> 0..255) and triggers as `0x80 + (value >> 3)` (0..1023 -> 128..255)

You can see some of my playing around here: https://gist.github.com/Hirato/05bb349e4bfd58cf33273293b97b860a
It's enough for me to be able to configure my controller.

Zer0xFF:
```c++
pselect(1,..);
```
doesnt make much sense, as 1 is stdout file descriptor. `fd` here event file descriptor.


I've had issues with ranges, as they dont seem to have a standard range, sometime they range from 0 to x, sometimes its -x to x, so the start cane vary from being x/2 or 0 but this will definitely need a 2nd look.

thanks for reporting this, ill see if i can replicate the issues, and see what i can do.

that gist would have been better as a commit, as i could have commented on the lines themselves,
anyway, making pselect blocking can be a problem, as it can hold a thread up during app termination (though perhaps, listening for app termination might work, making that a better solution)

kenshen112:
Can you run evtest and post the top half of it looking at your device? I say top half as in the static listing of buttons etc not the keys currently being pressed

Hirato:
> ```c++
> pselect(1,..);
> ```
> 
> doesnt make much sense, as 1 is stdout file descriptor. `fd` here event file descriptor.

[...]

PR #929, opened 2020-08-13
fix pselect usage

Hirato:
According to the manpage:

Note well: Upon return, each of the file descriptor sets is modified in place to indicate which file descriptors are currently "ready".  Thus, if using **select() within a loop, the sets must be reinitialized** before each call. 
The implementation of the fd_set arguments as value-result arguments is a design error that is avoided in poll(2) and epoll(7).

So, I've moved the FD_ZERO and the FD_ZET into the loop itself before the `pselect` call


Zer0xFF:
nicely spotted, does this fix your issue with #928?

just to note, we had to switch from `poll` to `select` to be able to support compiling for FreeBSD.

Hirato:
It fixes my controller not responding at all, but there are still many issues with how the axis are mapped (and missing support for trigger buttons)

jpd002:
Thank you very much for this!

PR #979, opened 2020-11-11
Don't care for mode on reading files from optical media

kernle32dll:
Mafia tries to open a file on disc as RW, and expects it to work.

This PR allows the game to save, and thus get into gameplay. Altho there are still other issues with the game.


jpd002:
Fantastic! Thanks!

PR #981, opened 2020-11-22
update vulkan sdk download link

Zer0xFF:
Link for SDK seems to have changed.

the window build was uneffected, since its using cached version.


jpd002:
Oh, I didn’t see the failures on Travis. Thanks again for your help!

Zer0xFF:
its easy to miss, I only noticed because the OSX download link was dead on the site.

PR #982, opened 2020-11-28
Handle DMA transfers with 0 QWC

kernle32dll:
DMA always transfers at least one QW, regardless of whats in the QWC register. If QWC is 0, it actually underflows the QWC register after sending the first QW.

Enter the Matrix migitates this, as it sets QWC every frame.

Thanks to refraction for checking this on real hardware!

### DIFF
--- a/Source/ee/Dmac_Channel.cpp
+++ b/Source/ee/Dmac_Channel.cpp
@@ -165,7 +165,19 @@ void CChannel::ExecuteNormal()
 		break;
 	}
 
+	// When starting a normal transfer, always at least one
+	// QW is send, regardless of the value of the QWC register.
+	// In case of a QWC of 0, this causes the QWC register to
+	// underflow down the line, which is intentional (actually
+	// happens on real hardware).
+	// Games that do exhibit a normal transfer with 0 QWC
+	// (Enter the Matrix, Star Wars Racer Revenge) mitigate
+	// this by setting the QWC register each frame.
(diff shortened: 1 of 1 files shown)


jpd002:
Thanks for this! The PS2 hardware will never cease to amaze me!

Out of curiousity: Does the STR bit of CHCR gets cleared if QWC underflows in this condition? 

Here, it will check if QWC is 0, but if it's 0xFFF, I guess the transfer could keep going? https://github.com/jpd002/Play-/pull/982/files#diff-77685f93f9fd2f62094dc2485317e81dac693cac6838642f255806c71a590b57L188

Merging this anyways since it fixes games that use this behavior.

kernle32dll:
> Thanks for this! The PS2 hardware will never cease to amaze me!

Indeed, me too :D

> Out of curiousity: Does the STR bit of CHCR gets cleared if QWC underflows in this condition?

Nope, it goes on till it wraps around at QWC 1->0 again. So in total it transfers 0x10000 (0xFFFF + 1) QWC, if unchecked.

Speaking of it, the way its currently implemented in Play!, it would actually first transfer 1 QW, and then 0xFFFF QW. That might not be the correct implementation in the long run (as in QW per cycle, not total amount). But since no game should rely on 0 QWC behavior anyway, this should be fine. Just one thing to keep in mind for the future.

jpd002:
Sounds good! Thanks for the explanation!

PR #989, opened 2020-12-12
Fix: Windows S3 AWS BIN

Zer0xFF:
fix cue/bin path concat on windows for S3, on windows the path will look like /s3/bucket\game.bin so aws s3 can't find the game

### DIFF
--- a/Source/DiskUtils.cpp
+++ b/Source/DiskUtils.cpp
@@ -33,6 +33,7 @@ static Framework::CStream* CreateImageStream(const fs::path& imagePath)
 	{
 #ifdef HAS_AMAZON_S3
 		auto fullObjectPath = std::string(imagePathString.c_str() + strlen(s3ImagePathPrefix));
+		std::replace(fullObjectPath.begin(), fullObjectPath.end(), '\\', '/');
 		auto objectPathPos = fullObjectPath.find('/');
 		if(objectPathPos == std::string::npos)
 		{


jpd002:
Good catch! That's what I get for trying to use std::filesystem for S3 paths 😛 

If it works, I think it would be better to add some logic before this line here: 
https://github.com/jpd002/Play-/blob/a4322e8510d415b074baaa581444b3d08b63cb2f/Source/DiskUtils.cpp#L42

Like, something to fix the slashes, since this is specific to the S3 driver. If there's another disc image format that does funky path manipulation like CUE/BIN, I wouldn't want it to be aware of the specificities of S3.

Also, could you rebase your branch on the latest changes on master? I noticed the GitHub Actions workflows weren't triggering on pull requests and the last commit should fix that.

Thanks!

Zer0xFF:
can probably just use `std::replace` `\\` with `/`  before constructing`CS3ObjectStream` and its probably the better solution, since the current commit might actually break local win32 cue/bin loading (dont have anything locally to test atm)

jpd002:
Sorry, I mislead you, the std::replace needs to happen before this:
https://github.com/jpd002/Play-/blob/a4322e8510d415b074baaa581444b3d08b63cb2f/Source/DiskUtils.cpp#L36

If the objects are in the root of the S3 bucket, it will never find the first slash to separate the bucket name from the object key name.

Zer0xFF:
> Sorry, I mislead you, the std::replace needs to happen before this:
> https://github.com/jpd002/Play-/blob/a4322e8510d415b074baaa581444b3d08b63cb2f/Source/DiskUtils.cpp#L36
> 
> If the objects are in the root of the S3 bucket, it will never find the first slash to separate the bucket name from the object key name.

Haha, I should have spotted that myself as well, I'll push an update once I'm behind a keyboard.

jpd002:
Thanks a bunch!

PR #991, opened 2020-12-14
GitHub actions

Zer0xFF:
GitHub actions fail on my fork, since it tries to upload artifcats to S3, added few checks to stop that,
I've also added action to upload to github artifacts, both useful for branches and forks.

P.S I also dont like how each job is its own workflow, because of that, each branch creates 4 different workflows, and all the artificats are inside each workflows, as oppose just 1 workflow with multiple jobs.
e.g i joined up all thwe work flows into 1, and now all the artifcats are listed under 1 workflow here https://github.com/Zer0xFF/Play-/actions and if you look at the workflow list, its 1 action for all of them https://github.com/Zer0xFF/Play-/actions/runs/421228891 with a workflow like this, you can introduce a new dependent job that you can use to automatically trigger aws update (since as you recall, when you push a change to master, if let say iOS fails, but the reset successed, then the link on the site for iOS will be dead)
e.g

```yaml
publish:
    needs: [build_linux, build_android, build_macos, build_ios]
    on:
      push:
        branches:
        - master
    runs-on: ubuntu-latest
    steps:
      - name: Trigger site update
        if: ${{ env.AWS_ACCESS_KEY_ID != null}}
        run: curl purei.org/?secret.trigger.update
    env:
        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```

as you can tell, i already have a commit for that, though i didnt push it to this, since i suppose that falls under preference, if you'd prefer it this way, i can attach it to this, https://github.com/Zer0xFF/Play-/commit/5b27b19b4e0390987586ce4208aa1e06e88f2c2a

P.S iOS fails on m work, though i would assume thats only on my fork, since i didnt change anything related to how cmds are run.

### DIFF
--- a/.github/workflows/build-android.yaml
+++ b/.github/workflows/build-android.yaml
@@ -49,17 +49,39 @@ jobs:
         export ANDROID_TOOLCHAIN_FILE=${ANDROID_NDK}/build/cmake/android.toolchain.cmake
         cd build_retro
         ./android_build.sh
-    - name: Prepare APKs for Distribution
-      if: ${{ github.event_name != 'pull_request' && matrix.build-type == 'apk' }}
+    - name: Prepare Unsigned APKs for Distribution
+      if: ${{ matrix.build-type == 'apk'}}
       run: |
+        export ANDROID_BUILD_TOOLS=$ANDROID_HOME/build-tools/29.0.3
         cp build_android/build/outputs/apk/release/Play-release-unsigned.apk .
         cp Play-release-unsigned.apk Play-release.apk
+        $ANDROID_BUILD_TOOLS/zipalign -c -v 4 Play-release-unsigned.apk
+    - name: Prepare Signed APKs for Distribution
+      if: ${{ github.event_name != 'pull_request' && matrix.build-type == 'apk'}}
+      run: |
         export ANDROID_BUILD_TOOLS=$ANDROID_HOME/build-tools/29.0.3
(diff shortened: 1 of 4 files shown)


jpd002:
Oh yeah, I forgot about that use case. I just wish there was a better way to specify that we don't want to sign/distribute. Thinking about this, could we get rid of the `github.event_name != 'pull_request'` condition if we're testing the secrets directly? They shouldn't be available when building pull requests anyways. And the artifacts are a nice addition!

I also like that idea of being able to publish the update when every job is complete, but I'm not a big fan of clumping up everything in a single file (that file is kinda huge with everything in it). Also, the Windows builds are not on GitHub Actions ATM, so, we'd need to migrate that over here before making that work. So, I think I'll pass on that for now.

Zer0xFF:
> Oh yeah, I forgot about that use case. I just wish there was a better way to specify that we don't want to sign/distribute. Thinking about this, could we get rid of the `github.event_name != 'pull_request'` condition if we're testing the secrets directly? They shouldn't be available when building pull requests anyways. And the artifacts are a nice addition!

ya the event check is redundant, I mainly kept it, since i didnt want to make any assumptions about how well written this system is 🙈 

> I also like that idea of being able to publish the update when every job is complete, but I'm not a big fan of clumping up everything in a single file (that file is kinda huge with everything in it)

ya this was really unfortunity, as far as i can tell, there is no way to reference another file (like an `#include`), the closest there is, is actions, but actions themselves can't `use` other actions, so for now atleast there is no way around this.

ill update this branch now and remove the `github.event_name != 'pull_request'` checks.

jpd002:
Thanks!

PR #992, opened 2020-12-24
Don't enable SSE on ppc64

pkubaj:
SSE is only for i386 / amd64.


Zer0xFF:
Hi @pkubaj,

 `Play!` doesnt have support for a PPC JIT compiler, and there is no interpreter, so unless you have plans to work on one, adding rules for PPC would only add confusion to whats support and whats not.

PR #995, opened 2020-12-31
make OSX build universal

### LINKED ISSUE #988: Play build errors on Apple Silicon
In file included from /Users/stlouiscpht/Play-/Source/PS2VM.cpp:3:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/memory:2368:5: error: 
      aligned deallocation function of type 'void (void *, std::align_val_t)
      noexcept' is only available on macOS 10.14 or newer
    delete __ptr;
    ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/memory:3033:28: error: 
      aligned allocation function of type
      'void *(std::size_t, std::align_val_t)' is only available on macOS 10.14
      or newer
    return unique_ptr<_Tp>(new _Tp(_VSTD::forward<_Args>(__args)...));
                           ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/memory:3033:28: error: 
      aligned deallocation function of type 'void (void *, std::align_val_t)
      noexcept' is only available on macOS 10.14 or newer
    return unique_ptr<_Tp>(new _Tp(_VSTD::forward<_Args>(__args)...));
                           ^
3 errors generated.

Zer0xFF:
I was bored earlier and decided to see if this works.... *spoiler alert*, it did*.

few things to note,
1) the generated app worked on a local tested Intel build (Mac Air 2014), I dont have M1 to test with.
2) ~~the CI generated app doesnt open, it complains about being corrupt, but it works when i use `xattr -cr Play.app` (i think this might have something to do with big sur not allowing untrusted apps from the internet, or unsigned release app?)~~ can confirm this is the case for any artifcats from previous PRs, so this is expected.
3) this solution works, though its a bit "messy", ideally we'd have an official Qt UB we can use, but that might be a while, since they "have to do it properly"

closes #988 

edit:
note, the shell script was made based on my local setup, thus the few warning on CI, we can "fix" those by uncommenting those lines (we can really comment all of `lib/Qt*.framework` par the few we actually use QtCore, QtSvg, QtGui, QtWidgets, QtDBus, QtPrinterSupport XD) (P.S the reason I didnt bother tailoring it to our specific setup, it to avoid it "randomly" breaking because we decided to use something new from Qt, plus beside the warning errors, there is actually no downside to this)


StLouisCPhT:
I haven't tested this on my M1 yet, but looks good to me. Only thing I'm not sure about is the minimum version. Is there any particular reason why 11.0 is preferred over 10.14?

StLouisCPhT:
Looks good to me. Only thing I'm not sure about is the minimum version. Is there any particular reason why 11.0 was used instead of 10.14?

Zer0xFF:
> Is there any particular reason why the minimum deployment version is 11.0 instead of 10.14?

because the build prints a warning when using a lower value.
which in retrospect is odd, since i was getting that warning on 10.15 when this [page](https://keith.github.io/xcode-man-pages/pthread_jit_write_protect_np.3.html) suggests it was added on 10.15

kode54:
The minimum version should be fine to set to 10.14 for this Universal build, even with the warning. Though the arm64 variant will only run on 11.0 or newer anyway. I tested building arm64, and it fails with errors if the SDK version isn't bumped to at least 10.14.

jpd002:
Agreed with the 10.14 minimum version. Besides that I think it would good to merge.

Zer0xFF:
woohoo, i didnt expect to be 5 month down the line with official solution from Qt in sight.
ill change it to `10.14` and add Qt caching.

jpd002:
Yep, it's quite sad that they still didn't provide an official solution!

Thanks a bunch for this!

StLouisCPhT:
Considering that QT5 is now considered depreciated and official QT6 macOS arm support is not planned to be released until 6.2 (instead of homebrew's release of the current code), I don't think it is surprising that a UB QT would still be aways off and only available from QT itself, like the universal build of cmake is only available from cmake.org.

PR #998, opened 2021-01-10
Qt debugger Fixes

### LINKED ISSUE #997: Qt Debugger Issues
Opening this to document the issues I've seen using the new Qt debugger:

- [x] Frame Debugger: Draw kicks are not highlighted.
- [x] Frame Debugger: Navigating draw kicks (using prev/next buttons) doesn't change the selected item in the list to the current draw kick.
- [x] Frame Debugger: Draw call highlight doesn't work properly if resolution multiplier is greater than 1.
- [x] Debugger: If we close views (ex.: threads or functions) and reopen them again, the contents are not visible anymore.

Probably not going to address these ATM, so, if you want to take a look at it @Zer0xFF, lemme know. Also, I've been doing some tweaks in the `debugger_adjustments` branch, just tell me if you see something strange.

Zer0xFF:
addresses issues mention in #997 

closes #997

### DIFF
--- a/Source/ui_qt/DebugSupport/DebugView.cpp
+++ b/Source/ui_qt/DebugSupport/DebugView.cpp
@@ -2,7 +2,7 @@
 #include "BiosDebugInfoProvider.h"
 #include <QVBoxLayout>
 
-CDebugView::CDebugView(QMdiArea* parent, CVirtualMachine& virtualMachine, CMIPS* ctx,
+CDebugView::CDebugView(QWidget* parent, QMdiArea* mdiArea, CVirtualMachine& virtualMachine, CMIPS* ctx,
                        const StepFunction& stepFunction, CBiosDebugInfoProvider* biosDebugInfoProvider, const char* name, int size, CQtDisAsmTableModel::DISASM_TYPE disAsmType)
     : m_virtualMachine(virtualMachine)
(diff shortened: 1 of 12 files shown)


Source/ui_qt/DebugSupport/DebugView.cpp
```diff
@@ -2,7 +2,7 @@
 #include "BiosDebugInfoProvider.h"
 #include <QVBoxLayout>
 
-CDebugView::CDebugView(QMdiArea* parent, CVirtualMachine& virtualMachine, CMIPS* ctx,
+CDebugView::CDebugView(QWidget* parent, QMdiArea* midArea, CVirtualMachine& virtualMachine, CMIPS* ctx,
```
jpd002, line 5: I guess this is a mistake, but the variable is named `midArea`. Shouldn't it be `mdiArea`?

jpd002:
Looking good, will merge after the little mistake is corrected. Thanks!

Source/ui_qt/DebugSupport/DebugView.cpp
Zer0xFF: indeed, changes pushed.

jpd002:
Danke!

PR #1006, opened 2021-01-22
let Qt handle Cell sizes

Zer0xFF:
related #997 

Cells will resize to content, thus wont show ellipsis on MacOS due to different font processing.


jpd002:
Humm, just tried this and while it fixes 8-bits integers, it seems to break 32-bits integer and 32-bits float visualizations on macOS. Do you have the same result on your end?

Zer0xFF:
Yes, i was just able to reproduce it,
for my test i had bytes per line set to auto, which causes `AutoColumn()` to be called, which will calls `Redraw()`, so this never happened to me before, but setting it to 16/32 I was able to reproduce it.

jpd002:
Working like a charm! Thanks!

PR #1009, opened 2021-01-27
remove assertion code in release

bigianb:
glCheckFramebufferStatus is called in release when the result is only used for an assertion.
Moving this into the assertion gives a noticeable speed increase.

In Baldurs Gate Dark Alliance, on the bartender dialog scene (first dialog of the game) this give an FPS increase from 40fps to 53 fps on my machine.


Zer0xFF:
nice catch, im getting really impressive results,
Megaman X8: generally runs at 60fps, but the fps drops 50fps during boss fights or when a text dialog appears, however with PR its consistently at 60fps
FFXII: Certain areas with low FPS are now at 60fps

image
image

jpd002:
Nice catch indeed! Also getting the same result on my end.

Before:
image
After:
image

Who would have thought that this would make such a huge difference? I did some research and couldn't find a super clear answer, but it seems `glCheckFramebufferStatus` can cause a pipeline flush, so I guess that's why it was causing the slowdown. I'll have to try on Android and iOS later to see if it gives a similar improvement.

Thanks a lot!

darkpigeon:
I think this fix only benefits the windows build
Nothing changed on android and ios as i've tested so far with those games mentioned above.

PR #1012, opened 2021-02-02
correctly use initialisation flag

bigianb:
The m_pageOffsetsInitialized flag is not used in the 2 template specialisation functions for 4 and 8bpp. That makes constructing the PixelIndexer object expensive.
The speedup is modest but worthwhile .. approx 10% in my testing 40fps -> 44fps in my Baldurs Gate test scene.


Zer0xFF:
really quick test, since i was interested to see how this effects text rendering (since i made the assumption they'd use 4/8bpp textures, it is an assumption, I will do a frame dump later to confirm, but it seems to hold up based on the results)
image
image

jpd002:
Ouch, what a bad mistake! And another great find! 😄 

Thanks again!

PR #1013, opened 2021-02-02
Win32: GitHub action build

Zer0xFF:
Move windows build to Github Action CI.

help unify CI builds, and Github Action CI can run 32/64 bits simultaneously resulting in faster overall build time.

thinking out loud, i wonder if we can break android build into seperate arches based on jobs, then combine them into 1 apk at the end, since jobs can have dependencies on each other, though i doubt the build system was designed to work like this.

edit:
so... i tried the android approach, and it cool and it works XD
https://github.com/jpd002/Play-/compare/master...Zer0xFF:actions_psf5
though it only saves about 5 minutes of build time (15mins down to 10 mins), so I dont think its really worth the effort of maintaining a CI system like that)


.github/workflows/build-windows.yaml
```diff
@@ -0,0 +1,108 @@
...
+        makensis.exe ./installer_win32/${{ matrix.installer-script}}
+    - name: Upload a Build Artifact Installer
+      uses: actions/upload-artifact@v2
+      with:
+        name: Play_Windows_${{ matrix.target-platform }}_Installer
+        path: installer_win32\*.exe
+    - name: Upload a Build Artifact Libretro Core
+      uses: actions/upload-artifact@v2
+      with:
+        name: Play_Windows_${{ matrix.target-platform }}_libreto
```
jpd002, line 96: Typo here, should be `libretro`

jpd002:
Nice work! :smiley: I was a bit lazy and didn't migrate the Windows build to GitHub Actions, but it's much cleaner that way. Thanks a lot!

> though it only saves about 5 minutes of build time (15mins down to 10 mins), so I dont think its really worth the effort of maintaining a CI system like that)

Pretty cool indeed, but I agree with your conclusion! 😄

jpd002:
Humm, seems there's a little problem with the S3 upload:
image

Zer0xFF:
gaaa, ill look into it.

though, I noticed your attempt, to note, this wont work still, since windows build runs on powershell, though we can probably specify it to use cmd for that step, might make life easier

PR #1014, opened 2021-02-03
Set correct defaults for SMODE, DISPLAY 1 and 2.

kernle32dll:
Fixes the ["PS2 Coding is good for you" Homebrew Demo](http://www.pouet.net/prod.php?which=32469).

Turns out, the game never sets any of the 3 mentioned registers, and assumes them to be set correctly. My guess is, that this happens via the BIOS on console, either implicitly (not getting reset) or explicitly (based on PAL/NTSC of game, etc.).

The values are calculated via [GsSetDefaultDisplayEnv from ps2sdk](https://github.com/ps2dev/ps2sdk/blob/3c4e3ef0df01c49bdb4620c32d3496eeab5f7f81/ee/libgs/src/draw.c#L74), and is correct for NTSC, interlacing and fullframe.

### DIFF
--- a/Source/gs/GSHandler.cpp
+++ b/Source/gs/GSHandler.cpp
@@ -147,15 +147,15 @@ void CGSHandler::ResetBase()
 	memset(m_pRAM, 0, RAMSIZE);
 	memset(m_pCLUT, 0, CLUTSIZE);
 	m_nPMODE = 0;
-	m_nSMODE2 = 0;
+	m_nSMODE2 = 0x3; // Interlacing with fullframe
 	m_nDISPFB1.heldValue = 0;
 	m_nDISPFB1.value.q = 0;
 	m_nDISPLAY1.heldValue = 0;
-	m_nDISPLAY1.value.q = 0;
+	m_nDISPLAY1.value.q = 0x1BF9FF72617467;
 	m_nDISPFB2.heldValue = 0;
 	m_nDISPFB2.value.q = 0;
 	m_nDISPLAY2.heldValue = 0;
-	m_nDISPLAY2.value.q = 0;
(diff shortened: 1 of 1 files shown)


Source/gs/GSHandler.cpp
```diff
@@ -147,15 +147,18 @@ void CGSHandler::ResetBase()
...
-	m_nSMODE2 = 0;
+	m_nSMODE2 = 0x3; // Interlacing with fullframe
 	m_nDISPFB1.heldValue = 0;
 	m_nDISPFB1.value.q = 0;
-	m_nDISPLAY1.heldValue = 0;
-	m_nDISPLAY1.value.q = 0;
+	m_nDISPLAY1.heldValue = 0x0183227c;
+	m_nDISPLAY1.value.d0 = m_nDISPLAY2.heldValue;
+	m_nDISPLAY1.value.d1 = 0x001bf9ff;
+	m_nDISPLAY1.heldValue = m_nDISPLAY1.value.d0;
```
jpd002, line 156: Held value can probably stay 0. It's just a temporary value that is used when a game writes to that register.

I think it would be nicer to set the `q` member as a whole instead of setting individual `d` values. ie.: `m_nDISPLAY1.value.q = 0x001bf9ff0183227c`.

jpd002:
Would make sense that these regs would be initialized by the BIOS. Even if they aren't, I'm not against having these default values in there. 😄 Will merge after the comment is resolved.

Thanks!

Source/gs/GSHandler.cpp
kernle32dll: Fixed and rebased :)

PR #1015, opened 2021-02-04
SSE version of column unswizzle for PSM8

bigianb:
Needs work .. needs an ARM Neon implementation. Also should implement for the other pixel formats.
Still, the result is impressive locally.
master:
image

with PR:
image

### DIFF
--- a/Source/gs/GSH_OpenGL/GSH_OpenGL_Texture.cpp
+++ b/Source/gs/GSH_OpenGL/GSH_OpenGL_Texture.cpp
@@ -25,7 +25,7 @@ void CGSH_OpenGL::SetupTextureUpdaters()
 	m_textureUpdater[PSMCT32_UNK] = &CGSH_OpenGL::TexUpdater_Psm32;
 	m_textureUpdater[PSMCT24_UNK] = &CGSH_OpenGL::TexUpdater_Psm32;
 	m_textureUpdater[PSMCT16S] = &CGSH_OpenGL::TexUpdater_Psm16<CGsPixelFormats::CPixelIndexorPSMCT16S>;
-	m_textureUpdater[PSMT8] = &CGSH_OpenGL::TexUpdater_Psm48<CGsPixelFormats::CPixelIndexorPSMT8>;
+	m_textureUpdater[PSMT8] = &CGSH_OpenGL::TexUpdater_Psm8;
 	m_textureUpdater[PSMT4] = &CGSH_OpenGL::TexUpdater_Psm48<CGsPixelFormats::CPixelIndexorPSMT4>;
 	m_textureUpdater[PSMT8H] = &CGSH_OpenGL::TexUpdater_Psm48H<24, 0xFF>;
 	m_textureUpdater[PSMT4HL] = &CGSH_OpenGL::TexUpdater_Psm48H<24, 0x0F>;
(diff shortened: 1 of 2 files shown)


Zer0xFF:
producing good results in Champions of Norrath, which uses the same engine as BGDA
running at x4 resolution
image
image

thinking out loud, wouldn't it either be better to refactor the code to encourage the compiler to generate equivalent code without needing to use arch specific instructions? or use the codegen to create this function?

jpd002:
Very interesting approach! I never thought of doing tex unswizzles that way!

We'll definitely need a ARM version for this, I wonder if something like [this](https://github.com/DLTcollab/sse2neon) could help us. Will probably help a lot on mobile.

Would there opportunities for further optimizations if we use AVX instructions with 256-bit registers?

bigianb:
I looked at that project and was also reading up on neon this weekend. I think we can use the neon instruction that read interleaved bytes directly and then some transpose instructions to do it better than sse. Mind you I’m sure the sse can be optimised a lot ... I got lost a few times trying to figure it out :) from micro profiling the big win comes from the wider memory access ie reading 16 bytes at a time rather than byte at a time is a big win (and you know the transfer will be aligned on column bounds). I think the benefit from reading 128 bits at a time to 512 will be minimal. I think more interesting may be to write the transfer to gs routines in the same manner.

bigianb:
The micro test I’m using is here: https://github.com/bigianb/ps2-speedtests/tree/main/hostTests/HostTests/GSTransferTests
It runs the baseline code extracted from play and then the sse version and times them. I’ll work on the neon version there too on my raspberry pi... unless someone does it first :)

rcaridade145:
I would like to bring your attention also to https://github.com/richgel999/CppSPMD_Fast .
In theory AVX versus SSE only means that you're half using a register.

bigianb:
I've written the neon version in my test harness.  https://github.com/bigianb/ps2-speedtests/tree/main/hostTests/HostTests/GSTransferTests
On the raspberrypi 3+ the speedup is from 33.4 seconds with the baseline (current Play code) to 2.7 seconds with the NEON code.
I'll update this PR when I get time.

rcaridade145:
Does including -ftree-vectorize help the performance of  C versions ?

bigianb:
> Does including -ftree-vectorize help the performance of C versions ?

I think the logic is too complex. From my experience the tree vecorisation can only really deal with basic loops. Try it and see though.

rcaridade145:
> 
> 
> > Does including -ftree-vectorize help the performance of C versions ?
> 
> I think the logic is too complex. From my experience the tree vecorisation can only really deal with basic loops. Try it and see though.

Ok. Will try later today after work. Great work! :)

jpd002:
Very promising! I'll give it a spin on a device of mine to see the result!

tellowkrinkle:
Side note: If you're willing to work only with Clang, Clang has a `__builtin_shufflevector` and some very nice vector swizzle optimizations to generate code for AVX2, SSE3, armv8, and others all with one piece of C
https://gcc.godbolt.org/z/4qvKeT

(Also vpermq pshufb pblendd is super fast for AVX2, I've been meaning to add it to PCSX2 for a while)

Zer0xFF:
> Side note: If you're willing to work only with Clang, Clang has a `__builtin_shufflevector` and some very nice vector swizzle optimizations to generate code for AVX2, SSE3, armv8, and others all with one piece of C
> https://gcc.godbolt.org/z/4qvKeT
> 
> (Also vpermq pshufb pblendd is super fast for AVX2, I've been meaning to add it to PCSX2 for a while)

this is really interesting.

though using this would generate the lowest common denominator code, but using it with godbolt could be interesting,
aka generate code for different configuration then add them in, though ofc supprting anything higher than SSE would mean either dropping support for older CPUs, or using the CodeGen to generate this function at runtime

tellowkrinkle:
There are some other things you can do, though they would probably end up being much bigger future projects

- PCSX2 multiply-compiles GSdx and loads the appropriate dylib at runtime.  We're trying to merge that dylib into the main exe now, and [I'm trying out multiply-compiling a subset under a preprocessor-added namespace and selecting an implementation at runtime](https://github.com/PCSX2/pcsx2/pull/4193)
- MacOS lets you add an `x86_64h` slice to a fat binary, to pretty much ship a \<haswell and \>=haswell version as one file, something [Citra does](https://github.com/citra-emu/citra/pull/2982)

bigianb:
> Side note: If you're willing to work only with Clang, Clang has a `__builtin_shufflevector` and some very nice vector swizzle optimizations to generate code for AVX2, SSE3, armv8, and others all with one piece of C
> https://gcc.godbolt.org/z/4qvKeT
> 
> (Also vpermq pshufb pblendd is super fast for AVX2, I've been meaning to add it to PCSX2 for a while)

That would certainly make the code easier to grok ... the only thing is the clang requirement. I'm less concerned about the speed - just moving to any SIMD gives an order of magnitude speedup due to the wider memory fetch/store and this takes it off the critical path. A further order of magnitude speed up would likely have little end result - we'll get better results I think converting more of the memory intensive paths to SIMD.

Source/gs/GSH_OpenGL/GSH_OpenGL_Texture.cpp
```diff
@@ -320,6 +320,162 @@ void CGSH_OpenGL::TexUpdater_Psm16(uint32 bufPtr, uint32 bufWidth, unsigned int
...
+#define USE_SSE
+#elif defined(__APPLE__)
+#include <TargetConditionals.h>
+#if TARGET_CPU_X86_64
+#define USE_SSE
+#elif TARGET_CPU_ARM64
+#define USE_NEON
+#endif
+#elif defined(__ANDROID__) || defined(__linux__) || defined(__FreeBSD__)
+#if defined(__x86_64__)
```
jpd002, line 333: We should add `defined(__i386__)` here for x86-32 support.

Source/gs/GSH_OpenGL/GSH_OpenGL_Texture.cpp
```diff
@@ -320,6 +320,162 @@ void CGSH_OpenGL::TexUpdater_Psm16(uint32 bufPtr, uint32 bufWidth, unsigned int
...
+#define USE_SSE
+#elif defined(__aarch64__)
```
jpd002, line 335: We should add defined(__arm__) for ARM support.

bigianb:
Just tested on my Mac mini m1 ... using the BGDA bartender scene we go from 40fps -> 58fps. I'm surprised the m1 is so slow to be honest - obviously more scope to improve there.

jpd002:
On my end, I tested Phantasy Star 1on my nVIDIA Shield, and I went from 40fps to 43 fps.
Also tested Xenosaga 2's title screen on Windows, getting a nice 3fps boost there too.

I'm going to merge this as this is pretty safe and probably helps a bit in many scenarios. 

I will definitively be interested to see how we could generalize this and maybe see if we could do similar routines for the GS transfer side of things. I think the biggest gains would be there, but we have to keep track the dirty flag which might make things super annoying.

Thanks a lot! 😄

bigianb:
xenosaga 2 uses PSMT4 which is not optimised. Making a SIMD version for that case would give it a similar boost.

image

PR #1024, opened 2021-02-21
Improve README.md Windows

wagrenier:
I added some information on how to compile with VS2019 and add the Play! debugger to the binaries. Included the fixes that I did for some problems I had like the plugins folder not being found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,12 +26,25 @@ To build for Windows you will need to have CMake installed on your system.
  ```
  ```
  # Not specifying -G will automatically generate 32-bit projects.
+ # This will generate a 64bits build with VS2017 compiler
  cmake .. -G "Visual Studio 15 2017 Win64" -DCMAKE_PREFIX_PATH="C:\Qt\5.10.1\msvc2017_64" -DUSE_QT=YES
+
+ # This will generate a 64bits build with VS2019 compiler
+  cmake .. -G "Visual Studio 16 2019" -DCMAKE_PREFIX_PATH="C:\Qt\5.15.2\msvc2019_64" -DUSE_QT=YES
  ```
 You can now build the project by opening the generated Visual Studio Solution or continue through cmdline:
  ```cmd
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,12 +26,25 @@ To build for Windows you will need to have CMake installed on your system.
  ```
  ```
  # Not specifying -G will automatically generate 32-bit projects.
+ # This will generate a 64bits build with VS2017 compiler
  cmake .. -G "Visual Studio 15 2017 Win64" -DCMAKE_PREFIX_PATH="C:\Qt\5.10.1\msvc2017_64" -DUSE_QT=YES
+
+ # This will generate a 64bits build with VS2019 compiler
+  cmake .. -G "Visual Studio 16 2019" -DCMAKE_PREFIX_PATH="C:\Qt\5.15.2\msvc2019_64" -DUSE_QT=YES
```
Zer0xFF, line 33: starting from VS 16 2019, you have to specify arch using `-A x64` or `-A Win32`, and i do believe it defualts to 32 bits if you dont specify

README.md
```diff
@@ -26,12 +26,25 @@ To build for Windows you will need to have CMake installed on your system.
...
  ```
 You can now build the project by opening the generated Visual Studio Solution or continue through cmdline:
  ```cmd
  cmake --build . --config Release
  ```
+
+ To add the Play! debugger to the binaries, simply add the following command to cmake
+ ```cmd
+   # To add the debugger to the build
```
Zer0xFF, line 42: you should probably be more specific here, this should be used in build generation setup, not during the building step

README.md
wagrenier: On my side it seems to be compiling in `x64`, perhaps because I specified the `msvc2019_64` folder.

README.md
wagrenier: I'll change the section it is in. I'll make a `Debugger` section and put it there.

PR #1027, opened 2021-02-23
[PoC] Docker CrossCompiling Armv7/64 action

Zer0xFF:
note this will need this PR https://github.com/jpd002/Play-Dependencies/pull/28 to link/build successfully

edit: ~~todo in comment + remove s4 commit into own branch/PR~~done
I'd need to squash commit and enable other builds as well (once we get build working)


Source/ui_qt/CMakeLists.txt
```diff
@@ -232,6 +240,21 @@ set(QT_COMPONENTS Widgets Core)
 if(TARGET_PLATFORM_UNIX)
 	list(APPEND QT_COMPONENTS X11Extras)
 endif()
+
+set(QT_MOC_EXECUTABLE /usr/bin/moc)
```
Zer0xFF, line 244: Note: only needed during when cross compiling, to avoid using target binary
TODO: add new flag `if(IS_CROSS_COMPILING)`

Source/ui_qt/CMakeLists.txt
```diff
@@ -305,7 +328,7 @@ else()
 	add_executable(Play ${QT_SOURCES} ${QT_MOC_SRCS} ${QT_RES_SOURCES} ${QT_UI_HEADERS})
 endif()
 
-target_link_libraries(Play ${PROJECT_LIBS})
+target_link_libraries(Play ${PROJECT_LIBS} stdc++fs)
```
Zer0xFF, line 331: Todo: `stdc++fs` not needed with gcc 9 or above? need to confirm check

.github/workflows/build-linux.yaml
```diff
@@ -3,38 +3,8 @@ name: Build Linux
 on: [push, pull_request]
 jobs:
   build_linux:
```
Zer0xFF, line 5: todo: change name to make it easier to track? + add to new yaml file.

.github/workflows/build-linux-armv7.yaml
```diff
@@ -0,0 +1,36 @@
+name: Build Linux
+
+on: [push, pull_request]
+jobs:
+  build_linux:
```
Zer0xFF, line 5: todo: change name to make it easier to track?

build_docker/x86_64-cross_compile/arm64/toolchain.cmake
```diff
@@ -0,0 +1,14 @@
+set(CMAKE_SYSTEM_NAME Linux)
+set(CMAKE_SYSTEM_VERSION 1)
+set(CMAKE_SYSTEM_PROCESSOR aarch64)
+
+set(CMAKE_C_COMPILER "aarch64-none-linux-gnu-gcc")
+set(CMAKE_CXX_COMPILER "aarch64-none-linux-gnu-g++")
+set(CMAKE_CXX_LINK_FLAGS "-Wl,-rpath-link,/usr/aarch64-none-linux-gnu/libc/lib64/  -Wl,-rpath-link,/lib/aarch64-linux-gnu/ ${CMAKE_CXX_LINK_FLAGS}")
```
Zer0xFF, line 7: seems compiler/link/cmake? adds `-Wl,-rpath-link, /usr/lib/aarch64-linux-gnu/ ` by default, though not sure where does it get this from, so we should add it for consistency

Zer0xFF:
just a quick note before closing this, this was mainly done as a PoC, though it can potentially be a viable faster (albeit cumbersome) ARM CI build.

though ideally, if we wanted a faster ARM CI build, we'd use self-hosted runners (for better integration with github action), as we wait for official github support (note that github runs on azure, so should azure ever get ARM servers, we'd also see ARM runners).

jpd002:
Sorry for not replying before, but thanks for looking into this. 

I don't have a super good experience with cross compiling setups, so, I'm trying to avoid them, but this is still an interesting alternative if we need something that runs faster. But for now, I don't really mind the 1 hour build time as long as we can know that everything builds and that tests run fine on the target architectures.

Zer0xFF:
no worries, this was always a PoC, as its a trade off between speed vs maintance.

PR #1042, opened 2021-03-12
fix a bad assertion and a crash with 16 bit textures in the debugger

bigianb:


Source/ui_qt/DebugSupport/FrameDebugger/GsPacketData.cpp
```diff
@@ -20,7 +20,6 @@ void GsPacketData::appendChild(GsPacketData* item)
 
 GsPacketData* GsPacketData::child(int row)
 {
-	assert(row < 0 || row >= m_children.size());
```
bigianb, line 23: This original assertion was obviously backwards ... but more to the point the Qt code asserts for the out of bounds condition, so there's not point in this at all.

jpd002:
Nice catches! Thanks!

PR #1044, opened 2021-03-20
Fix android signed artifact upload

Zer0xFF:
note, i dont have keystore set on my repo, so its probably best to push this to a branch to make sure its working correctly, though it is a small fix and should work just fine.


jpd002:
Pretty confident it'll work fine :) Thanks!

Zer0xFF:
Seems to be working correctly :)

PR #1046, opened 2021-03-28
(Libretro) Add .gitlab-ci.yml

twinaphex:
Hi there,

as per my discussion with @Zer0xFF , here is a PR for the Gitlab CI file that we'll be needing in order to build cores from this repository. Thanks in advance.


jpd002:
Thanks!

PR #1047, opened 2021-04-14
(Android) add values-ru/strings

6lackmag3:
Russian translation


jpd002:
Awesome, thanks for this!

6lackmag3:
Je vous en prie! Thank you for making Play!

PR #1049, opened 2021-04-26
Partially implement TEST_1/2 AFAIL on mobile via framebuffer fetch

kernle32dll:
When a alpha test fail method is set, and the trivial cases cannot be handled via color mask, we need to handle it in shader code.

TODO: This implementation based on framebuffer fetch, and thus only works on mobile. Also, this does not correctly handle cases where no depth should be written to the framebuffer.


jpd002:
Thanks for this! 😃

PR #1052, opened 2021-04-28
Make AFAIL a bit more lenient on desktop

kernle32dll:
Currently, on master, we have two scenarios:

Firstly, if we can use the framebuffer fetch extension, alpha test fail is implemented mostly correctly (cases which should prevent writing depth are not handled correctly). If we don't have access to the extension, we discard the write on test failure completely.

This PR tackles the second case. Specifically `ALPHA_TEST_FAIL_FBONLY` (only write color and alpha, but no depth). This change allows the pixel to be written with color, alpha and depth in full. While this is not correct (it should not write depth), this is better than discarding the pixel write completely, and actually fixes quite some games (I will add screenshots later). `ALPHA_TEST_FAIL_RGBONLY` is basically the same thing (writing all three components), when it should only write color.

This also contains a minor shader cleanup, where shader code for conditional color and alpha writing (via framebuffer fetch extension) is only added to the shader if the extension is available.

### DIFF
--- a/Source/gs/GSH_OpenGL/GSH_OpenGL_Shader.cpp
+++ b/Source/gs/GSH_OpenGL/GSH_OpenGL_Shader.cpp
@@ -346,8 +346,11 @@ Framework::OpenGl::CShader CGSH_OpenGL::GenerateFragmentShader(const SHADERCAPS&
 		shaderBuilder << "	textureColor = v_color;" << std::endl;
 	}
 
-	shaderBuilder << "	bool outputColor = true;" << std::endl;
-	shaderBuilder << "	bool outputAlpha = true;" << std::endl;
+	if(useFramebufferFetch)
+	{
+		shaderBuilder << "	bool outputColor = true;" << std::endl;
+		shaderBuilder << "	bool outputAlpha = true;" << std::endl;
+	}
 
 	if(caps.hasAlphaTest)
 	{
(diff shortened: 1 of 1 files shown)


kernle32dll:
# SLUS-20994: Full Metal Alchemist

Previously did not render anything - now renders titlescreen, introduction and in-game

Screenshot from 2021-04-28 19-43-20
Screenshot from 2021-04-28 19-43-43
Screenshot from 2021-04-28 19-44-40
Screenshot from 2021-04-28 19-47-08
Screenshot from 2021-04-28 19-48-21

# SLES-55133: Lego Indiana Jones

Now renders title screen text, and the loading screen map

Screenshot from 2021-04-28 19-29-49 Screenshot from 2021-04-28 20-01-39

Screenshot from 2021-04-28 19-29-35 Screenshot from 2021-04-28 20-01-50

# SLUS-20387: Suikoden 3

Renders the title screen text (copyright text is a bit crisper, too)

Screenshot from 2021-04-28 19-26-41 Screenshot from 2021-04-28 19-42-20

# SLES-54087: Suikoden 5

UI highlights (orange) render correctly, and the alpha transition in the message box is correctly fading instead of being cut-off

Screenshot from 2021-04-28 19-32-02 Screenshot from 2021-04-28 19-38-57

Screenshot from 2021-04-28 19-33-23 Screenshot from 2021-04-28 19-39-50

Screenshot from 2021-04-28 19-34-02
Screenshot from 2021-04-28 19-40-05

Screenshot from 2021-04-28 19-35-28 Screenshot from 2021-04-28 19-37-46

# SLUS-21782: Persona 4

Title screen rendering correctly

Screenshot from 2021-04-28 19-52-27
Screenshot from 2021-04-28 19-57-08

jpd002:
As discussed on Discord, keeping the color even if depth/alpha should be discarded is a better policy that discarding everything. It does make things a bit wonky in some games (Grandia X has some problems with this), but it's not a deal breaker (and it fixes other problems in this game too). It makes the OpenGL renderer better overall.

Thanks for this!

PR #1053, opened 2021-04-29
MacOS: update rpath for ARM64 builds

Zer0xFF:
Quick note:
* I dont know if this is the "correct" fix for a os install, as comparing the generated libraries (ARM) to installed (x86) ones, we see that we have different paths, aka these use `@rpath/blah`, while installed ones have absolute path (though this shouldnt matter as we're not preparing these for install, but packaging)

* the resulting output from otool is almost identical with 1 noticable difference (when inspected with `otool -l`)
  * framework libs : `LC_RPATH` `@loader_path/../../..`
  * plugins libs:  `LC_RPATH` `@loader_path/../../lib`
  * these paths are ordered after `LC_CODE_SIGNATURE` which might cause issue with signature verifcation?
  * I also can't make sense of these paths, so not sure if they're even needed?

I dont have an M1 to test with, so 🤞

related: #988

### DIFF
--- a/.github/CI_MISC/pre_qt_universal_build.py
+++ b/.github/CI_MISC/pre_qt_universal_build.py
@@ -3,6 +3,40 @@ import os
 import shutil
 import glob
 
+def fix_id(lib_name, is_fw):
+    if(lib_name.count("HOMEBREW_PREFIX") == 0):
+        return ""
+
+    if(is_fw):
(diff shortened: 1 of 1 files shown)


jpd002:
Pushed on my repo to test with the full signature/notarization process. Going to test once it's done building.

jpd002:
Working perfectly on my Apple Silicon Mac! Thanks!

PR #1054, opened 2021-05-08
Make CMcServ::GetInfo return -1 on first invocation

kernle32dll:
If GetInfo is called for a previously not encountered memory card, return -1 (formatted memory card). All subsequent calls (to the same card) return 0, indicating that no new card was inserted.

Onimusha 2 depends on this for the initial memory card detection.


kernle32dll:
Note, that the previous implementation of this (https://github.com/kernle32dll/Play-/commit/4f58807dee2f9ab8b73cef1a1a8a5d02aeda8e17) made (at least) Silent Hill 3 regress. That game checks for both ports at a time, before actually trying to read anything. Because of this, the old implementation would effectively always return -1, thus making the game loop indefinitely.

kernle32dll:
Also fixes saving in Samurai Western

Screenshot from 2021-05-08 19-43-54

Source/iop/Iop_McServ.h
```diff
@@ -198,6 +198,11 @@ namespace Iop
 		static const char* m_mcPathPreference[2];
 		std::string m_currentDirectory;
 		CPathFinder m_pathFinder;
+
+		// Keeps track, if the memory card in
+		// a given slot has already been read,
+		// or if it is a newly inserted card.
+		std::vector<bool> m_knownMemoryCards;
```
jpd002, line 205: This doesn't need to be dynamic. All the code assumes there's two ports. It would probably be a good occasion to add a `MAX_PORTS` enum inside the class and use it instead of the hardcoded `2`.

jpd002:
Nice! The change make sense!

As mentioned, the array doesn't need to be dynamic. Also, I think it would be best to save the state of this array when saving the emulator's state. Just to avoid some strange things when we are loading a state from a freshly started emulator.

Source/iop/Iop_McServ.h
kernle32dll: Force pushed the change, and added the enum in a separate commit

kernle32dll:
Like mentioned in the PR comment, made the array non-dynamic, and added state persisting:

```xml
<Memorycards>
	<Memorycard Known="true" Port="0" />
	<Memorycard Known="false" Port="1" />
</Memorycards>
```

Source/iop/Iop_McServ.h
```diff
@@ -198,6 +202,11 @@ namespace Iop
 		static const char* m_mcPathPreference[2];
 		std::string m_currentDirectory;
 		CPathFinder m_pathFinder;
+
+		// Keeps track, if the memory card in
+		// a given slot has already been read,
+		// or if it is a newly inserted card.
+		bool m_knownMemoryCards[2];
```
jpd002, line 209: Could we use the newly added `MAX_PORTS` enum here?

kernle32dll: Done

PR #1057, opened 2021-05-10
Clean up LibMC2 warnings

kernle32dll:
Remove the assert for unmatched function calls, and instead log all unmatched functions as warnings


jpd002:
Thanks! Much better than way!

PR #1060, opened 2021-05-11
Add iSifSendCmdIntr

kernle32dll:
Makes Test Drive Unlimited boot (takes very long to load, and gets stuck at logo screen - so just a small fix)


jpd002:
Thanks!

PR #1062, opened 2021-05-15
Fixes a bug in Fatal Frame 1 that makes the character stuck in the doorway

darkpigeon:
At certain points during the game, the character can get "stuck" after opening a door. The character can still turn at any directions when being stuck, but not able to move any further.
For example, this always occurs when going into the dressing room in the hallway after getting past the door behind folding screens.
Or you can check the video below to have a better understanding of this bug.

https://user-images.githubusercontent.com/12695961/118354483-76a1c580-b595-11eb-9c80-7017b1507b5e.mp4

This line I added will fix that issue completely.


jpd002:
Nice find!

I think it's better to use `ClampVector` to fix incoming registers instead of fixing the result and even better, not to use it at all if possible. 

Do you have the instruction sequence that led you to find that usage of `ClampVector`? I'd like to see if it's really the last resort.

darkpigeon:
Pcsx2 has the same issue with this game, which can only be solved by turning on VU clamping since they don't have any other fix for it yet. I took that in mind and spent like a whole day trying to find the exact function that broke the game in Play! through trials and errors and put ClampVector into it.
I agree that this is not a proper way to fix the issue and I'll be pretty glad if there is a better solution.

jpd002:
Oh, cool! Alright, well, I'll check the game in the debugger to see if I can find an alternative.

jpd002:
Sorry, I didn't check your patch correctly: the `ClampVector` is appropriately placed already.

I ran it through the debugger quickly to see if anything stood out. It seems to be using garbage data, don't have much choice but to clamp the vector.

Thanks for this!

PR #1064, opened 2021-05-15
Rule of rose fixes

kernle32dll:
Includes two things:

- Missing OpenGL Blendmode 1021
- Preliminary implementation of SetFIleInfo in McServ

The game uses the latter to rename savegames during saving. I presume the game uses the filename to encode some meta data, but thats just speculation.


Source/iop/Iop_McServ.cpp
```diff
@@ -659,6 +663,33 @@ void CMcServ::GetDir(uint32* args, uint32 argsSize, uint32* ret, uint32 retSize,
 	ret[0] = result;
 }
 
+void CMcServ::SetFileInfo(uint32* args, uint32 argsSize, uint32* ret, uint32 retSize, uint8* ram)
+{
+	auto cmd = reinterpret_cast<const CMD*>(args);
+	CLog::GetInstance().Print(LOG_NAME, "SetFileInfo(port = %d, slot = %d, name = '%s');\r\n", cmd->port, cmd->slot, cmd->name);
```
jpd002, line 669: There's a `flags` parameter to that function (as shown here: https://github.com/ps2dev/ps2sdk/blob/11500fe1dcec03592e71406cbd9d34b39aa52589/iop/memorycard/mcserv/src/mcserv.c#L1217). Could you also get that value and print it out?

Also, I think it's used to specify what kind of change can be made to the entry, I think it would be interesting to add a warning if it's not something that we expect. Like, here, I think the the only flag we handle in that implementation would be `sceMcFileAttrFile` (ex.: https://github.com/ps2dev/ps2sdk/blob/11500fe1dcec03592e71406cbd9d34b39aa52589/iop/memorycard/mcman/src/ps2mc_fio.c#L485).

Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
```diff
@@ -770,6 +770,13 @@ void CGSH_OpenGL::SetupBlendingFunction(uint64 alphaReg)
 		glBlendColor(0, 0, 0, static_cast<float>(alpha.nFix) / 128.0f);
 		glBlendFuncSeparate(GL_ONE_MINUS_CONSTANT_ALPHA, GL_CONSTANT_ALPHA, GL_ONE, GL_ZERO);
 	}
+	else if((alpha.nA == 1) && (alpha.nB == 0) && (alpha.nC == 2) && (alpha.nD == 1))
```
jpd002, line 773: Please use `ALPHABLEND_ABD_*` and `ALPHABLEND_C_*` constants instead of numbers.

kernle32dll: Done

Source/iop/Iop_McServ.cpp
kernle32dll: Done. I hope the enum is okay this way? Maybe we can fold it together with `OPEN_FLAGS`, but I would suppose that should be done in a separate step afterwards.

literalmente-game:
This also fixed the save file creation for Burnout https://github.com/jpd002/Play-Compatibility/issues/49

Source/iop/Iop_McServ.cpp
jpd002: Yup, all good!

jpd002:
Thanks again! Loving all these fixes!

PR #1067, opened 2021-05-19
Patch codes which help Onimusha - Dawn of Dreams goes ingame

darkpigeon:
As stated in the title, these codes will help Onimusha - Dawn of Dreams gets past the memory card checking procedure and goes ingame properly. Works with both NTSC and PAL versions.
Credit goes to ShadowLady for providing the original codes which were used in pcsx2. I just ported them over.


jpd002:
This is not necessary anymore. Fixed by https://github.com/jpd002/Play-/commit/cc614c662649fc0c07346296c535258d04c6ffbd.

PR #1069, opened 2021-05-25
Sysclib memcpy address sanitation

kernle32dll:
Makes sysclib's `memcpy` use `GetPtr` for address sanitation, and implements IOP ram mirroring in said function.

Makes Max Payne 2 not segfault before going into menu on desktop, and makes menu music in that game work in general.


jpd002:
Nice work! Thanks! 😄

PR #1071, opened 2021-05-31
[Android] Migrate to AndroidX Preference

TacoTheDank:
### 1st commit
- Preliminary changes before the main migration:
- - Move all preference strings to Constants
- - Rename xml preferences stuff
- - Move preference fragments into their own classes
- Some miscellaneous stuff in the gradle:
- - Remove unnecessary maven.google in the repositories block
- - Add support for Java 8

### 2nd commit
- Migrate to AndroidX Preferences
- - Get rid of severely deprecated PreferenceActivity
- - Don't reserve icon space for settings without icons
- Other stuff:
- - Use the same SharedPreferences instance for registering and unregistering

### 3rd commit
- Change all instances of CheckBoxPreference to SwitchPreferenceCompat (see [here](https://developer.android.com/guide/topics/ui/settings/components-and-attributes#individual_preferences) why; scroll to the note)

### 4th commit
- Remove unused layout
- Refactor ThemeManager.applyTheme. Having all activities require a Toolbar that needs to specifically be named ["my_awesome_toolbar"](https://github.com/jpd002/Play-/commit/71056a7b814b6b7fbc655287beb6b3e63ce4928c#diff-ec3cf29914629bba625af5044f0c7658667f1f5871cb548a0b11f00d2f5d30baL46) to work correctly is sketch as fuck and asking for trouble, never mind the fact that EmulatorActivity doesn't even have one. Therefore:
- - Add a nullable parameter for Toolbar that each activity can pass its toolbar through (and EmulatorActivity; its lack thereof).
- - Basically it works the same way but now it doesn't hang in the balance of a specifically named attribute.

I built and tested an APK, and everything works fine.


jpd002:
Very nice work! Thank you very much!

TacoTheDank:
Welcome! :)

PR #1073, opened 2021-06-01
Fixes for Darkwatch

kernle32dll:
Includes two fixes:

- Adjustment for `CIopBios::TryGetImageVersionFromPath` to be case insensitive, as the game calls it with `rom0:UDNL host0:ioprp300.img` - note the lowercase
- ~Basic re-mapping in `Ioman::PreOpen` to try to find files in cdrom0, if they cannot be found in host0. The game expects to find its data files in host0.~
- Upon loading the executable, A0 is set to the entrypoint address. The game relies on this (undefined behavior).

### DIFF
--- a/Source/ee/PS2OS.cpp
+++ b/Source/ee/PS2OS.cpp
@@ -492,6 +492,7 @@ void CPS2OS::LoadExecutableInternal()
 	}
 
 	m_ee.m_State.nPC = header.nEntryPoint;
+	m_ee.m_State.nGPR[CMIPS::A0].nV[0] = header.nEntryPoint;
 
 #ifdef DEBUGGER_INCLUDED
 	std::pair<uint32, uint32> executableRange = GetExecutableRange();


jpd002:
I have doubts on the remapping thing. The game must be doing something to make this `host0` -> `cdrom0` redirection happen. 

Does it uses `AddDrv` anywhere to register a device? 

Also, lots of games check if its running on tool hardware and selects `host` or `cdrom` depending on the result. Could the game be doing something of the like?

kernle32dll:
As discussed, I removed the host0 re-mapping, and added the fix for the A0 register.

jpd002:
I think commit https://github.com/jpd002/Play-/pull/1073/commits/16106ea94e75a7163dbdaa4f2c174dc45ed65db0 can be removed too. The game now sends `rom0:UDNL cdrom0:\IOPRP300.IMG;1` when resetting the IOP, which works fine with the current code.

jpd002:
Thanks!

PR #1075, opened 2021-06-03
Persist framebuffer nacking CLUT data to GS memory 

kernle32dll:
Dragonball Z Tenkaichi 2 renders its CLUT information into a framebuffer, and then tries to use it from there. This works for Vulkan, but for OpenGL the framebuffer content is never propagated back into the GS RAM, and thus unavailable when the CLUT information is required.

This PR uses a light-weight approach of reading framebuffer contents when required for CLUT calculation, and persists it (if not already done so). This works pretty well, and only decreases performance marginally.

Special note on the "downSample" part. For whatever reason, on resolution upscaling > 1 I got artifacts in the CLUT rendering. I don't know why, but down-sampling it in GPU memory again helped clear the issues.


kernle32dll:
Master:
Screenshot from 2021-06-04 01-43-42

PR:
Screenshot from 2021-06-04 01-36-41

Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
```diff
@@ -1833,6 +1890,27 @@ void CGSH_OpenGL::CopyToFb(
 
 void CGSH_OpenGL::WriteRegisterImpl(uint8 nRegister, uint64 nData)
 {
+	switch(nRegister)
+	{
+	case GS_REG_TEX0_1:
+	case GS_REG_TEX0_2:
+	{
+		// If we use a CLUT based texture, its possible that the memory
```
jpd002, line 1898: Would it work to put that logic inside a new SyncCLUT function inside the `CGSH_OpenGL` (that overrides the base function from `CGSHandler`)?

kernle32dll: Done

Source/gs/GSH_OpenGL/GSH_OpenGL.cpp
```diff
@@ -1833,6 +1890,27 @@ void CGSH_OpenGL::CopyToFb(
...
+		// referenced is inside a framebuffer. In that case, we need
+		// to persist the framebuffer content into the GS RAM.
+
+		auto tex0 = make_convertible<TEX0>(nData);
+		const uint32 ptr = tex0.GetCLUTPtr();
+
+		if(ptr == 0 || !ProcessCLD(tex0)) break;
```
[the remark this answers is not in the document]
kernle32dll: Done

Bogdans29:
Is Vulkan supported on android?

XiaomiRedmi9:
Any updates about this?

PR #1082, opened 2021-06-08
Delay all McServ commands

kernle32dll:
Fixes Operation WinBack 2 memory card access

### DIFF
--- a/Source/iop/Iop_McServ.cpp
+++ b/Source/iop/Iop_McServ.cpp
@@ -33,7 +33,7 @@ using namespace Iop;
 
 #define SEPARATOR_CHAR '/'
 
-#define CMD_DELAY_GETINFO 100000
+#define CMD_DELAY_DEFAULT 100000
 
 #define STATE_MEMCARDS_FILE ("iop_mcserv/memcards.xml")
 #define STATE_MEMCARDS_NODE "Memorycards"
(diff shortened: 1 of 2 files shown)


Source/iop/Iop_McServ.cpp
```diff
@@ -91,7 +92,6 @@ void CMcServ::CountTicks(uint32 ticks, CSifMan* sifMan)
 	moduleData->pendingCommandDelay -= std::min<uint32>(moduleData->pendingCommandDelay, ticks);
 	if(moduleData->pendingCommandDelay == 0)
 	{
-		assert(moduleData->pendingCommand == CMD_ID_GETINFO);
```
jpd002, line 94: We should check that we have a pending command here (ie.: `assert(moduleData->pendingCommand != CMD_ID_NONE`).

jpd002:
Do you think it would make sense to consolidate all delay code in the same place? Basically, get rid of the code that's at the end of `GetInfo` since your change takes care of delaying everything?

If we do that, I think we should also bring in the comments that say which games are sensitive to delays (and also add a note about Operation WinBack 2 in there).

Source/iop/Iop_McServ.cpp
kernle32dll: This could never occur, as there is a `if(moduleData->pendingCommand == CMD_ID_NONE) return;` check above.

jpd002:
Tested plenty of games, seems to be working good. Thanks!

PR #1086, opened 2021-06-10
[Android] Clean up AlertDialogs

TacoTheDank:
I cleaned up the AlertDialog code because it was wrought with redundancy :)


jpd002:
Very nice! Thanks! :)

PR #1087, opened 2021-06-11
[Android] Preference and theming improvements

TacoTheDank:
### 1st commit
- Fix preference that somehow got itself out of order during my AndroidX Preference PR

### 2nd commit
- Use the (relatively) new simple summary provider feature (introduced in [this](https://developer.android.com/jetpack/androidx/releases/preference#version_110_3) version)
- Rename old checkBox names to switch

### 3rd commit
- Make all themes use one common parent
- Create a colorAccent, which will theme stuff like dialogs and settings. This gets rid of the outdated holo blue that's still visible in several locations.
  - For the above to apply correctly to settings, the settings activity needs to be recreated with the new theme whenever it is changed. The `setDefaultValues` is for this reason:
    - The settings' default values are first set when its XML layout is inflated for the first time. So when the UI settings layout is inflated for the first time, it causes `onSharedPreferenceChanged` to trigger (its interface method notes say "Called when a shared preference is changed, added, or removed. This may be called **even if a preference is set to its existing value**"). Consequently, the screen will flash black for half a second because the activity will be recreated. However, this could likely be confusing to an end user. So what `setDefaultValues` does is initialize the preferences to be set to their default value on the setting activity's first launch (THIS DOES NOT RESET PREFERENCES; this only officially sets their values in the layout if they haven't already been set). So now, because the theme preference value has already been initialized, the screen will not flash when the user goes into the UI settings for the first time.
    - NOTE: If this isn't done, the screen will not flash on future inflations of UI settings. It is only on the first inflation. However, with `setDefaultValues` as explained above, the activity will still be recreated when the theme is changed. We use `setDefaultValues` simply so users will not confused by the flashing when they haven't actually changed the theme yet.
    - We do all this in the first place because if the activity is not recreated, certain stuff within settings (like switches) will not have been updated to reflect the new chosen theme.

Sorry that last bit was so long, hopefully it made sense lol.

### DIFF
--- a/Source/ui_android/java/com/virtualapplications/play/Constants.java
+++ b/Source/ui_android/java/com/virtualapplications/play/Constants.java
@@ -10,9 +10,6 @@ public class Constants
 	 */
 	public static final String PREF_EMU_GENERAL_SHOWFPS = "ui.showfps";
 	public static final String PREF_EMU_GENERAL_SHOWVIRTUALPAD = "ui.showvirtualpad";
-	public static final String PREF_EMU_VIDEO_RESFACTOR = "renderer.opengl.resfactor";
-	public static final String PREF_EMU_VIDEO_PRESENTATIONMODE = "renderer.presentationmode";
-	public static final String PREF_EMU_AUDIO_BUFFERSIZE = "audio.spublockcount";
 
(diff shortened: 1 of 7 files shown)


jpd002:
Very nice! Thanks again!

PR #1088, opened 2021-06-14
add missing include

alucryd:
Hi, been getting the following error with the past few commits: https://buildbot.alucryd.xyz/#/builders/287/builds/155/steps/8/logs/stdio

The compiler was kind enough to print the fix.


jpd002:
Thanks!

Issue #1093, opened 2021-06-25
[Feature Request iOS] ability to use files system


Spidy123222:
On iOS there is a files app that is made from apple for developers to use for their apps and this can be used to have settings files and games be shown in to be able to manage play! On the iOS device it is on.

escape1:
Tested the latest build and the Play! app doesn’t seem to appear in the files app.

yoshisuga:
This was done as part of the PR that I worked on. If you update and install the app, you should see the files visible in the Files app.

PR #1097, opened 2021-07-03
[iOS] Support saving/loading to multiple save states

yoshisuga:
Add a view controller (written in Swift) that lets you save/load multiple save states.

Tapping outside of the controls will cancel and hide the save state controller view.

IMG_5289F0050030-1

Also supports viewing the files in the Files app (addresses #1093)

I'm not too versed in how Cmake will generate the project as I hope it will just bring in the new Swift file and bridging header.

I've had to edit the Build Settings/Project Workspace post-cmake to:
- specify a writable libMoltenVK.dylib in the project
- add a Run Script in Build Phases in the Play target to code sign libMoltenVK.dylib


Gamemulatorer:
Nice! this is also possible on android?

Source/ui_ios/SaveStateController.swift
```diff
@@ -0,0 +1,117 @@
+//
```
jpd002, line 1: Please remove copyright header.

Source/ui_ios/EmulatorViewController.mm
```diff
@@ -63,6 +65,18 @@ - (void)viewDidLoad
 	[self.view addSubview:self.iCadeReader];
 	self.iCadeReader.delegate = self;
 	self.iCadeReader.active = YES;
+    
+    self.saveStateController = [[SaveStateController alloc] init];
```
jpd002, line 69: Is there a way to create this through the interface builder? I'm trying to avoid adding UI elements programatically, I find it easier to manipulate them in the editor.

jpd002 (changes requested):

jpd002:
Very nice addition! Thank you for this! Left some comments/questions for review.

Still waiting for the build to be done, but I think we'll need to bump the target iOS version to prevent having to bundle all the Swift libraries. Had that problem when I was trying to integrate AltKit. I'll merge that commit if that's the case: https://github.com/jpd002/Play-Dependencies/commit/201207241a535395438ae033cf6ecdfc1321f624.

Source/ui_ios/EmulatorViewController.mm
yoshisuga: Yes it can be done through a storyboard using interface builder as well. I prefer creating the UI programmatically because:
- All the UI definition is in one place, no need to make connections with IBOutlets so less room for confusion
- Interface Builder can get really slow
- Manipulating autolayout constraints in Interface Builder is kind of a pain

I have no problem doing this with a storyboard though, if that's what you prefer.

yoshisuga:
I've been looking into the CMake Xcode project generation to work with this. I've been able to get my swift file into the Xcode project, but having trouble with enabling Swift properly.

There are two issues:
- It looks like `enable_language(Swift)` needs to be called. Doing this in Source/ui_ios/CMakeLists.txt seems to work, but it adds some unneeded Swift flags (-x objective-c++ -fobjc-arc) that don't work with the Swift compiler.
- The Objective C bridging header needs to be specified in the Build Settings, and this is done by saying this in ios.cmake:
```
set(CMAKE_XCODE_ATTRIBUTE_SWIFT_OBJC_BRIDGING_HEADER "${PROJECT_SOURCE_DIR}/Source/ui_ios/Play-Bridging-Header.h")
```

However, specifying this causes the Swift compiler check done by CMake to fail because it can't find the bridging header file that was defined.

As a whole, I get the impression that the CMake Xcode project generation isn't mature enough to support Swift, so I'm planning on just writing the new view controller in objective-c instead.

yoshisuga:
> Nice! this is also possible on android?

Yes, totally doable on Android since the API supports this. My Android skills are way out of date but I'm sure someone else can do this.

yoshisuga:
Ok, ported the view controller to objective c and added the file entries to the CMakeLists.txt so it should get brought over to the Xcode project without any issues...(hopefully - I did test it though)

jpd002:
Yeah, using Swift with CMake has some pitfalls. 

For compiler flags, you need to use generator expressions to prevent gcc/clang flags from being used in the Swift compiler (example here: https://github.com/jpd002/Play-/commit/0da7732d668662e9e309ae9905a7a8393ddef40d).

The current state of the code is good, please just make sure formatting issues are addressed. The check format action generates a patch that you can apply on your changes: https://github.com/jpd002/Play-/pull/1097/checks?check_run_id=2983886595.

Thanks!

jpd002 (approved):

yoshisuga:
> Yeah, using Swift with CMake has some pitfalls.
> 
> For compiler flags, you need to use generator expressions to prevent gcc/clang flags from being used in the Swift compiler (example here: [0da7732](https://github.com/jpd002/Play-/commit/0da7732d668662e9e309ae9905a7a8393ddef40d)).
> 
> The current state of the code is good, please just make sure formatting issues are addressed. The check format action generates a patch that you can apply on your changes: https://github.com/jpd002/Play-/pull/1097/checks?check_run_id=2983886595.
> 
> Thanks!

Thanks! Applied the formatting changes. Please merge at your convenience if it looks ok. Thanks again!

jpd002:
Thanks a bunch for this!

PR #1100, opened 2021-07-06
Some Android code lint cleanup

TacoTheDank:
Yeah.


jpd002:
Nice. Out of curiosity, which tool/command line did you use to lint the code?

TacoTheDank:
Android Studio has an "Inspect Code" tool that gives you a big list of every lint it can find in a project, and then you can go through and lint things to your heart's content :)

jpd002:
Ah, cool :) Well, would be nice if it could be automated somehow, but for now, this will do!

Thanks!

PR #1104, opened 2021-07-12
Added mass

strikersix23:
Added mass to vfs, unsure if its correct buts its worth a shot


jpd002:
Hey!

You're missing the registration of "mass" in here: https://github.com/jpd002/Play-/blob/master/Source/PS2VM.cpp#L432.

Otherwise, even though it's not a built-in feature in the PS2 (it's implemented by custom modules), I think it works, not really planning on emulating USB storage devices.

strikersix23:
Just need it to see mass in vfs

jpd002:
Thanks! I think something went wrong when you did your change, the `PS2VM.cpp` file is now completely changed. Could you redo your change to make sure that the diff only includes your additions?

strikersix23:
Id just need to rebase

strikersix23:
Rebase is done

strikersix23:
Im having a problem trying to just show the changes

Zer0xFF:
1) rebasing won't fix this, the commit needs to be redone.
2) this is very likely happening because you changed git default line ending handling
3) I remember messing with mass a while back, I don't recall finding any homebrew that works with just these changes (if you have an example that this works with please mention it)

uyjulian:
You may need to implement the rename IOCTL for rename support to work properly, since the homebrew usbhdfsd doesn't use iomanX.

strikersix23:
would implementing this work from this work https://github.com/jackun/USBqemu-wheel/blob/master/src/USB.h

andymage593:
image


image


https://www.youtube.com/watch?v=oVzcl6Lf8Qg

Is it compatible with USB key dongle?

strikersix23:
Sorry no longer working on this poll request as its no longer needed

PR #1106, opened 2021-07-16
Fix compile errors

effective-light:
Modern versions of `gcc` and `clang` require these changes to compile.

see: https://reviews.llvm.org/D86936


jpd002:
Thanks for this!

PR #1124, opened 2021-09-12
Android: Update Chinese - Simplified translation

Croden1999:


jpd002:
Hey there! Thanks for this!

I think it doesn't make sense to have the `strings.xml` file moved from `src/main/res/values-zh/` to `源文件/主要的/资源/值-zh/` (maybe it's a mistake?). Can you move it back to its original location?

Croden1999:
> Hey there! Thanks for this!
> 
> I think it doesn't make sense to have the `strings.xml` file moved from `src/main/res/values-zh/` to `源文件/主要的/资源/值-zh/` (maybe it's a mistake?). Can you move it back to its original location?

Sorry, i got it.

PR #1153, opened 2021-11-29
cmake: Better git handling

orbea:
This fixes some problems.

* `git describe` without `--always` can fail in shallow clones.
* git will end up working in the wrong directory when doing an out of tree build.
* git may not actually be installed
* The `.git` directory may not be present, for example the user can download a source archive from any github branch, tag or commit where the git history will not be included.


jpd002:
Interesting, way better than what I did 😄 Thanks a bunch!

PR #1154, opened 2021-11-29
cmake: Use GNUInstallDirs

orbea:
This is helpful for distros and anyone installing to a non-standard potentially arbitrary location. For example a distro may want `/usr/games/Play` rather than the default of `/usr/bin/Play` and can configure this with `-DCMAKE_INSTALL_BINDIR=game`.

Reference: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

I also reformatted it for better readability.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,8 @@ endif()
 
 project(Play)
 
+include(GNUInstallDirs)
+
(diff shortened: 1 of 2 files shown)


jpd002:
Thanks for this!

PR #1158, opened 2022-01-12
Qt : allow filtering bootable list (though path, title and disc id)

Zer0xFF:
I'm debating using the model proxy to also sort the list,
as that will make the overall code simplier (no need to destory and repopulate `m_bootables` or `m_model`),
but it will mean we will have 2 different sorting codes, one Qt specific and the current (sqlite) one.


jpd002:
Nice! Thanks for this!

Humm, I think I would rather have only one sorting code in the Bootables handling code (with SQLite) and pay for the extra complexity in the Qt UI code. I think it would be a bit confusing to have two sorting passes. Also, if we improve sorting in Bootables, all platforms can benefit from it.

PR #1161, opened 2022-01-14
Bootable layout

Zer0xFF:
Just opening the PR, to show the progress and ask for preferred direction.

1) Should we remap Ctrl+L to show this layout
2) if (1) is yes, this will imply removing the dialog option all together.... then should the dialog options (refresh, add game, scan, title, disc id) then be added to the new layout (thus keeping the same layout format)
2.b) or add those options to a new menu option? (potentially one thats only shown when this layout is shown)
3) background or no background

contain filter changes from #1158 , closes #1158


jpd002:
> Should we remap Ctrl+L to show this layout

Yes. I think we should include the dialog options to this layout.

And the background is nice, we should keep it 😃.

Zer0xFF:
ready to be reviewed, this should now contain changes from #1158 and all the functionality that was in the dialog.

just to note, code was moved to the new `QBootablesView.cpp/h` both as cleanup and better source/header file naming, but also because i was adapting any needed code as I was going alone.

I'm using lambda/binds to boot games, and change to the correct view, to avoid making `QBootablesView` dependent on virtual machine, but also to avoid circular dependecy with `MainWindow` (though this dependency would probably make the code a bit simplier).

Source/VirtualMachine.h
```diff
@@ -1,21 +1,22 @@
...
+		RUNNING = 1,
+		PAUSED = 2,
+	};
+
+	virtual ~CVirtualMachine() = default;
+	virtual STATUS GetStatus() const = 0;
+	virtual void Pause() = 0;
+	virtual void Resume() = 0;
+
+	Framework::CSignal<void()> OnMachineStateChange;
```
jpd002, line 19: Can we give this the same treatment as RunningStateChange (ie. adding a MachineStateChangeEvent type)?

Source/ui_qt/BootableModel.cpp
```diff
@@ -69,9 +69,10 @@ void BootableModel::SetWidth(int width)
 }
 
 /* start of BootImageItemDelegate */
-BootableCoverQVarient::BootableCoverQVarient(std::string key, std::string title)
+BootableCoverQVarient::BootableCoverQVarient(std::string key, std::string title, std::string path)
```
jpd002, line 72: Could we fix the typo in BootableCoverQVarient? I think it should be BootableCoverQVariant.

jpd002:
Looking good! I added some review comments, just superficial stuff.

Did you try it on macOS? I'll give it a spin there just to be sure everything it ok.

Zer0xFF:
Changes made.

Though I noticed few issues:
Just fixed my MacOS VM, emu window is flickering (not an issue with emu from site).
Ubuntu emu window works fine, but covers are not loading correctly :S (i guess thats why i used scroll last time?)

Zer0xFF:
the ubuntu issue has been fixed.

can you confirm if the flickering issue happens on real OSX hardware? (this might be a result of how widgets are handled on OSX, if so, I'd just have to go back to my initial attempt with hiding/showing widgets as needed).

jpd002:
No flickering on my MacBook. So, I think we're good for that.

I noticed another issue though: it seems games are booted twice when double clicking on a game.
QBootablesView::on_listView_doubleClicked is called (which seems to have the wrong game if filtering is enabled) and QBootablesView::DoubleClicked is called after. Dunno if the handlers are registered twice or something like that?

Zer0xFF:
`QBootablesView::DoubleClicked` should have been removed after changing `QBootableListView` into `QBootablesView`, so there was double trigger, but the issue was caused by using proxy index.

jpd002:
Working great! Thanks a lot for this!

Improvement ideas (for other merge requests);
- Starting a game from the File menu doesn't update the bootable list.
- Make "Add Games" and "Refresh" run on a background thread and have some feed back that something is happening (like on iOS and Android).

Zer0xFF:
Sure ill look into those changes.

> Starting a game from the File menu doesn't update the bootable list.

is this isolated to file menu or also booting from the bootables view?
since i noticed (1) new games booted through file menu are not added to the boot view (thats because we previously relied on the dialog closing/reopening to update the list), but (2) also games booted from bootable view dont update their order in "recently played" for the same reason. (I just want to make sure its just the same problem we're targetting).

jpd002:
Yep, I think they are both the same problem.

jose-c-mar:
M

PR #1162, opened 2022-01-20
Bootable layout cleanup

Zer0xFF:


Source/ui_qt/QBootablesView.cpp
```diff
@@ -246,7 +267,7 @@ void QBootablesView::on_awsS3Button_clicked()
 		if(new_entry)
 		{
 			AsyncUpdateStatus("Refreshing Model.");
-			resetModel();
```
Zer0xFF, line 249: not sure entirely whats happening here, it seems CC segfault if its gets another future thats registered inside a future?

jpd002: I guess this is because the new future is being registered while we are iterating over the vector here: https://github.com/jpd002/Play-/blob/master/Source/FutureContinuationManager.h#L20. 

If registering within a callback would make code simpler, I wouldn't be against changing the ContinuationManager to handle this (like, we could collect a list of callback inside the `remove_if` iteration instead of calling the callbacks directly).

Zer0xFF: > If registering within a callback would make code simpler

not necessarily, atleast in this very case, we already have the async signal function, so it works fine by just using it, but its something to note incase it becomes a problem in the future.

jpd002:
This is working good, thanks!

Just wondering... would is be easy to block the whole bootables interface while there is ongoing work? Because it's possible to press "Add game..." and then press "Refresh" while it's still doing work. It didn't seem to crash when I tried that, but I'm not sure it won't cause undesired effects.

Zer0xFF:
that should be straight forward.

though something to note, even if we disable this interface, the user can still boot games (through menu options, which will trigger `m_bootables` update), so I dont there is a point in preventing booting (unless we also disable those), but even just this should minimise chances of any race conditions or crash happening.

jpd002:
This is good enough, thanks a lot!

PR #1168, opened 2022-02-19
Qt state

Zer0xFF:
- [x] Windows
- [x] MacOS
- [x] Linux (requires https://github.com/jpd002/Play--Framework/pull/33)

this seems to be working well on Windows/MacOS and is also working on Linux (partially requires PR).

Just to note, this https://github.com/jpd002/Play-/pull/1168/files#diff-339f49333be859a94e15ae84bfb784283133b333f468281a3003b2e9f8d3cdb8R162 can be "improved" performance wise if we instead left join in the 2 functions that `select * from bootables`, but considering the size of the db, the speed diff is almost going to be negliable, i opted for cleaner code instead.

### DIFF
--- a/.github/CI_MISC/build_compatibility_db.py
+++ b/.github/CI_MISC/build_compatibility_db.py
@@ -1,0 +1,112 @@
+
+import datetime
+import requests
+import sqlite3
+import re
+import os.path
+import os
+
+DB_NAME = 'states.db'
+
+def getStateLabels():
+	gh_token = os.environ['GITHUB_TOKEN']
+	r = requests.get("https://api.github.com/repos/jpd002/Play-compatibility/labels", headers={'Authorization': f'token {gh_token}'})
+
+	assert(r.status_code == 200)
+	assert(r.headers['content-type'] == "application/json; charset=utf-8")
+
+	res = []
+	for entry in r.json():
+		d = {}
+		d["name"] = entry["name"]
+		d["color"] = entry["color"]
+		res.append(d)
+	return res
+
+def requestGamesStateUpdates(url, res):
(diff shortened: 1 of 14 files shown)


.github/CI_MISC/build_compatibility_db.py
```diff
@@ -0,0 +1,112 @@
...
+def updateDatabase(con, games, labels):
+	con.execute('BEGIN')
+	cur = con.cursor()
+
+	cur.execute("DELETE FROM labels;")
+	for label in labels:
+			cur.execute("INSERT OR IGNORE INTO labels(name, color) values(?, ?);", (label['name'], label['color'],))
+
+	for game in games:
+		print(f"Updating {game['discId']}")
```
jpd002, line 94: I would remove printing every single disc id on the console. There's 1.5k right now and it's just going to grow.

.github/CI_MISC/build_compatibility_db.py
```diff
@@ -0,0 +1,112 @@
...
+	URL = "https://api.github.com/repos/jpd002/Play-compatibility/issues?state=open&per_page=100&since={}&page={}"
+
+	today = datetime.date.today()
+	yesterday = today - datetime.timedelta(days=2)
+
+	i = 1
+	res = []
+	hasNext = True
+	while(hasNext):
+		print(f"Retrieving Page {i}")
```
jpd002, line 57: I would remove this as it doesn't really give interesting information (like for disc ids).

jpd002:
Added a few comments, but other than that, this is looking good!

jpd002:
Thanks a lot for this!

Zer0xFF:
my pleasure :)

PR #1171, opened 2022-02-24
Qt State: fix condition

Zer0xFF:
this conditional limitation is not needed, as it breaks filtering with none-states.


jpd002:
Makes sense! Thanks!

PR #1173, opened 2022-03-04
Handle retrieving root file record via dot

kernle32dll:
Fixes Constantine not booting.

I'm not entirely sure if this is the most elegant solution, but it works. The alternative could be to store the root entry via "." instead of "", but maybe this could break other games?

### DIFF
--- a/Source/iop/ioman/OpticalMediaDevice.cpp
+++ b/Source/iop/ioman/OpticalMediaDevice.cpp
@@ -48,6 +48,8 @@ DirectoryIteratorPtr COpticalMediaDevice::GetDirectory(const char* devicePath)
 	if(!m_opticalMedia) return nullptr;
 	std::string fixedString(devicePath);
 	std::transform(fixedString.begin(), fixedString.end(), fixedString.begin(), &COpticalMediaDevice::FixSlashes);
+	fixedString.erase(fixedString.find_last_not_of('.') + 1); // Remove trailing dot, if there is any
+
 	auto fileSystem = m_opticalMedia->GetFileSystem();
 	auto directoryStream = fileSystem->OpenDirectory(fixedString.c_str());
 	if(directoryStream == nullptr)


jpd002:
Cool :) Yeah, I'm not sure if that's the best way to fix it. Is the game using CdSearchFile to look for the directory? I think I would try to add the logic there if that's the case.

kernle32dll:
The stack goes like this:

```
COpticalMediaDevice::GetDirectory
CISO9660::OpenDirectory
CISO9660::GetFileRecord
CISO9660::GetFileRecordFromDirectory
```

Basically the game is listing files on the disc root, to then load things via the result. Before you implemented `COpticalMediaDevice::GetDirectory`, I made it return an empty list, and that did not work :D

The root cause is, that the way the `ISO9660::CPathTable` is initialized, the root folder is an empty string. And I don't think we want to tamper with that.

I implemented a new approach, where I strip trailing dots (so basically `/.` really is just `/`) for the `COpticalMediaDevice::GetDirectory` call. I think this is the better solution, but I don't think if the location is approiate. Then again, down the line we are only using `const char*` parameters, which make path / filename modifications pretty annoying.

jpd002:
I think it makes sense to put it in `OpticalMediaDevice`. Since it's possible to pass relative paths to device operations such as `Dopen`, and also because each device seem to maintain their own current directory (changed through `chdir`), the device is the only one that can resolve relative paths.

Like, if a game would call `Dopen("cdrom0:..\test")`, the device would be the only one to know what directory it really points to.

I think the fix you did is good for now: I don't want to go info full blown canonicalization at the moment. I just need to run some tests and if everything's alright, I'll merge your branch.

kernle32dll:
> I think the fix you did is good for now: I don't want to go info full blown canonicalization at the moment. I just need to run some tests and if everything's alright, I'll merge your branch.

I just realized, that a call which ends in ".." would probably break now :( So the erase call should probably be `/.`

jpd002:
It's probably fine, I don't think ISO9660 allows filenames to end with . or .. and if a game uses `..`, it's probably not gonna work anyways since we don't have `chdir` yet.

Tested the Big Game Hunter game which also uses `Dopen`, works fine with this change.

Thanks!

jpd002:
Ref: https://wiki.osdev.org/ISO_9660#Filenames.

PR #1176, opened 2022-03-10
Add ioman directory calls to invoke function

kernle32dll:
Adds a few directory related calls to the ioman invoke function. These are used by OPL to do some MC things.


jpd002:
Nice! Thanks for this!

PR #1178, opened 2022-03-14
Implement SifCmd call 7 (SifSetSreg)

kernle32dll:
Makes Hitman Contracts and Hitman Blood Money progress a bit further, but still not quite boot


jpd002:
Nice! Thanks for this! :)

PR #1179, opened 2022-03-16
Add PwrOff IOP module stub

kernle32dll:
Makes ScummVM boot, if LIBSD is provided in rom0

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -319,6 +319,8 @@ set(COMMON_SRC_FILES
 	iop/Iop_PadMan.h
 	iop/Iop_PathUtils.cpp
 	iop/Iop_PathUtils.h
+	iop/Iop_PowerOff.cpp
+	iop/Iop_PowerOff.h
 	iop/Iop_RootCounters.cpp
 	iop/Iop_RootCounters.h
 	iop/Iop_Secrman.cpp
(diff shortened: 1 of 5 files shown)


uyjulian:
ps2sdk is full of hooks, and poweroff modules is one of them.  
I think it's fine to patch/HLE the stuff that hooks.

jpd002:
Agreed, I think it has its place in the emulator. I think we should just be careful to only enable our HLE version if we figure out that the ELF is trying to load that module, just to avoid a conflict in the case a game uses SIF RPC server id 0x09090900.

kernle32dll:
> Is the ELF making any attempt at loading a POWEROFF module? Isn't it bundled with a compiled version of this: https://github.com/ps2dev/ps2sdk/tree/master/iop/dev9/poweroff ?

That is an excellent question, I was asking myself. The thing I'm trying to get running is version 1.7.0 (last version with PS2 support) of [ScummVM](https://www.scummvm.org/). It does come with a precompiled `POWEROFF.IRX`. However, from what I can tell, the ELF does not even try to load that module. Hence I was under the impression it might be BIOS bundled - which is of course silly (*facepalm*).

uyjulian:
It may be loaded by SifLoadModuleBuffer

kernle32dll:
I looked into the source code a bit further, and I found what is happening. So, if the ELF detects being run from `host:`, it skips loading (amongst other things) `POWEROFF.IRX`. @jpd002 suspects that this could be due to for the ELF to be run via `host:`, it needs to be e.g. loaded by ps2link - which loads `POWEROFF.IRX` itself. Would this module be kept loaded, when ps2link loads the ELF @uyjulian?

uyjulian:
As long as the IOP is not rebooted, `POWEROFF.IRX` will be kept loaded. ps2link does not reboot the IOP itself.

jpd002:
Thanks for the confirmation @uyjulian.

Fair enough then, I guess it makes sense to have this in the emulator in its current form. 

Having a ELF/homebrew "runtime environment" is a good idea for the future, we could have something to load some HLE modules specifically in the "Boot ELF" scenario.

jpd002:
Thanks for this!

Issue #1122, opened 2021-09-09
Provide a Flatpak


j8r:
Providing a Flatpak will allow Linux users to easily install and keep Play! up-to-date, on any distribution.
Overall, it is more open and less Ubuntu-centric than snap.

I've already ported some applications to it, so I can do it here too.
What I'll need is your approval to submit the `org.purei.Play-` application to Flathub. Of course, I'll ping you when the PR will be submitted.

CHJ85:
+1 on this request.
Only way to install 3rd party apps easily on the steam deck is through flatpak.

Zer0xFF:
Just a small update, I've gotten a flatpak build working locally, https://gist.github.com/Zer0xFF/c0e3187a94cad177941d03fa9b42a4c2

I do believe its just missing 2 things right now:
1) flatpak recommends to keep flatpak setting seperate from a local install
2) add the build process into CI to autoupdate on master updates

I've been busy lately, so this is taking a bit longer than expected.

j8r:
@Zer0xFF do not worry about auto update yet, I can submit a PR after either by using GitHub Actions or Flatpak External Data Checker, after the repository creation - won't be hard.

Zer0xFF:
@j8r can you explain what PR are you going to submit? from my understanding so far, is that we can upload the build directly to  repo, and i believe flathub is one of them, so where does PR come into this?

j8r:
After your package submission will be approved at https://github.com/flathub/flathub, https://github.com/org.purei.Play- should be created. Then, I'll fork the repo  and make a PR to add the auto update of the commit.

Zer0xFF:
alright, i was under the impression that we can use this https://github.com/marketplace/actions/flatpak-builder
but that does not seem to be the case, atlesat not with flathub.

secondly, Ive updated the missing features list, with point 3, which i guess is important for this process

lastly, I've used the name `org.purei.play` for the package, though not sure if jpd002 would prefer that or `org.purei.Play-`.

jpd002:
I prefer `org.purei.play`.

j8r:
@jpd002 the emulator being `Play!`, with an uppercase letter, won't it make more sense to be `org.purei.Play`?

Zer0xFF:
my 2 cents regarding this, unix systems in general prefer lower case naming convention, mainly to make things easier to type.
in that case that also applies to the `-`

j8r:
[...]

PR #1180, opened 2022-03-18
prepare desktop for flatpak

Zer0xFF:
related: https://github.com/jpd002/Play-Dependencies/pull/32 #1122 

relevant config: https://gist.github.com/Zer0xFF/c0e3187a94cad177941d03fa9b42a4c2

as is, this would work fine, though flatpak prefers to separate app config from non-flatpak apps,
haven't the change to check this out, but I believe flatpak provides a sandboxed data directory that we cn use.


j8r:
Note that this not strictly required to get the Flatpak packages working.
This modifications can be done in the build process, in the `post-install` step.

j8r:
In my opinion the first commit should not be merged as-is. The `Play` named makes more sense than `org.purei.play`. The renaming will be done in the Flatpak build.
For the 512x512 icon is welcome though, maybe smaller ones may be welcome too - have to check this out.

Zer0xFF:
those names are just names, they dont make any different to the operation of either appimage or flatpak, so I dont see any harm in just going with 1 standard.

while the `post-install` can used to "fix" this, i find it redundant adding an extra step to the build system, when it can be avoided.
note, that this cmake step will run regardless of flatpak or not, so we might as well standardise it and keep it consistant

Zer0xFF:
the dash doesnt really make sense., since its meant to be `!` but thats not a valid repo name so github converts it to a dash, and its definitely not going to be a valid package name. as for `Play` vs `play`, as i said, unix prefers lower case, which is how i made my decision, jpd002 can override that decision if he likes as I dont mind either way.

jpd002:
If the doc has a capital first letter for the example app name, then, let's make our package name `org.purei.Play`. The `-` as @Zer0xFF mentioned doesn't belong in the name.

j8r:
Related to the application ID conventions, I just found this https://docs.flatpak.org/en/latest/conventions.html?highlight=conventions#application-ids

jpd002:
Thanks a bunch for this!

Zer0xFF:
@j8r I've prepared the flathub repo request, you can check it out here https://github.com/Zer0xFF/flathub/tree/new-pr

Zer0xFF:
@j8r I've prepared the flathub repo request, you can check it out here https://github.com/Zer0xFF/flathub/tree/new-pr
if im missing anything, please let me know., ill send the request once #1194 is merged.

PR #1181, opened 2022-03-18
Inform about new disc via CdTrayReq

kernle32dll:
On first invocation of CdTrayReq, return 1 for a check request (which indicates a new disc).

Allows Iron Man to boot (probably playable, but its very slow).


jpd002:
Cool, thanks for this!

PR #1188, opened 2022-03-28
Handle negative relocation address for R_MIPS_*16

kernle32dll:
Done by Hitman 2 - Silent Assassin, and The Lord of the Rings: The Fellowship of the Ring. Fixes both games to at least not crash the emulator.


kernle32dll:
Converted to draft, as implementation seems wrong, but I cannot further test for unrelated reasons atm.

PR #1190, opened 2022-03-31
Unregister SIF RPC servers on IOP reset

kernle32dll:
Fixes Golden Eye Rogue Agent stuck on loading screen


jpd002:
Thanks!

Just as a note, the modules that were loaded by the game will still be there even though they don't respond to SIF requests anymore. Probably not a problem for that specific game that's doing multiple reboots in a single session.

PR #1191, opened 2022-03-31
Ignore McServ ChDir commands with invalid port or slot

kernle32dll:
When https://github.com/jpd002/Play-/pull/1190 is merged, 007 Agent Under Fire does crash on startup. With this fix, the game can go in-game with okayish performance.

Note, savegame saving works, but loading does not :(


kernle32dll:
Thinking about this, I might want to create a follow-up MR were we want to add that `-1` handling to all mcserv calls. But I want to debug the game a bit further, to first understand why it sends that `-1` in the first place (could be a clue for some undocumented behavior).

Source/iop/Iop_McServ.cpp
```diff
@@ -605,6 +605,13 @@ void CMcServ::ChDir(uint32* args, uint32 argsSize, uint32* ret, uint32 retSize,
 	assert(argsSize >= 0x414);
 	CMD* cmd = reinterpret_cast<CMD*>(args);
 
+	// 007 Agent Under Fire sends junk for the port
+	if(cmd->port == -1 || cmd->slot == -1)
+	{
+		ret[0] = -1;
```
jpd002, line 611: Could we add a little warning (ex.: `CLog::GetInstance().Warn(...)` here just to catch other games that might be doing this?

jpd002:
Thanks!

PR #1194, opened 2022-04-04
Add AppStream metadata file

j8r:
Outside Flatpak, also useful for App stores and desktop environments.

Validated with `flatpak run org.freedesktop.appstream-glib validate org.purei.Play.metainfo.xml`.

Using a RDNS for the desktop file can be fine, few do it (KDE and I noticed Minestest).

For the version, you *don't have* to keep it up-to-date, but a `<release>` is required for Flatpak.
A `sed` replace will be done in the Flatpak repository.

Finally, will we keep track of master or the releases for the package?


j8r:
Of course, related to #1122 and also #1180

jpd002:
Thanks for this!

And to answer your question, please keep track of `master` for the package.

PR #1197, opened 2022-04-07
fix std::clamp assert, as currentVolume can be 0

Zer0xFF:
just for my reference:

```
# master branch
$ git ls-files --eol CMakeLists.txt Source/iop/Iop_SpuBase.cpp
i/lf    w/crlf  attr/                   CMakeLists.txt
i/crlf  w/crlf  attr/                   Source/iop/Iop_SpuBase.cpp

# local branch
$ git ls-files --eol CMakeLists.txt Source/iop/Iop_SpuBase.cpp
i/lf    w/crlf  attr/                   CMakeLists.txt
i/lf    w/lf     attr/                   Source/iop/Iop_SpuBase.cpp

# checked out `std_clamp_assert`
$ git ls-files --eol CMakeLists.txt Source/iop/Iop_SpuBase.cpp
i/lf    w/crlf  attr/                   CMakeLists.txt
i/lf    w/crlf  attr/                   Source/iop/Iop_SpuBase.cpp
```

the line ending was confusing me a little bit, since git default line ending should be `crlf`
but for this commit, i had to change it to `lf` commit it, and on checking out that same branch, the local source shows up as `crlf`.

basically, if i understood that correctly, remotely git prefer to keep a `lf` version, while locally an `crlf` version, there was a mismatch, thus the need for the clean up commit?


jpd002:
Thanks for this!

Yeah, there's probably a cleanup to do with line endings. Don't think the config is correct.

PR #1204, opened 2022-04-21
Use abs offset in file seek end

kernle32dll:
Fixes Shadow of Ganymede crashing on boot

### DIFF
--- a/Source/ISO9660/File.cpp
+++ b/Source/ISO9660/File.cpp
@@ -36,7 +36,11 @@ void CFile::Seek(int64 amount, Framework::STREAM_SEEK_DIRECTION whence)
 	case Framework::STREAM_SEEK_END:
 		//Cannot seek from end with unbounded files
 		assert(m_end != ULLONG_MAX);
-		m_position = size + amount;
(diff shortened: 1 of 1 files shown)


jpd002:
I'm not comfortable with having this inside `ISO9660` since this is meant to be a more generic thing for reading disc images and the fix we have to do for Shadow of Ganymede (and there's one other game that does that) is IOP specific.

I think we should change `COpticalMediaDevice::GetFile` to return a wrapper around the `CStream*` returned by `CISO9660` that does the adjustment for the `Seek` behavior for that particular case. The code would look a bit like this:

```
void COpticalMediaFile::Seek(int64 offset, STREAM_SEEK_DIRECTION whence)
{
    if((whence == Framework::STREAM_SEEK_END) && (whence > 0))
    {
        //Comment here about Shadow of Ganymede doing weird stuff
        offset = -offset;
    }
    m_baseStream->Seek(offset, whence);
}
```

It's a bit annoying to have to do that, but this way, we keep this in the IOP implementation side of things.

jpd002:
Just as a note, Guerrilla Strike also relies on this behavior.

PR #1246, opened 2023-03-20
Adding more sound buffer sizes

literalmente-game:
There's some values that would better fit some devices that weren't listed


jpd002:
Could you try deleting the `pref_emu_audio_spublockcount_entries` array from language specific files and just have it live in `build_android/src/main/res/values/strings.xml`?

I think you also need to update `pref_emu_audio_spublockcount_values` to have the values matching the descriptions from `pref_emu_audio_spublockcount_entries`.

Issue #1089, opened 2021-06-17
Input from controllers highlights entire screen on OxygenOS


Pesukarhu7:
![Screenshot_20210617-125249_26c022622f54c9f7d3f2019cc473e591](https://user-images.githubusercontent.com/84292564/122374794-2361c080-cf0f-11eb-9de7-decd38ed742f.jpg)

![Screenshot_20210617-125245_26c022622f54c9f7d3f2019cc473e591](https://user-images.githubusercontent.com/84292564/122374976-4ab88d80-cf0f-11eb-8649-58757a5867ce.jpg)

PR #1259, opened 2023-06-26
Android: Fix controller input causing screen highlight

ghost:
Fixes: https://github.com/jpd002/Play-/issues/1089

Fix an issue on Android where input from a controller would cause the SurfaceView to become highlighted

### DIFF
--- a/build_android/src/main/res/layout/emulator.xml
+++ b/build_android/src/main/res/layout/emulator.xml
@@ -38,7 +38,7 @@
 				android:id="@+id/emulator_profile"
 				android:layout_width="wrap_content" 
 				android:layout_height="wrap_content" 
-				android:layout_gravity="end"
+				android:layout_gravity="right"
 				android:typeface="monospace"
 			/>
 			<!-- 


Sykomaniac:
@jpd002 not sure if this is how you want to ultimately implement the fix, the issue is that the SurfaceView is becoming highlighted as Android is, by default, focusing on the view when a user uses a controllers D-pad. The option to disable this behaviour is only available since API 26. There might be a better way to fix this so it works for all API level's? But I'm not an Android dev!

Zer0xFF:
why not add that directly to the main xml? (yes im aware this was only added for api 26, but wouldnt it just be ignored on older versions?)

Sykomaniac:
Will give it a try and update :)

Sykomaniac:
@jpd002 @Zer0xFF changed to just add attibute in required view.

Android does indeed i[gnore unsupported XML attrbiutes](https://developer.android.com/training/basics/supporting-devices/platforms.html#version-codes)

build_android/src/main/res/layout/emulator.xml
```diff
@@ -37,7 +38,7 @@
 				android:id="@+id/emulator_profile"
 				android:layout_width="wrap_content" 
 				android:layout_height="wrap_content" 
-				android:layout_gravity="right"
+				android:layout_gravity="end"
```
jpd002, line 41: Is this an intentional change? If not, probably best to leave it out.

jpd002:
Thanks for this!

Sykomaniac:
Just tried the latest build and working perfectly! Thank you for merging it!

PR #1283, opened 2023-08-24
Update AppConfig.cpp "Portable Mode"

Aynshe:
Add a "portable.txt" file to the root of Play to use the "Play Data Files" folder in the program folder. Or change the folder name and/or path by editing "portable.txt.

### DIFF
--- a/Source/AppConfig.cpp
+++ b/Source/AppConfig.cpp
@@ -1,8 +1,12 @@
 #include "AppConfig.h"
 #include "PathUtils.h"
+#include <iostream>
+#include <fstream>    // Pour la lecture du fichier texte
+#include <filesystem> // Pour les opérations sur le système de fichiers
 
-#define BASE_DATA_PATH ("Play Data Files")
 #define CONFIG_FILENAME ("config.xml")
+#define PORTABLE_BASE_DATA_PATH ("Play Data Files")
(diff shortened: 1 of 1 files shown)


jpd002:
Hey! Thanks for this!

Checking for `portable.txt` next to the emulator's exe path to make the emu portable makes sense, but I don't like the reading of the contents of the file to figure out the path. Paths can have unicode chars in them, so, it makes them a bit annoying to read if it's something the user can input manually. I would get rid of that code.

Aynshe:
Hi,

Thanks for your feedback.

Yes, indeed there can be errors if a user indicates a bad path.

This is why by default the portable.txt file must be empty, change the path is optional to mount the custom path other than the executable folder.

Aynshe:
I come back to you to find out if I understood your answer correctly (English is not my language and I use google trad).

Are you going to take the code proposal into account and edit it to remove the option that allows you to choose another disk path or if I should offer you the code cleaned of the part of the code that does not suit you?

jpd002:
On peut faire en français pour simplifier les choses 😄 

Oui, je préfèrerais que vous changiez votre contribution pour retirer le morceau de code qui lit le contenu du fichier `portable.txt`.

Merci!

PR #1294, opened 2023-09-08
Add Idolm@ster arcadedefs

Keropon:
Added the arcadedefs that are working right now for Idolm@ster. It shares the same HDD with the two types of cabinet of the game, as the dongle is the responsible for deciding which version will it run (Tower or Station).

The main game happens in the Station cabinet, while the Tower shows the nationwide rankings runs a permanent Attract Mode.

As it goes, the Tower functions even without the implementation of Touchscreen support, while the Station still needs the Card RW module and Touchscreen support to be played.

### DIFF
--- a/arcadedefs/idolmtower.arcadedef
+++ b/arcadedefs/idolmtower.arcadedef
@@ -1,0 +1,17 @@
+{
+	"id": "idolm",
+	"name": "The iDOLM@STER/偶像大师",
+	"dongle":
+	{
+		"name": "NM00022 IDMT1, Ver.D a026061342184a [Tower] [Rebuilt].bin"
+	},
+	"hdd":
+	{
+		"name": "NM00022 IDM1-HA (HDD).chd"
+	},
+	"eeFrequencyScale": [4, 3],
+	"boot": "ac0:IDLLOAD",
+	"patches":
+	[
+	]
+}
--- a/arcadedefs/idolm.arcadedef
+++ b/arcadedefs/idolm.arcadedef
@@ -1,0 +1,17 @@
+{
+	"id": "idolm",
+	"name": "The iDOLM@STER/偶像大师",
+	"dongle":
+	{
+		"name": "NM00022 IDMT1, Ver.D a026061342184a [Tower] [Rebuilt].bin"
+	},
+	"hdd":
+	{
+		"name": "NM00022 IDM1-HA (HDD).chd"
+	},
+	"eeFrequencyScale": [4, 3],
+	"boot": "ac0:IDLLOAD",
+	"patches":
+	[
+	]
+}
--- a/arcadedefs/idolm.arcadedef
+++ b/arcadedefs/idolm.arcadedef
@@ -3,7 +3,7 @@
 	"name": "The iDOLM@STER/偶像大师",
 	"dongle":
 	{
-		"name": "NM00022 IDMT1, Ver.D a026061342184a [Tower] [Rebuilt].bin"
+		"name": "NM00022 IDMS1, Ver.D a026061343464a [Station].bin"
 	},
 	"hdd":
 	{


GXTX:
If a maintainer or contributor wishes, feel free to use my project for the card reader, which is known working with this game on hardware. Either directly integrating it is fine, or going the route that TP users use which is to have it connect to a named pipe called `\\.\pipe\YACardEmu`.

https://github.com/GXTX/YACardEmu

I'll look into it myself as i've added support in flycast and Dolphin (although both not mainlined) for the arcade games which require readers.

Keropon:
> If a maintainer or contributor wishes, feel free to use my project for the card reader, which is known working with this game on hardware. Either directly integrating it is fine, or going the route that TP users use which is to have it connect to a named pipe called `\\.\pipe\YACardEmu`.
> 
> https://github.com/GXTX/YACardEmu
> 
> I'll look into it myself as i've added support in flycast and Dolphin (although both not mainlined) for the arcade games which require readers.

Oh my god, I was just modifying the old Card Reader Emulator that was in python, for sure this will be better than it!

arcadedefs/idolmtower.arcadedef
```diff
@@ -0,0 +1,17 @@
+{
+	"id": "idolm",
```
jpd002, line 2: This should have its very own id: `idolmtower` would make sense since you've used that name for this file.

jpd002:
Just as a note, I've been trying to follow MAME's nomenclature for the ids and ROM filenames, but since this game isn't in MAME atm, I guess we'll have to come up with our own. I have many mixed feelings about the whole naming thing since MAME's has some limitations. But I guess we can go ahead with this for now even if it diverges from the actual system.

arcadedefs/idolmtower.arcadedef
Keropon: I was under the impression that the game ID doesn't change since it's the same HDD and game code, it just starts in one state or another. I'll change it if it wasn't the case.

Keropon:
> Just as a note, I've been trying to follow MAME's nomenclature for the ids and ROM filenames, but since this game isn't in MAME atm, I guess we'll have to come up with our own. I have many mixed feelings about the whole naming thing since MAME's has some limitations. But I guess we can go ahead with this for now even if it diverges from the actual system.

Maybe if it changes in the future we could just update it, but going along the lines of the limitations... how hard you think it would be to mantain IDs for both (mame's and play's)? I mean, since the number of games is limited (more so in Arcade games) and the .arcadedefs come packed with each new Play! version

HeroponRikiBestest:
> Just as a note, I've been trying to follow MAME's nomenclature for the ids and ROM filenames, but since this game isn't in MAME atm, I guess we'll have to come up with our own. I have many mixed feelings about the whole naming thing since MAME's has some limitations. But I guess we can go ahead with this for now even if it diverges from the actual system.

Worth mentioning, you or someone else could always just submit the existing dumps to mame, at least as long as they aren't heavily modified (if they are, then please forgive me). I submitted the python game dogstation as a raw hdd with nothing else, despite the fact that python is very nonworking and the game would still need keys/etc to work, and they still added it to the DB. Especially in a case like this, I'm pretty sure you wouldn't meet any resistance submitting iDOLM@STER, and could then use the resulting mameset name,

jpd002:
Yes, right, my bad 😄

HeroponRikiBestest:
> Alright, I'll see what I can do. There's probably other ROM sets I'd like to submit at some point: like there's a lot of issues with CHD files being HDDs dumps instead of DVDs which causes a lot of confusion which I'd like to fix too.

As in, games that came on DVDs being compressed as hard drive chds mistakenly? It would be good to report that to mame too; I know they had that same issue resolved for the Konami Firebeat pop'n music games after the overhaul for that board not too long ago.

jpd002:
> As in, games that came on DVDs being compressed as hard drive chds mistakenly?

Yes, exactly. They added DVD support recently, so, I guess it would be the right time to fix that.

HeroponRikiBestest:
> Yes, exactly. They added DVD support recently, so, I guess it would be the right time to fix that.

Yep; as I mentioned, this already happened with the firebeat pop'n games, so a good time to do so

PR #1295, opened 2023-09-08
Android: Add Spanish Translation

Keropon:


build_android/src/main/res/values-es/strings.xml
```diff
@@ -0,0 +1,183 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+	<!-- Application Navigation -->
+
+	<string name="about_play">Acerca de Play!</string>
+	<string name="about_play_emu">Play! Emulador de PlayStation 2</string>
```
Croden1999, line 7: here is should `Play! PlayStation 2 Emulator`

Croden1999 (changes requested):

build_android/src/main/res/values-es/strings.xml
Keropon: Ok, done. So this string shouldn't be translated... 🤔

jpd002:
Thanks a bunch for this!

Keropon:
You're welcome, you don't know how happy the 256 emulation made me. It's the least I could do.

PR #1296, opened 2023-09-08
Fix dipswitch documentation

HeroponRikiBestest:
I'm not sure what manual https://www.arcade-projects.com/threads/yet-another-256-display-issue.2792/#post-37365 is from, but they appear to have gotten the setting for switch 4 wrong; OFF sets composite sync, and ON sets separate sync. The diagram above even correctly shows this, as all (or at least most) games shipped with all switches set to off for 15khz, and switches 2-4 are all meant to be flipped up for 31khz. Backed up against the F/UC operator's manual (link at https://www.arcade-projects.com/threads/fate-unlimited-codes-japanese-manual-pdf.7799/, direct link https://mega.nz/#!shwzSY5K!E1gzeJaA6oGHNqARNBvsmJB1LlsF6qO6Z9Ke4vLQJ_U in case it's taken down)
fateswitch


jpd002:
Thanks!

PR #1301, opened 2023-09-28
[WIP] Touchscreen Support

Keropon:
Hi!
Here's a first pass at a touchscreen support implementation. This should work on the Touchscreens found on Idolm@ster (Namco 2X6) & Dragon Chronicle (Namco 2X6). I'll try to test the remainder two games Druaga Online: The Story of Aon & Minna de Kitaeru Zenno Training (this one should work because it seems to be a reutilization of the Idolm@ster's Tower unit).
The idea is that Gun & Touchscreen support derive from the same data: The JVS Analogue Controller, so there goes a little name change to some of the Listeners and Setters to reflect that and making it a little easier to follow up... also an initial creation of the functions needed for the implementation.

TO-DO: Now the [mouse is being followed by a barrage of taps](https://youtu.be/1GK6aECWw88). This is not the expected behaviour, as it should only work when the mouse click is being held down. This is necessary as one of the requirements to activate a [hidden code in Idolm@ster to activate Offline Mode](https://www.idolmaster.jp/imas/arcade/idolmaster_offline.pdf) is to press any spot of the touchscreen for at least 10 seconds.

### DIFF
--- a/Source/AnalogueListener.h
+++ b/Source/AnalogueListener.h
@@ -1,0 +1,8 @@
+#pragma once
+
+class CAnalogueListener
+{
+public:
+	virtual ~CAnalogueListener() = default;
+	virtual void SetAnaloguePosition(float, float) = 0;
+};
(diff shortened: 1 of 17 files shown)


jpd002:
Superseded by this: https://github.com/jpd002/Play-/pull/1352.

PR #1305, opened 2023-10-01
Fix Mailbox potential race condition

Zer0xFF:
This might have some performance ramifications, since previously all waits got released once a single call finished.

having said that, I do think emu thread is the only thread that calls on GS, so if its waiting, calling all or once is all the same and will have no impact or notable change at that.


jpd002:
Thanks for this!

PR #1313, opened 2023-10-25
Moderanise Mailbox to use promises and futures

Zer0xFF:
1) remove `m_callFinished`/`m_callDone`
    * future will now handle the waiting
 2) `sync` replaced with `promise`
    * simplify sync logic
       * delegating the logic to the future (if was created)


jpd002:
Very nice improvement! Thanks!

PR #1314, opened 2023-10-26
Update on input

Zer0xFF:
Change `OnInput` from a functional pointer to signal, for our use case, signals are better, since they're mutex protected (thread safe), thus reduce race chance, they simplify the code (no (explicit) guarding required), and allows multiple targets to request input.

In Framework, we also added a new call `CSignal<>::ConnectOverride(...)`, which can be used to temporarly redirect events (e.g when configuring gamepads).

Requires: https://github.com/jpd002/Play--Framework/pull/40 (would need to rebase once that is merged)

Notes: tested on Windows only (changes are simple that if it compiles, we should be fine)


jpd002:
Thanks for this!

PR #1315, opened 2023-10-31
Win32: Add Rumble Support

Zer0xFF:
Requries: https://github.com/jpd002/Play--Framework/pull/41

Add basic Rumble/Vibration Support.

Just to note,
1) I've added few virtual functions, all of them are non-pure (aka they have a stub implementation),
this will allow all platforms to compile, until we add support for the reset (there is also some derviative class that wont support those calls, aka QtProvider (aka keyboard) will never need to implement them)
2) `CMotorBinding` isn't a deriviate of `CBinding` (as the name might suggest)
3) can't test DirectInput support (I dont have any device that currently support it)


Source/input/PH_GenericInput.cpp
```diff
@@ -2,7 +2,8 @@
 
 void CPH_GenericInput::Update(uint8* ram)
 {
-	for(auto* listener : m_listeners)
+	std::map<unsigned int, std::pair<uint8, uint8>> vibrationMap;
```
jpd002, line 5: `std::map` is probably a bit overkill for this. We should use `std::array<std::pair<uint8, uint8>, MAX_PADS>`.

Zer0xFF: updated, I've also added code to choose max value, just incase we implement vibration, in the other interfaces

jpd002:
Thanks for this!

PR #1317, opened 2023-11-15
Update strings.xml

DDinghoya:


build_android/src/main/res/values-ko/strings.xml
```diff
@@ -129,6 +129,14 @@
...
 	</string-array>
 
+	<string-array name="pref_emu_audio_spublockcount_entries">
+		<item>400</item>
+		<item>200</item>
+		<item>100</item>
+		<item>50</item>
+		<item>10</item>
+	</string-array>
+
```
Croden1999, line 139:
```suggestion

```

Croden1999 (changes requested):
no need add here, it is not translation

jpd002:
Thanks for this!

Issue #1329, opened 2023-11-27
Champions: Return to Arms memcard hang


bigianb:
## Quick summary
Champions return to arms PAL version (SLES 530.39) hangs reading from the memory card.

## System Details
MacOS, m1 MacBook Pro
commit: 906c4d2d969bfbee8316d33213d2f6d090ae878a

## Issue Details
Reading from the memory card throws an exception.
The game tries to read /BSLES-52325/BSLES-52325_config which is the config file for Champions of Norrath (previous game in the series). I don't have a save game for that on my memory card and so the directory does not exist. I suspect the IOP module is not returning the correct signal for that. The assert happens because the game tries to read the file with handle 0. Removing that assert (or running in release mode) will cause the game to hang. I changed it to return -4 in this case rather than 0 (no file found) which seems more correct but a hang still happens.

## Screenshots/Videos
iop_mcserv.log
```
Init();
GetInfo(port = 0, slot = 0, wantType = 1, wantFreeSpace = 0, wantFormatted = 0, retBuffer = 0x008FA140);
GetDir(port = 0, slot = 0, flags = 0, maxEntries = 64, tableAddress = 0x004FAE80, name = 'debug.sht');
GetInfo(port = 0, slot = 0, wantType = 1, wantFreeSpace = 0, wantFormatted = 0, retBuffer = 0x008FA140);
GetInfo(port = 0, slot = 1, wantType = 1, wantFreeSpace = 0, wantFormatted = 0, retBuffer = 0x008FA140);
Called mc function with invalid slot 1
GetInfo(port = 0, slot = 2, wantType = 1, wantFreeSpace = 0, wantFormatted = 0, retBuffer = 0x008FA140);
Called mc function with invalid slot 2
GetInfo(port = 0, slot = 3, wantType = 1, wantFreeSpace = 0, wantFormatted = 0, retBuffer = 0x008FA140);
Called mc function with invalid slot 3
GetInfo(port = 1, slot = 0, wantType = 1, wantFreeSpace = 0, wantFormatted = 0, retBuffer = 0x008FA140);
GetInfo(port = 1, slot = 1, wantType = 1, wantFreeSpace = 0, wantFormatted = 0, retBuffer = 0x008FA140);
Called mc function with invalid slot 1
GetInfo(port = 1, slot = 2, wantType = 1, wantFreeSpace = 0, wantFormatted = 0, retBuffer = 0x008FA140);
Called mc function with invalid slot 2
[...]

PR #1330, opened 2023-11-30
McServe fixes for Champions Return To Arms

bigianb:
fixes issue #1329


jpd002:
Also, did you try returning -5 when the game tries to read a invalid fd? That's what seems to happen here: https://github.com/ps2dev/ps2sdk/blob/master/iop/memorycard/mcman/src/main.c#L749.

bigianb:
Bit more info.
When the RPC returns for the Read command on the EE side, this function is always called:
https://github.com/ps2dev/ps2sdk/blob/7efb3b409b9c12067f9455de97fa5190b6bd1037/ee/rpc/memorycard/src/libmc.c#L197C9-L197C9
So, if we don't initialise the first 2 elements to zero then it will have garbage in it and that function will overwrite random memory.

jpd002:
I've pushed the change to move the `paramAddress` area clearing.

For writing homebrew, I think using ps2dev's docker image makes the process a lot easier. It runs a bit slower on Apple Silicon Mac, but it's still pretty good if you wanna write simple things.

PR #1336, opened 2023-12-22
Dead or Alive 2 fix (PAL)

bigianb:
US version may need a different patch. I only have the EU version.

The first change is that DisableIntrLine should return the status. This is later passed to ResumeInterrupts and without the fix the interrupt is never resumed. You're not really meant to use the functions this way (mixing the Line functions with the non-line functions but the game does).

Second is that the game hangs loading files at the start. Not sure why this is (PCSX2 also needs this fix). Changing the RPC call from async to sync fixes it.

We now get in-game with some graphics issues.

Screenshot 2023-12-22 at 13 58 17


Source/iop/Iop_Intrman.cpp
```diff
@@ -165,6 +165,12 @@ uint32 CIntrman::DisableIntrLine(CMIPS& context, uint32 line, uint32 res)
 	mask.f &= ~(1LL << line);
 	context.m_pMemoryMap->SetWord(CIntc::MASK0, mask.h0);
 	context.m_pMemoryMap->SetWord(CIntc::MASK1, mask.h1);
+	if(res != 0)
```
jpd002, line 168: So, I've checked some disassembly of `INTRMAN` and it seems that `res` contains the number of the interrupt that was disabled. I think it also can `or` the interrupt number with 0x200 if the interrupt was already off. Probably would benefit from a more in depth check for the `0x200` part, but could you try returning the interrupt number instead of the interrupt state?

Also some info regarding the `res` value here:
https://github.com/ps2dev/ps2sdk/blob/416eb739a261dba5f565584e06db1782c43377f1/iop/kernel/include/intrman.h#L126

bigianb: That header is what I was going from. It says:
A pointer to a variable that will store the current interrupt status
Which is what the code does. Looking at what the BIOS actually does though is of course better :) I should also extend my functional tests to test this on PCSX2 with a real BIOS ... but that will be after the new year (probably)

bigianb: The Sony docs say:

All interrupts and thread dispatches are disabled.
At the same time, the state that was active immediately before interrupts were disabled is stored in the
variable specified by oldstat (use CpuResumeIntr() to return to that state).

Goatman13:
You may want to read 2 articles linked in this pr: https://github.com/PCSX2/pcsx2/pull/8814

Because without implementing of what is in first article you may came up with really wrong behavior on hle side. Doa is affected by the same sif interrupts quirk as KK2001.

Source/iop/Iop_Intrman.cpp
bigianb: the real bios does this:

```c++
int DisableIntr(int interrupt, int *res)
{
    int low_irq = interrupt & 0xFF;

    int ictrl;
    CpuSuspendIntr(&ictrl);

    int stat = -103;

    int ret = 0;
    if (low_irq < 0x20) {
        int old_IMASK = IMASK;
        IMASK &= ~(1 << low_irq);

        if (old_IMASK & (1 << low_irq)){
            stat = low_irq;
        } else {
            ret = -103;
        }
    } else if (low_irq < 0x27) {
        if (DMA_ICR & 0xFFFFFF & (1 << (low_irq - 0x10))){
            if (((DMA_ICR & 0xFFFFFF) >> (low_irq - 0x20)) & 0x01){
                stat = low_irq | 0x100;
            }
            if (DMA_ICR2 & (1 << (low_irq - 0x20))){
                stat |= 0x200;
            }
            DMA_ICR2 &= ~(1 << (low_irq - 0x20)) & 0xFFFFFF;
        } else {
            ret = -103;
        }
    } else if (low_irq < 0x2E) {
        if (DMA_ICR2 & 0xFFFFFF & (1 << (low_irq - 0x18))) {
            stat = low_irq;
            if (((DMA_ICR2 & 0xFFFFFF) >> (low_irq - 0x21)) & 0x01){
                stat |= 0x200;
            }
            DMA_ICR2 &= ~(1 << (low_irq - 0x18)) & 0xFFFFFF;
        } else {
            ret = -103;
        }
    }

    if (res){
        *(volatile int *)res = stat;
    }
    CpuResumeIntr(ictrl);
    return ret;
}
```

So it looks more complex than described. for 0-31 it does what it says on the tin. For values greater than that though it does something different and can set the flags at 0x100 and 0x200.
The enable routine has similar code to handle this.
I think DOA passes a value < 0x20 so it's fine. It would be good to find a game which actually passes in a value >= 0x20 so that we can understand what it is doing better.

I can add the correct implementation but it will take a while to write some test code to verify it works like the real bios.

bigianb:
> You may want to read 2 articles linked in this pr: [PCSX2/pcsx2#8814](https://github.com/PCSX2/pcsx2/pull/8814)
> 
> Because without implementing of what is in first article you may came up with really wrong behavior on hle side. Doa is affected by the same sif interrupts quirk as KK2001.

Interesting - I remember reading those a while ago but forgot. It explains the patch and I think that's the best way it can be done. Regarding the interrupt handling ... that's talking about how the hardware interprets changes to the registers isn't it and not the BIOS behaviour. That wouldn't be something to worry about in the bios code but something to implement in the hardware virtualisation layer (if it is not already).

patches.xml
```diff
@@ -2,4 +2,10 @@
 	<Executable Name="SLES_501.76;1" Title="Oni" Region="EU">
 		<Patch Address="0x001cef7c" Value="0x00000000 // bc0f $001cef7c" Description="Fix hang by skip branch if copro 0 condition false." />
 	</Executable>
+	<Executable Name="SCES_500.03;1" Title="Dead or Alive 2" Region="EU">
+		<Patch Address="0x002b4c44" Value="0x24060000" Description="Make the file reading RPC call synchronous. Boot hangs without this." />
+	</Executable>
+	<Executable Name="SLPS_250.02;1" Title="Dead or Alive 2" Region="JP">
```
bigianb, line 8: The SLPS one is a guess as I don't have that version. Strelok confirmed it worked though ... just not sure if the region is correct or if it even matters.

jpd002:
Nice work! Thanks a lot for this!

Issue #1227, opened 2023-01-13
Building collection, scanning games on filesystem


NtandoNdima:
When I open the play emulator on my iPhone 7 it takes time to build collection and it also takes time to scan for games on the filesystem 
![Uploading 73F15691-84A6-4FDA-B8BA-B0C442585E44.png…]()

tqoo:
Same issue here
iPad Pro 10.5-inch on iPadOS 15.5 jailbroken on palera1n
Did you find a solution?

PR #1337, opened 2023-12-23
Don't scan /private/var/mobile

CastanaEddd:
Results in large loading times on jailbroken devices and causes some devices to hang.
Probably something that shouldn't be done on launch, maybe have this as a manual button/toggle inside the app?
Fixes https://github.com/jpd002/Play-/issues/1227

### DIFF
--- a/Source/ui_ios/CoverViewController.mm
+++ b/Source/ui_ios/CoverViewController.mm
@@ -56,14 +56,14 @@ static NSString* const reuseIdentifier = @"coverCell";
 	dispatch_queue_t queue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
 	dispatch_async(queue, ^{
 	  auto activeDirs = GetActiveBootableDirectories();
-	  if(forceFullDeviceScan || activeDirs.empty())
+	  if(forceFullDeviceScan)
 	  {
 		  dispatch_async(dispatch_get_main_queue(), ^{
 			alert.message = @"Scanning games on filesystem...";
 		  });
 		  ScanBootables("/private/var/mobile");
 	  }
-	  else
+	  else if(!activeDirs.empty())
 	  {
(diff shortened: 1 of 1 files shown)


CastanaEddd:
Before:
https://files.catbox.moe/9qpdyp.mp4

After:
https://github.com/jpd002/Play-/assets/128545778/4a67606c-51e1-4aa2-8c9b-1e1ae68f5c53

CastanaEddd:
Sorry for the delayed response this never showed up in my inbox for some reason.
Also I'm not quite sure what you're asking about... You can view the changes I've made with this pull request in the "files changed" section of this page. 

If you're confused on what's happening, basically on launch Play! will attempt to scan `/private/var/mobile` if an "active directory?" is not specified (_default behavior_) or if a full device scan is forced (_done when you scan through the settings page_). For some reason this causes a bit of lag and results in the device being stuck on "Building Collection" indefinitely. 
What this pull request does is it changes it so that it only does that scan when a full device scan is forced. 

This normally isn't a problem unless given permission to, something that's done on jailbroken installs and probably trollstore too when ran unsandboxed.

A compiled ipa with this change can be found in the [github actions page of this fork.](https://github.com/CastanaEddd/Play-/actions/runs/7733690350)

CastanaEddd:
Hmm seems like the issue you're facing might be unrelated to what this patch is trying to address, but I'm glad to hear you got it working :)

Also on an unrelated note trollstore recently got [native support](https://github.com/opa334/TrollStore/commit/647f43087c2c554c13e5aced60a3cbe86815ac64) for launching apps with JIT. With the new update you can open trollstore then launch Play! with JIT from there. Combined with the new URL-scheme you can create a shortcut to automatically enable and launch the app with JIT. Could potentially be a better alternative to whatever tweak you're trying to inject to into it.

jpd002:
I've been super out of the loop with iOS stuff, sorry about that. Since everyone seems to agree this is a good change, I'll merge it in. Thanks!

PR #1347, opened 2024-02-01
Gauntlet 7 sorrows

bigianb:
implements the VU edge case where an integer altering instruction is in the delay slot of a conditional branch and the first instruction following it is another conditional branch based on that integer register.
The original code will take the value of the integer register at that time assuming a pipeline shortcut but that is wrong because having an integer altering instruction directly before a conditional branch on that same integer register causes the pipeline to fail and we need to use the value most recently written to memory.

Can potentially break a lot of games so needs quite a bit of testing.

### DIFF
--- a/Source/ee/VuBasicBlock.cpp
+++ b/Source/ee/VuBasicBlock.cpp
@@ -442,9 +442,9 @@ CVuBasicBlock::INTEGER_BRANCH_DELAY_INFO CVuBasicBlock::ComputeTrailingIntegerBr
 	uint32 fmacDelayOnBranch = fmacStallDelays[fmacStallDelays.size() - 2];
 	if((endLoOps.writeI != 0) && !endLoOps.branchValue && (fmacDelayOnBranch == 0))
 	{
-		// we need to use the value of intReg 3 steps prior or use initial value.
+		// we need to save the value of the integer register before the instruction is executed
 		result.regIndex = endLoOps.writeI;
-		result.saveRegAddress = std::max<int32>(adjustedEnd - 5 * 8, m_begin);
+		result.saveRegAddress = adjustedEnd;
 	}
 	return result;
 }
--- a/Source/ee/VuBasicBlock.cpp
+++ b/Source/ee/VuBasicBlock.cpp
@@ -35,6 +35,8 @@ void CVuBasicBlock::CompileRange(CMipsJitter* jitter)
 
 	auto integerBranchDelayInfo = ComputeIntegerBranchDelayInfo(fmacPipelineInfo.stallDelays);
 
+	auto integerTrailingDelayInfo = ComputeTrailingIntegerBranchDelayInfo(fmacPipelineInfo.stallDelays);
+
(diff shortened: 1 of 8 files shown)


Source/ee/VuBasicBlock.cpp
```diff
@@ -325,6 +430,29 @@ bool CVuBasicBlock::IsNonConditionalBranch(uint32 opcodeLo)
...
+	uint32 endOpcodeLo = m_context.m_pMemoryMap->GetInstruction(endOpcodeLoAddr);
+	auto endLoOps = arch->GetAffectedOperands(&m_context, endOpcodeLoAddr, endOpcodeLo);
+
+	// assume this is same as for the case if there was a conditional branch at the end.
+	uint32 fmacDelayOnBranch = fmacStallDelays[fmacStallDelays.size() - 2];
+	if((endLoOps.writeI != 0) && !endLoOps.branchValue && (fmacDelayOnBranch == 0))
+	{
+		// we need to use the value of intReg 3 steps prior or use initial value.
+		result.regIndex = endLoOps.writeI;
+		result.saveRegAddress = std::max<int32>(adjustedEnd - 5 * 8, m_begin);
```
bigianb, line 451: I'm not convinced on the 5 value here - I think it may need to be 4.

Source/ee/VuBasicBlock.cpp
```diff
@@ -199,6 +261,46 @@ void CVuBasicBlock::CompileRange(CMipsJitter* jitter)
...
+				        m_State.nCOP2VI[m_State.savedNextBlockIntRegIdx] = m_State.savedNextBlockIntRegValTemp;
+				    }
+				 */
+				auto branchLoOps = arch->GetAffectedOperands(&m_context, address, opcodeLo);
+
+				auto resetDelayedValLabel = jitter->CreateLabel();
+
+				jitter->PushRel(offsetof(CMIPS, m_State.savedNextBlockIntRegIdx));
+				jitter->PushCst(branchLoOps.readI0);
+				jitter->BeginIf(Jitter::CONDITION_EQ);
```
jpd002, line 282: Would it work to do something like that?
```
jitter->PushRel(offsetof(CMIPS, m_State.savedNextBlockIntRegIdx));
jitter->PushCst(branchLoOps.readI0);
jitter->Cmp(CONDITION_EQ);
jitter->PushRel(offsetof(CMIPS, m_State.savedNextBlockIntRegIdx));
jitter->PushCst(branchLoOps.readI1);
jitter->Cmp(CONDITION_EQ);
jitter->Or();
jitter->PushCst(0);
jitter->BeginIf(Jitter::CONDITION_NE);
{
    //... Logic here
}
jitter->EndIf();

bigianb: ah yes - I missed the Cmp instruction. That's much more expressive.

bigianb: changed and squashed.

PR #1352, opened 2024-02-15
Touchpanel Support

LapisRegia:
Implements touch panel functionality used in games such as The Idolm@ster, Minna De Kitaeru Zenno Training (ZNT), Druaga Online, and Dragon Chronicle, and allows the player to set The Idolm@ster to Offline mode (Doesn't bypass shutter and reader/writer check). As of now, only Idolm@ster has been tested to work as ZNT, Druaga and Dragon Chronicle doesn't boot at the moment.

Sets m_jvsGunPosX to 0xFFFF upon mouse release, otherwise the game will register the last clicked position as an input.
ReleaseTouchPosition() was put into GunListener.h for ease of merging, perhaps renaming the file to ScreenPositionListener would be better?


jpd002:
Disregard the ARM32/ARM64 build errors, they're my fault.

Keropon:
+1 on changing GunListener's name. 

`ScreenPositionListener` should work but I don't know if the Analogue JVS data will always carry Screen Position data, so I would vouch for something that sticks more to the JVS documentation like JvsAnalogueListener.

On that vein, I was thinking that `GunPos` should be `JvsAnalogue`, and variables like `SetLightGunXform` should be something like `SetJvsAnalogueXform` for better readability?

LapisRegia:
I was thinking from the perspective of the games which use them. The data sent by the Lightguns and Touchpanels are in essence screen position input data, both of which are sent back from IO board to the machine continuously. While there are exceptions (TCR4 with its IR lightguns), and actual hardware is analog, it doesn't beat the fact that all it really does is tell the game where the player is pointing on the screen. Which is why I suggested `ScreenPositionListener`.

Keropon:
I'd prefer system over games but it makes total sense! Here I thought that there were some JVS driving games use the analog channel for the steering wheel data, but alas, I reserched a lil' bit and it seems that there are **no games that do so on the S2X6**, so... `ScreenPositionListener` is perfect.

jpd002:
Thanks for this! Lemme do some tests to make sure everything works fine, but this looks sensible.

jpd002:
Working great! Probably want to add a check here to prevent fullscreen mode to be toggled: https://github.com/jpd002/Play-/blob/master/Source/ui_qt/mainwindow.cpp#L829

Also, did you manage to enable offline mode with this? I tried it, but I couldn't get it to work using the instructions provided in the official guide.

LapisRegia:
> Working great! Probably want to add a check here to prevent fullscreen mode to be toggled: https://github.com/jpd002/Play-/blob/master/Source/ui_qt/mainwindow.cpp#L829

Right, I'll add it in the next push.
> Also, did you manage to enable offline mode with this? I tried it, but I couldn't get it to work using the instructions provided in the official guide.

I did, you just have to press the service switch, then enter switch and then test switch. After that you'll be in the seemingly normal test menu. Press anywhere on the screen for more than 10 seconds, then a Game Mode label will appear above coin options. In there you can set the operation mode to offline or online. It should look like this.
image
image
Notice the greyed out location pass and the lack of the network icon on the bottom left. It should look like that once you set it to offline. After setting it to a different operation mode, you need to reboot the game as the game softlocks you from doing anything else.

jpd002:
Thanks for this and for the confirmation that Offline mode can be enabled!

PR #1353, opened 2024-02-17
Change Xcode debugging note in README

Spidy123222:
Noticing from reading the readme. I saw that it mentioned needing a paid developer account for debugging the app. You do not need one at the moment. I’m not sure even if there was a time that you needed one for debugging the app. You also don’t need a paid dev account for using the app or compiling it.

I changed these to reflect how using Xcode is currently. It also describes detaching the debugger to act like how the jit enablers do it which is just the detach button.


jpd002:
Thanks for clarifying the process! I've pushed a commit that changes the wording a little bit.

PR #1359, opened 2024-03-05
Build libretro core for iOS

warmenhoven:
Similar to the flycast core, I've added a check that retroarch does on iOS to see if JIT is likely to work, and fail to load the game if JIT is not enabled.

In order to pick up that environ flag, I had to pull in a newer version of libretro.h. There shouldn't be any functional change, but the diff looks abnormally large because 1) it's been a while and 2) it looks like there was code style formatting applied to the play! version?

### DIFF
--- a/Source/ui_libretro/main_libretro.cpp
+++ b/Source/ui_libretro/main_libretro.cpp
@@ -493,6 +493,19 @@ bool retro_load_game(const retro_game_info* info)
 {
 	CLog::GetInstance().Print(LOG_NAME, "%s\n", __FUNCTION__);
 
+#if defined(IOS)
+	bool can_jit = false;
+	if(g_environ_cb(RETRO_ENVIRONMENT_GET_JIT_CAPABLE, &can_jit) && !can_jit)
+	{
+		// trying to run without the jit will cause a crash.
+		retro_message retromsg;
+		retromsg.msg = "Cannot run without JIT";
+		retromsg.frames = 5000 / 17;
+		g_environ_cb(RETRO_ENVIRONMENT_SET_MESSAGE, &retromsg);
+		return false;
+	}
+#endif
+
 	fs::path filePath = info->path;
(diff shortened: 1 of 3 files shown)


Source/ui_libretro/main_libretro.cpp
```diff
@@ -493,6 +493,19 @@ bool retro_load_game(const retro_game_info* info)
 {
 	CLog::GetInstance().Print(LOG_NAME, "%s\n", __FUNCTION__);
 
+#if defined(IOS)
+	bool can_jit;
```
jpd002, line 497: Could we default init this to `false` just to be sure?

warmenhoven: Yes of course, I've updated the PR with that change.

jpd002:
Thanks for this! Yeah, the `libretro.h` file has been formatted mistakenly in the past, but it's been removed from the script that checks formatting since then, so, it's a good thing that it's been restored to its good state.

Source/ui_libretro/main_libretro.cpp
```diff
@@ -493,6 +493,19 @@ bool retro_load_game(const retro_game_info* info)
 {
 	CLog::GetInstance().Print(LOG_NAME, "%s\n", __FUNCTION__);
 
+#if defined(IOS)
```
jpd002, line 496: Oh, are you sure `IOS` is defined, I'm not sure this is set in any CMake file or in the iOS SDK.

warmenhoven: I actually am, https://git.libretro.com/libretro-infrastructure/ci-templates/-/blob/master/ios-cmake.yml?ref_type=heads#L14, but I'm happy to change the preprocessor check to something like `#if TARGET_OS_IPHONE` if you prefer.

warmenhoven: Thanks very much!

PR #1360, opened 2024-03-07
Build libretro core for tvOS

warmenhoven:


Source/ee/EeExecutor.cpp
```diff
@@ -60,7 +60,7 @@ void CEeExecutor::AddExceptionHandler()
 	sigemptyset(&sigAction.sa_mask);
 	int result = sigaction(SIGSEGV, &sigAction, nullptr);
 	assert(result >= 0);
-#elif defined(__APPLE__)
+#elif defined(__APPLE__) && !TARGET_OS_TV
```
jpd002, line 63: Is there really no way to use the `mach_*` functions on tvOS? Without the exception handler, it's probably gonna cause unexpected issues due to JIT code mismatch.

warmenhoven: Let me do some more research and get back to you. Those calls have deliberately been made unavailable on tvOS but let me see if there are alternatives.

warmenhoven: Both mach_msg and mach_msg_override are unavailable to tvOS. It seems that ppsspp just uses the standard posix signal() handling, which seems to work here as well. I've updated the PR. Thank you for asking me to take another look!

jpd002: Humm, I've tried the `signal` based exception handling and I'm not quite sure it works on Apple platforms. I've tried it on macOS and the signal doesn't get processed correctly and the emulator will just shut down. You can try FF10, I think that one will cause some exceptions to be triggered and the emulator will crash.

I think you're probably better off defining `DISABLE_PROTECTION` or going back to your previous version which removed the `mach_*` calls (also make sure to remove `mprotect`).

warmenhoven: Thanks, let me experiment more. tvOS is quite a bit different from macOS (and slightly different from iOS in subtle ways, though it is considered an iOS variant).

warmenhoven: > I think you're probably better off defining DISABLE_PROTECTION

Thank you very much for this, this is what I've found works best on tvOS. I did use FFX and did see that my second attempt didn't work. Thank you very much for the suggestions!

jpd002: Glad it works!

jpd002:
Thanks!

PR #1367, opened 2024-04-09
Fix error: ‘clamp’ is not a member of ‘std’

pstef:
by including `<algorithm>`.


jpd002:
Nice, thanks for this!

PR #1401, opened 2024-09-12
docs: update README.md

eltociear:
proccess -> process


jpd002:
Oh, good catch! Thanks for that!

PR #1411, opened 2024-10-20
qt: Allow compiling on Unix-likes without EVDEV

alarixnia:
This is needed to compile on NetBSD, OpenBSD, illumos, etc.


jpd002:
Makes sense! Thanks a lot for this!

PR #1417, opened 2024-10-22
Fix a compilation error on some platforms.

alarixnia:
std::max expects both arguments to be the same type (in this case const char *), while the C version of strrchr returns non-const char *, which results in a type error with some C libraries.


jpd002:
Ah, I didn't know about that subtlety! Thanks!

PR #1419, opened 2024-10-24
Qt & build system improvements

Zer0xFF:


Source/ui_qt/QBootablesView.cpp
```diff
@@ -31,10 +31,37 @@ QBootablesView::QBootablesView(QWidget* parent)
 	m_proxyModel = new BootableModelProxy(this);
 	ui->listView->setModel(m_proxyModel);
 
-	CAppConfig::GetInstance().RegisterPreferenceInteger("ui.sortmethod", BootablesDb::CClient::SORT_METHOD_NONE);
-	m_sortingMethod = CAppConfig::GetInstance().GetPreferenceInteger("ui.sortmethod");
+	CAppConfig::GetInstance().RegisterPreferenceInteger("ui.sortmethod.v2", BootablesDb::CClient::SORT_METHOD_NONE);
+	m_sortingMethod = CAppConfig::GetInstance().GetPreferenceInteger("ui.sortmethod.v2") & BootablesDb::CClient::SORT_METHOD_NONE;
```
jpd002, line 35: Do we need to `and` the value here since this is a whole new value?

jpd002:
Looking good, I'll merge after we get the question sorted out.

Source/ui_qt/QBootablesView.cpp
Zer0xFF: just basic user input sanitisation, its not really needed as you mentioned, but user could edit config manually or a future user reverts to this version and is using an invalid newer enum (though worst case is that we show all user games, unsorted, and the sort option will just show a blank option, until the choose a different sorting option). I can remove it if you like, it has very little effect 1 way or another

however, since you mentioned that, I realised we also need to update android side, since enum where rearranged and dont match what we have on java side. so we need to change preference name to v2.

jpd002: Oh yeah, we only want `recent` or `none`, that makes sense.

jpd002:
Thanks!

PR #1420, opened 2024-10-25
Build on more Unix-like OSes (BSD, illumos, etc)

alarixnia:


jpd002:
Thanks again for these!

PR #1421, opened 2024-10-25
Cleanup/Fix: Fix list order list to reflect new ordering, use new pref name for sorting

Zer0xFF:
It seems I missed few android related changes. (I assume there are iOS changes as well that would need to be updated :S)

### DIFF
--- a/Source/ui_android/java/com/virtualapplications/play/BootablesInterop.java
+++ b/Source/ui_android/java/com/virtualapplications/play/BootablesInterop.java
@@ -8,7 +8,7 @@ public class BootablesInterop
 	}
 
 	// changes to SORT_* order should also be reflected in
-	// NavigationDrawerFragment.onActivityCreated adpter string list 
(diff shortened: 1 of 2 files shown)


Source/ui_android/java/com/virtualapplications/play/BootablesInterop.java
```diff
@@ -7,6 +7,8 @@ public class BootablesInterop
 		System.loadLibrary("Play");
 	}
 
+	// changes to SORT_* order should also be reflected in
+	// NavigationDrawerFragment.onActivityCreated adpter string list
```
jpd002, line 11:
```suggestion
	// NavigationDrawerFragment.onActivityCreated adapter string list 
```

Zer0xFF:
Do you mind squash merging, to avoid the 2nd commit.

jpd002:
Done!

PR #1437, opened 2025-01-08
CI : update upload-artifact version

Mrlinkwii:
actions/upload-artifact@v3 is being Deprecated at the end of the month  see https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/  so i move its to the next version 

hopefully this dosent break anything , but it shouldn't


jpd002:
Nice! Thanks for this!

PR #1440, opened 2025-02-01
CI: Use Native arm64 linux host for arm64 linux CI builds

Zer0xFF:


jpd002:
Awesome 😄 No more 1 hour builds for ARM platforms, yay! Thanks a lot for this!

PR #1472, opened 2025-06-24
CMake updates

Zer0xFF:
"Requires" https://github.com/jpd002/Play--Framework/pull/46 & https://github.com/jpd002/Play-Dependencies/pull/37
(well, it doesnt actually "require" those, this can still be merged as a standalone PR, but for the cmake changes to be complete, its best to have those 2 PR merged 1st)

- Updated libchdr to upstream
  - to update cmake min requirement
  - includes required header fix
- Changed cmake min version to 3.18
- I removed references to  as much as possible references to `cmake_minimum_required` and `set(CMAKE_MODULE_PATH ...)`
  - we can probably go a step further and also remove the `include(Header)` in the same scripts, but its probably good to have as a basic assert (if path is not set, cmake generation will fail, which will indicate an issue in cmake hierarchy)
  - the reason those specific script had this removed, is because we dont expect those to ever be used as a standalone, and any callee should already have set `CMAKE_MODULE_PATH` (and even included `Header`) 
    - thats is also why I've kept the `ui_*/CMakeLists.txt`, since its valid to call `add_subdirectory(...)` on those (skipping the root cmake)

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.18)
 
 set(CMAKE_MODULE_PATH
 	${CMAKE_CURRENT_SOURCE_DIR}/../deps/Dependencies/cmake-modules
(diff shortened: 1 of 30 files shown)


jpd002:
Very nice improvements! Thanks again for this!

PR #1491, opened 2025-09-21
Namco 246 - ACRAM emulation improvements + Battle Gear 3 Steering PCB emulation

derole1:
This PR addresses the following:
- The AcRam iop module now calls the callback provided to it during initalisation upon completion of a read/write if one was passed, this fixes Battle Gear 3 indefinetly hanging on the "Taito" logo and the loading screen when going ingame
- A rudimentary steering board emulator has been added to the Sys246 iop module, the values have been pulled from a mixture of packet captures of a real board and some reverse engineering work done with a newer pc-based version of the game which shares the same board. The board in question is labelled with a model number of "UDDF41B1TA", with silkscreen "ADHP0077ZA", and I believe this is the only 246 game that uses it

The game is now functionally at parity with the PS2 port, which sits on a black screen with only sound when loading into a race. Interestingly the GS falls completely silent and framebuffer updates halt, however the game still responds. This is with both OpenGL and Vulkan backends.

Any constructive critisim or advice is greatly appreciated :)

### DIFF
--- a/GameConfig.xml
+++ b/GameConfig.xml
@@ -88,6 +88,10 @@
 	<GameConfig Executable="SLPM_657.75;1" Title="DDR Festival: Dance Dance Revolution">
 		<BlockFpRoundingMode Address="0x00122F08" Mode="NEAREST" Description="Fixes issue with inability to select some options in UI." />
 	</GameConfig>
+	<GameConfig Executable="SLPM_666.75;1" Title="Kingdom Hearts II Final Mix">
+		<BlockFpUseAccurateAddSub Address="0x001DA9FC" Description="Fixes issue with Tron level boss" />
+		<BlockFpUseAccurateAddSub Address="0x001DAA0C" Description="Fixes issue with Tron level boss" />
+	</GameConfig>
 	<GameConfig Executable="SCES_500.03;1" Title="Dead or Alive 2">
 		<Patch Address="0x002b4c44" Value="0x24060000" Description="Make the file reading RPC call synchronous. Requires proper data cache handling to avoid issue." />
 	</GameConfig>
--- a/Source/Ps2Const.h
+++ b/Source/Ps2Const.h
@@ -32,7 +32,8 @@ namespace PS2
 	enum
 	{
 		GS_NTSC_HSYNC_FREQ = 15734,
-		GS_PAL_HSYNC_FREQ = 15625
+		GS_PAL_HSYNC_FREQ = 15625,
+		GS_HDTV_480P_HSYNC_FREQ = 31469,
 	};
 
 	enum
--- a/Source/ee/MA_EE.cpp
+++ b/Source/ee/MA_EE.cpp
@@ -689,6 +689,17 @@ void CMA_EE::PADDSB()
 	PullVector(m_nRD);
 }
 
+//19
+void CMA_EE::PSUBSB()
+{
+	if(m_nRD == 0) return;
+
+	PushVector(m_nRS);
+	PushVector(m_nRT);
+	m_codeGen->MD_SubBSS();
+	PullVector(m_nRD);
+}
+
 //1A
 void CMA_EE::PEXTLB()
 {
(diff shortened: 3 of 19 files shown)


Hippo-Pippo:
interesting, I tested the patch with the US version (DrayTek Vigor22) and I confirm that at the start of the race you can clearly hear all the bgm/sfx sounds but the video remains black (vulkan/gl)

derole1:
> interesting, I tested the patch with the US version (DrayTek Vigor22) and I confirm that at the start of the race you can clearly hear all the bgm/sfx sounds but the video remains black (vulkan/gl)

Yeah it's really strange, I should also note when this happens in demo mode, the demo never ends (cars keep driving around past when the music stops) and if you press start/insert a coin to start the game the GS recovers and starts rendering again as normal.

jpd002:
Very nice work! Test locally and it works great!

About the rendering issues, it seems the idle detector is making the game unhappy. Adding `return false;` [here](https://github.com/jpd002/Play-/blob/16c4250b1687cf6f45e5c62a79d0fb44bdb34354/Source/ee/Ee_SubSystem.cpp#L304) fixes the issue, but ofc, it makes things really slow.

image

I can try to dig deeper, though my time is super limited right now, I'll see what I can do.

derole1:
Awesome thanks! I have tested my end and it does indeed work, I did some digging and found that it seems to be caused by the MIPS_EXCEPTION_IDLE exception, removing that makes it work at a locked 60fps

image

I'm going to leave this out of my PR as it will likely affect other games, but might be a clue to whats going on

derole1:
Also no rush at all with looking at this I appreciate the reply :)

derole1:
PR will be ready for review shortly unless theres anything I need to amend/add/test?

Steering is now fully emulated (Battle gear 3 doesnt handle this through JVS but through its force feedback board, so it requires more in-depth emulation than the namco stuff), I have also left the MIPS_EXECUTION_IDLE patch out as I don't believe this is the right way to handle the idle detector issues and doesnt actually solve the root cause.

One observation, the ghost replays seem to de-sync after about a minute, both in demo mode and in game, this I can only assume is to do with timing/floating point inconsistencies between real ps2 and the emulator, causing ghosts inputs to de-sync and cause the car to drive into a wall.

Other than that game is perfectly playable, I am also fairly confident the PS2 port of this game will become playable after the idle detector issues are fixed.

Theres also still the white/black level function patch, which im unsure why this fixes the white screen issue but it doesnt seem to have any adverse effects so probably makes sense to leave that in place.

image

Thanks for everyones input with this :)

derole1:
Also pleased to report that Battle Gear 3 Tuned works as well with identical performance after removing MIPS_EXCEPTION_IDLE and patching the "RACK ERROR" that appears for some reason im not quite sure of yet?:

image
image

Theres a few ways to tackle the rack error, one is to implement SIF module 0x00055145 , and return 0x50 for method call 0xFE, the SIF module in question is used in BG3 USA verison as well for some other things with different commands, but only BG3 Tuned requires a valid response from this module to operate, unsure of the reason currently or what the module even does?

The other (less correct) way to do things is to just patch 0x00160D58 to value 0x10000011, this changes the beq opcode to a b, forcing a jump over the rack error which allows the game to boot normally, currently there doesnt seem to be any adverse effects caused by this patch?

I'm not sure if I should add this as there isnt an official MAME romset for tuned from what I can see, just Ver 2.01A which is the american version. The above rack error patch is for version 2.05J of tuned.

Just going to fix the formatting issues and double check against the contribution guidance for formatting then will raise this for review unless theres anything you'd like me to amend :)

Hippo-Pippo:
I did a quick TP port just for testing and I get 60fps with MIPS_EXCEPTION_IDLE false (tested on US version)

bg32111

derole1:
Going to put this as ready to review, a few minor graphical glitches in some areas but overall quite playable now, at least once the mips idle exception is figured out for this game.

Only comments would be about the ghost desync, this actually doesnt happen on replays recorded by the emulator so im guessing its a timing inaccuracy with real hardware or something similar. Hardly game-breaking though. Also not sure about tuned if we are sticking strictly to what is listed in MAME or not? If not I can add an arcadedef for it.

image
The same mips idle exception is also what is stopping the PS2 port form working as suspected!

Any issues/amendments please let me know, thanks again :)

jpd002:
Very nice! 

We're not strictly following MAME, but I'm trying not to create divergent naming if we can. I see that MAME now has a `batlgr3` romset, maybe we could rename our `bg3` to `batlgr3` and add `batlgr3t` for Tuned.

Once the formatting issues are fixed, I can merge this. Still haven't had time to check for the idle skip problem though, I'll try to do that soon.

derole1:
> Very nice!
> 
> We're not strictly following MAME, but I'm trying not to create divergent naming if we can. I see that MAME now has a `batlgr3` romset, maybe we could rename our `bg3` to `batlgr3` and add `batlgr3t` for Tuned.
> 
> Once the formatting issues are fixed, I can merge this. Still haven't had time to check for the idle skip problem though, I'll try to do that soon.

All done :)

For good measure ive also gone ahead and added 2.07J and 2.08J of battle gear 3, again not officially in MAME but hopefully the naming is close enough

jpd002:
Thanks again for this!

jpd002:
Here's a game patch that fixes the hanging issue: https://github.com/jpd002/Play-/commit/29bf47316ae409099a36ea4cae084d9574e95dd1.

The game keeps reading from T1_COUNT and that throws off the idle loop detector. I think it's only doing that for some performance tracking that is not necessary for the game.

PR #1494, opened 2025-10-01
Port T1_COUNT patches to other battle gear 3 arcadedefs and fix hang in race mode caused by another T1_COUNT read

derole1:
As the title suggests, I noticed in PR #1491 that the T1 count patch was implemented for battle gear 3 2.01A but not the other games. I have quickly ported these to the other arcadedefs and raised another PR (hopefully this is okay).

Have tested and works perfectly my end :)


jpd002:
Nice! I don't have the BG3 variants handy, thanks for looking into those!

PR #1495, opened 2025-10-14
System 2x6 Lightgun recoil output

bagelswitch:
This extends the Namco TSS-I/O board emulation to include handling of GPIO writes, which are used for System 2x6 lightgun recoil output. 

It provides MAMEHooker recoil output (P1 only) by adding MAME-compatible windows message interop to the main window.

### DIFF
--- a/Source/ui_qt/mainwindow.h
+++ b/Source/ui_qt/mainwindow.h
@@ -158,7 +158,7 @@ private:
 		PVOID lpData;
 	} COPYDATASTRUCT, *PCOPYDATASTRUCT;
 
-	bool nativeEvent(const QByteArray& eventType, void* message, qintptr* result) override;
+	bool nativeEvent(const QByteArray& eventType, void* message, qintptr* result);
 	//
 #endif
 


bagelswitch:
I should add that this is tested working with TC3 and TC4. Cobra uses a different mechanism (a rumble motor, not a recoil solenoid, and not driven by GPIO output, AFAICT) that I'm still investigating.

MrG1001:
Would really like to try this out for TC3. I know the original arcade allowed for switching to different weapons like the machine gun and had unique solenoid outputs for each weapon. Hoping this gets implemented in the future 😁

bagelswitch:
> Would really like to try this out for TC3. I know the original arcade allowed for switching to different weapons like the machine gun and had unique solenoid outputs for each weapon. Hoping this gets implemented in the future 😁

That is already working here. The game only produces one kind of recoil signal - when the machine gun is used it just rapidly toggles on/off so long as the trigger is held.

MrG1001:
Oh cool good to know. I believe Vampire night also had recoil outputs and ran on similar hardware unless I’m mistaken

bagelswitch:
> Oh cool good to know. I believe Vampire night also had recoil outputs and ran on similar hardware unless I’m mistaken

It does, but used a different I/O board than TC, and seems to use different outputs for recoil. I am looking at it as well, but without a bidirectional JVS capture from real hardware it is slow going.

bagelswitch:
I'm giving up on Cobra and Vampire Night for now - I've tried 5 different valid i/o board types (RAYS, V291, V329, V221, FCA), implemented every capability and command they support, and all master<->node JVS traffic matches real traffic captures, but neither of these titles ever produce any outputs (and for what it's worth, both titles report "Connect NG!" for i/o board status, while TC3 and TC4 both report "Connect OK"). Whatever the problem is, I don't think it is in the i/o board emulation.

bagelswitch:
Well Vampire Night mystery is solved:

image

There are no recoil outputs, because the game has no recoil :)

MrG1001:
Wow that’s interesting I never realized that. Guess I always just assumed the game had outputs since the arcade cabinet used the time crisis guns lol. I looked for videos of people playing the game to double check and the few that I could find (of people playing the arcade and not ps2 version) showed that yeah the guns don’t recoil. 

Still makes me wonder why they went with using the time crisis guns for the arcade cab instead of cheaper lightguns if the game doesn’t have recoil 🤔

bagelswitch:
I apologize for the repeated build failures - Visual Studio seems to not be using the .clang-format from the repo, I will sort that out. In any case, the PR should be clean now.

jpd002:
No worry :smile: Thanks a lot for this, gimme some time to review it, I'll get back to you soon.

bagelswitch:
Closing, in favor of https://github.com/jpd002/Play-/pull/1496

PR #1496, opened 2025-10-19
Alternative System 2x6 lightgun recoil output implemention, using MAME-compatible network (TCP:8000) listener rather than Windows messages

bagelswitch:
After further researching current retail and DIY lightgun options, it looks like the MAMEHooker/Windows Message-style interop is no longer popular, and more users would be served by instead mimicking the MAME network outputs (MAME listens on TCP:8000 and sends events to connected clients). Supermodel and Demulshooter do the same, and Sinden's latest gun software incorporates a client to connect directly to either w/out any intermediate software.

This PR is an alternative to/supersedes [PR #1495](https://github.com/jpd002/Play-/pull/1495)

### DIFF
--- a/Source/iop/namco_sys246/MameCompatOutput.cpp
+++ b/Source/iop/namco_sys246/MameCompatOutput.cpp
@@ -24,8 +24,8 @@ MameCompatOutput::MameCompatOutput(std::string gameId)
 MameCompatOutput::~MameCompatOutput()
 {
 	m_doListen = false;
-	std::this_thread::sleep_for(std::chrono::milliseconds(100));
 	Stop();
+	m_listenThread.join();
 }
 
 void MameCompatOutput::Listen(std::string gameId)
@@ -104,11 +104,10 @@ void MameCompatOutput::Listen(std::string gameId)
 
 void MameCompatOutput::Start(std::string gameId)
 {
-	std::thread t([this](std::string gameId) {
-		Listen(gameId);
-	},
-	              gameId);
-	t.detach();
+	m_listenThread = std::thread(
+	    [this, gameId]() {
+		    Listen(gameId);
+	    });
 }
 
 void MameCompatOutput::Stop()
--- a/Source/iop/namco_sys246/MameCompatOutput.h
+++ b/Source/iop/namco_sys246/MameCompatOutput.h
@@ -16,6 +16,7 @@ namespace Iop
 
 			bool m_doListen = false;
 			int m_clientSocket = 0;
+			std::thread m_listenThread;
 		};
 	}
 }


PolybiusExtreme:
I would like to have the old Mamehooker option too. So both options would be really cool.

jpd002:
Very cool!

Do you think we could make the OutputServer calls completely contained inside `CSys246` rather than making it dependent on the UI code? Basically, could we do something like this? https://github.com/jpd002/Play-/blob/master/Source/iop/namco_sys147/Iop_NamcoSys147.cpp#L67. This is for Animal Kaiser's card scanner server code, but `CSys147` is responsible for starting/stopping the server.

Also, could we rename `OutputNetwork` to something like `MameHookerOutput` and move it in `Source/iop/namco_sys246/` dir?

Thanks!

bagelswitch:
> Very cool!
> 
> Do you think we could make the OutputServer calls completely contained inside `CSys246` rather than making it dependent on the UI code? Basically, could we do something like this? https://github.com/jpd002/Play-/blob/master/Source/iop/namco_sys147/Iop_NamcoSys147.cpp#L67. This is for Animal Kaiser's card scanner server code, but `CSys147` is responsible for starting/stopping the server.
> 
> Also, could we rename `OutputNetwork` to something like `MameHookerOutput` and move it in `Source/iop/namco_sys246/` dir?
> 
> Thanks!

Yep, done! I couldn't call the listener "MameHookerOutput" since this mechanism explicitly _won't_ work with MAMEHooker, as that program requires the Windows message interaction (vs. QMAMEHook, Hook of the Reaper, etc. that can consume this). I went with the more generic "MameCompatOutput" - hopefully that's OK.

jpd002:
Yup, all good! Gonna do a few tests and merge this if everything's alright.

jpd002:
I've tested a bit and I've noticed there's an issue when you boot an arcade game after you've started one already. It seems `WSACleanup` is being called while the other thread is inside `accept`, and this causes the call to fail and return -1. Then it gets in some infinite loop.

Could we use the dtor of the class to manage the lifetime of the thread and the socket? I don't think it needs to be a class with static members.

bagelswitch:
> I've tested a bit and I've noticed there's an issue when you boot an arcade game after you've started one already. It seems `WSACleanup` is being called while the other thread is inside `accept`, and this causes the call to fail and return -1. Then it gets in some infinite loop.
> 
> Could we use the dtor of the class to manage the lifetime of the thread and the socket? I don't think it needs to be a class with static members.

Yep, I see what you mean. I've changed things as you suggest, the listener lifecycle is now instanced, and bound to that of the CSys246 instance. I tested starting/restarting multiple games in the same emulator session, and things now work as expected.

jpd002:
Do you mind if I push this commit to your branch? https://github.com/jpd002/Play-/commit/5f6b6422fa3f674d59bb9ddb2b56c336656e6ec3

I also tied the thread's lifetime to the MameCompatOutput class's lifetime to avoid some other issues.

bagelswitch:
> Do you mind if I push this commit to your branch? [5f6b642](https://github.com/jpd002/Play-/commit/5f6b6422fa3f674d59bb9ddb2b56c336656e6ec3)
> 
> I also tied the thread's lifetime to the MameCompatOutput class's lifetime to avoid some other issues.

[edit]: I went ahead and pushed it.

jpd002:
Thanks a lot for this!

Issue #1265, opened 2023-07-09
Time Crisis 4 navigation voice plays very slow


Hydreigon223:
The navigation voice that plays during gameplay plays slower than it should on h/w. This is because of an additional board connecting the navigation voice channel (likely one of the PS2's surround sound channels) either converting/filtering the voice channel to increase the speed/rate or something else. The functions of that additional board, named V329 EXSOUND PCB, are currently unknown.

Video reference: https://www.youtube.com/watch?v=KYe3qH_TcPc

bagelswitch:
Some of the music in game as well as the tutorial audio and the radio/support audio are streamed PCM (written to sound data input area on the SPU rather than to SPU-processed voices like the other sound effects). For the radio call audio the game is indeed setting SPDIF output type to "bypass" (I see the write of 0x0100 to the spdif output type register on the SPU), which means data in SPU core 0 input area is meant to be streamed directly to the SPDIF output, not sampled and mixed with the 24-voice stereo output. 

This stream data seems to contain twice the number of elements _per unit time_ than would be needed for playback at the correct speed and pitch if sampled at 48kHz, as the emulator was doing - the only way to get it played back correctly is to either sample it at 96kHz, or step twice as far through the data for each sample. It's _possible_ that this has something to do with the actual audio being multi-channel, but I can't tell - In my PR I just double the sampling step size and get a correct-sounding result.

bagelswitch:
I should add that I don't think the EXSOUND board does anything special - I suspect that if you sent this stream data to any S/PDIF receiver (like a home stereo receiver) it would play it correctly, as clock sync can be obtained from the audio data itself.

bagelswitch:
Just following up here for completeness:  turns out this had nothing to do with sample rate per se - rather, the streamed data written to the sound data input area in S/PDIF bypass mode is in a linear interleaved stereo format. When this is de-interleaved correctly by pulling two adjacent samples per cycle to mix into L and R outputs, you get the correct playback rate and pitch.

jpd002:
Fix now in `master`.

PR #1502, opened 2025-10-29
Process sound input data on spu core 0 as interleaved stereo, when S/PDIF bypass mode is active

bagelswitch:
if the SPDIF_OUT register (0x1F9007C0) contains value 0x0100, then data from the sound data input area should bypass spu2 internal processing. On real hardware this would then be fed directly to outputs as PCM or bitstream data. In emulation this data is sampled and mixed in with the spu channel/voice data. In order to get the correct speed and pitch in emulation, spdif bypass data needs to be sampled with twice the stepping of voice data.

This will resolve issue https://github.com/jpd002/Play-/issues/1265

### DIFF
--- a/Source/iop/Iop_Spu2.cpp
+++ b/Source/iop/Iop_Spu2.cpp
@@ -129,10 +129,10 @@ void CSpu2::LogRead(uint32 address)
 
 	switch(address)
 	{
-	LOG_GET(C_IRQINFO)
-	LOG_GET(C_SPDIF_OUT)
-	LOG_GET(C_SPDIF_MODE)
-	LOG_GET(C_SPDIF_MEDIA)
+		LOG_GET(C_IRQINFO)
+		LOG_GET(C_SPDIF_OUT)
+		LOG_GET(C_SPDIF_MODE)
+		LOG_GET(C_SPDIF_MEDIA)
 
 	default:
 		CLog::GetInstance().Warn(LOG_NAME, "Read an unknown register 0x%08X.\r\n", address);


Source/iop/Iop_Spu2.h
```diff
@@ -32,6 +32,16 @@ namespace Iop
 			REGS_END = 0x1F90FFFF,
 		};
 
+		enum REGISTERS
+		{
+			// SPDIF interface
+			SPDIF_OUT = 0x1F9007C0, // SPDIF Out: OFF/'PCM'/Bitstream/Bypass
```
jpd002: Can we move the definitions for these in the `enum` that already has `C_IRQINFO`?

Can we also rename the definitions to match these? https://github.com/ps2dev/ps2sdk/blob/60fa652b3ec082f6871e7d00af06ce7b1b3eeaa3/iop/sound/libsd/include/spu2regs.h#L85

I know `SPDIF_PROTECT` isn't there, but we can name that one `C_SPDIF_PROTECT`.

Source/iop/Iop_Spu2.h
```diff
@@ -23,7 +23,11 @@ namespace Iop
 
 		enum
 		{
-			C_IRQINFO = 0x1F9007C2
+			C_SPDIF_OUT = 0x1F9007C0,
```
jpd002: We don't need `C_SPDIF_OUT`.

jpd002:
Added a bunch of comments, but I think it's good in general :) Thanks again for the contribution!

Source/iop/Iop_Spu2.h
jpd002: Nvm, I had too much sake. 😄

bagelswitch:
> Added a bunch of comments, but I think it's good in general :) Thanks again for the contribution!

Thank you - unfortunately I think I got something fundamentally wrong here - looking at this stream data some more, I think it is actually interleaved stereo - ie. something like "[LL][RR] [LL][RR] [LL][RR] . . ." So the issue isn't really the sample rate - doubling the rate just results in correct but monaural output. 

What is actually needed (I think) is to pull two adjacent samples in each cycle, and mix one into each of L/R output, rather than pulling R output from the second half of the buffer. This will give stereo output at the correct rate.

I will revise the PR to de-interleave the stream correctly.

bagelswitch:
Revised, behavior is fully correct now.

Source/iop/Iop_SpuBase.h
```diff
@@ -274,6 +274,10 @@ namespace Iop
 
 		static bool g_reverbParamIsAddress[REVERB_PARAM_COUNT];
 
+		static uint32 m_spdifOutput;
```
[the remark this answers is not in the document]
jpd002: Way nicer now, thanks 👍

Source/iop/Iop_Spu2.cpp
```diff
@@ -85,24 +87,46 @@ uint32 CSpu2::ReadRegisterImpl(uint32 address, uint32 value)
...
+		m_core[0]->GetSpuBase().SetInputBypass(value & SPDIF_OUT_BYPASS ? true : false);
+		break;
+	case C_SPDIF_MODE:
+		m_spdifMode = value;
+		break;
+	case C_SPDIF_MEDIA:
+		m_spdifMedia = value;
+		break;
+	default:
+		LogWrite(address, value);
```
jpd002: Same comment as `LogRead` above, we can keep it where it was before.

jpd002:
Nice finding about the interleaving of the stereo samples. 

Got some more minor/cosmetic things to address, but I think this is in a good place!

bagelswitch:
Addressed comments re. debug logging in Iop_Spu2.

jpd002:
Very nice, thanks! I'll do some more testing on my end and merge if everything's working good.

jpd002:
Working great! Thanks again!

Issue #1273, opened 2023-07-29
Vampire Night: Bad sound mixing.


Hydreigon223:
Voices during cutscenes are overdriven. Music during gameplay is overdriven while sounds/voices are barely audible (except for a voice clip after a boss hits you that plays too loud). All of this doesn't happen on pcb.

Video reference: <iframe width="560" height="315" src="https://www.youtube.com/embed/ziSPREPlk8U" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>

PR #1510, opened 2025-12-01
Add SPU2 BVOL and AVOL register write handling, and correct mixing to preserve relative volumes of channel and sound input data.

bagelswitch:
This fixes incorrect relative volume in System 246 Vampire Night (and possibly other titles), where the BGM and dialogue was much too loud relative to SFX, by changing the following, per "SPU2_Overview_Manual" section 2.2:

- Sound data input samples are mixed with channel audio at a volume corresponding to per-core BVOL register values. 
- Core0 combined output is mixed into Core1 output at a volume corresponding to Core1 AVOL register values.
- All sample mixing is changed from simple additive (+clamping), to an algorithm that preserves relative volume of both sources w/in the dynamic range.

This will resolve https://github.com/jpd002/Play-/issues/1273

### DIFF
--- a/Source/iop/Iop_SpuBase.cpp
+++ b/Source/iop/Iop_SpuBase.cpp
@@ -211,7 +211,7 @@ void CSpuBase::Reset()
 
 	m_inputVolL = 0x7FFF;
 	m_inputVolR = 0x7FFF;
-    m_extInputVolL = 0x7FFF;
+	m_extInputVolL = 0x7FFF;
 	m_extInputVolR = 0x7FFF;
 
 	memset(m_channel, 0, sizeof(m_channel));
@@ -252,7 +252,7 @@ void CSpuBase::LoadState(Framework::CZipArchiveReader& archive)
 		m_inputVolL = state.GetRegister32(STATE_REGS_INPVOLUMELEFT);
 		m_inputVolR = state.GetRegister32(STATE_REGS_INPVOLUMERIGHT);
 		m_extInputVolL = state.GetRegister32(STATE_REGS_EXTINPVOLUMELEFT);
-	    m_extInputVolR = state.GetRegister32(STATE_REGS_EXTINPVOLUMERIGHT);
+		m_extInputVolR = state.GetRegister32(STATE_REGS_EXTINPVOLUMERIGHT);
 		RegisterStateUtils::ReadArray(state, m_reverb, STATE_REGS_REVERB_FORMAT);
 	}
 


jpd002:
Sorry for the long delay, but I'll be checking this in the upcoming days.

bagelswitch:
> Sorry for the long delay, but I'll be checking this in the upcoming days.

No worries! Thank you.

Source/iop/Iop_Spu2_Core.h
```diff
@@ -89,6 +89,8 @@ namespace Iop
 				P_EVOLR = 0x1F900766,
 				P_BVOLL = 0x1F90076C, //Audio input volume
 				P_BVOLR = 0x1F90076E,
+				P_AVOLL = 0x1F900768, //External Input Volume (Core 1)
```
jpd002: Could we move those definitions above BVOLL? Just to keep things in incrementing order.

Source/iop/Iop_Spu2_Core.cpp
```diff
@@ -267,6 +267,18 @@ uint32 CCore::WriteRegisterCore(unsigned int channelId, uint32 address, uint32 v
 		case A_EEA_HI:
 			m_spuBase.SetReverbWorkAddressEnd(((value & 0x0F) << 17) | 0x1FFFF);
 			break;
+		case P_BVOLL:
```
jpd002: Could we have the cases in this order: P_AVOLL, P_AVOLR, P_BVOLL, P_BVOLR?

Source/iop/Iop_SpuBase.cpp
```diff
@@ -690,9 +690,27 @@ int32 CSpuBase::ComputeChannelVolume(const CHANNEL_VOLUME& volume, int32 current
 
 void CSpuBase::MixSamples(int32 inputSample, int32 volumeLevel, int16* output)
 {
+	// Mixing based on https://www.vttoth.com/CMS/technical-notes/?view=article&id=68
```
jpd002: Could we leave the changes to this function out for now? While this might give a better result, but it's a tad more complex than the original version, so, I'm not sure it's worth the performance penalty.

jpd002 (changes requested):

Source/iop/Iop_Spu2_Core.h
bagelswitch: Done

Source/iop/Iop_Spu2_Core.cpp
```diff
@@ -481,6 +493,8 @@ void CCore::LogWrite(uint32 address, uint32 value)
 		LOG_SET(P_EVOLR)
 		LOG_SET(P_BVOLL)
 		LOG_SET(P_BVOLR)
+		LOG_SET(P_AVOLL)
```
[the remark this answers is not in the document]
bagelswitch: Done

Source/iop/Iop_Spu2_Core.cpp
bagelswitch: Done

Source/iop/Iop_SpuBase.cpp
bagelswitch: Yep, makes sense. The core issue w/ Vampire Night is resolved by the volume register handling - this may subjectively improve the mix but isn't critical.

Source/iop/Iop_SpuBase.h
```diff
@@ -275,6 +275,14 @@ namespace Iop
 
 		static bool g_reverbParamIsAddress[REVERB_PARAM_COUNT];
 
+		uint32 m_inputVolL = 0x7FFF;
```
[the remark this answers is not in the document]
bagelswitch: Done

jpd002:
Nice work! Thanks!

Issue #1473, opened 2025-06-28
Request: Time Crisis 3 to Use CCD Camera Gun Inputs Instead Of Optical Light Gun


kim-sama:
Time Crisis 3 in Japan and Asia regions used a different tech of light gun, the same as Crisis Zone and Vampire Night where the gun has a IR light inside and an CCD camera is used to track the light. While the rest of the world (USA and Europe) use the usual light gun tech for their SD cabinets. And Time Crisis 3 on Play uses the light gun inputs. The advantage of using CCD camera gun input instead of light gun is that it doesn't not flash the screen since it relies on IR light detection instead of the old light gun detection.

pizzzza19:
There are 3 kinds of DVD disc are known to be exist but image should be all same. Also v291 gun I/O board is same. If Time crisis 4 is woking fine without screen flashing, it may be easy to modify just a little bit code but not sure where it is.

jcstahl1:
This would only really matter on real arcade hardware, and would make no difference for emulation.   the game software itself is only looking for positional coordinates.

Hydreigon223:
> There are 3 kinds of DVD disc are known to be exist but image should be all same. Also v291 gun I/O board is same. If Time crisis 4 is woking fine without screen flashing, it may be easy to modify just a little bit code but not sure where it is.

I recall Time Crisis 3 has two gun trigger inputs instead of one. One would produce a white flash like with most other optical lightgun games while the other would not.

pizzzza19:
If it's true, just disable one of trigger inputs will solve the problem. It depends on the developer decision which to pick whether accuracy of real hardware emulation or gaming experience.

Issue #1254, opened 2023-06-10
[Sys2x6 Feature Request] Allow JVS ID string to be configured through arcadedefs


Hydreigon223:
https://github.com/jpd002/Play-/blob/namco_system_256/Source/iop/Iop_NamcoArcade.cpp#LL156C3-L170C4
`case JVS_CMD_IOIDENT:
		{
			(*output++) = 0x01; //Command success
			(*dstSize)++;

			//const char* boardName = "namco ltd.;RAYS PCB;";
			const char* boardName = "namco ltd.;TSS-I/O;";
			size_t length = strlen(boardName);

			for(int i = 0; i < length + 1; i++)
			{
				(*output++) = boardName[i];
				(*dstSize)++;
			}
		}`

We should have the string "boardname" determined through arcadedefs instead of a hardcoded default value. Keeping the ID string as TSS-I/O  is not very sane as it's a one player optical lightgun board used in Time Crisis 2 and 3. The Taito games running on System 2x6 (at least Battle Gear 3) are known for checking this string for a specific value (as well as checking the I/O board serial port for a servo motor board but that's a different topic).

jpd002:
Yup, makes sense, definitely an improvement I need to make. That ID seemed to work for games I've tested, but I haven't tested most of the games yet.

Keropon:
Does it just compare the ID string? I know each IO has different configurations on available channels and stuff like that

Hydreigon223:
> Does it just compare the ID string? I know each IO has different configurations on available channels and stuff like that

For BG3, it's stated on s2x6 compatibility document that it requires a very specific JVS IO board (likely checks the ID along with I/O bits and A/D channels being an equal number). I also made the mistake of mentioning where the JVS serial port connects to and correctly edited that it connects to a card reader instead. The servo motor board on the other hand connects through the same serial port used to connect an STR motor feedback pcb.

Idk if any other checks are used for the card reader but that in itself is nothing fancy. All it just does is read a serial ID. The profile saving/loading on the other hand is done with an online server instead (requires usb ethernet adapter support and later nesys server emulation with the latter being "someone else's" responsibility when that time comes).

Keropon:
Do you know if BG3 has a hidden offline mode as im@s does? (this is obviously not related to the issue)

[...]

PR #1516, opened 2025-12-20
Set JPN version MIU-I/O board ID for TC3, which will prevent screen flash on trigger pulls, and swap R2 and TRIANGLE inputs

bagelswitch:
JPN TC3 cabinets used a CCD-based lightgun, while other regions used a traditional screenflash+timer setup. All versions of the software will work with either gun type - they decide which based on the I/O board's ID string. "namco ltd.;MIU-I/O;Ver2.05;JPN,GUN-EXTENTION" will cause the game to use the CCD-type gun. This is preferable for emulation because it doesn't flash the screen on trigger pulls. This PR changes the I/O board ID string when the game ID is timecrs3. It also swaps R2 and TRIANGLE button inputs, since the game expects the former as trigger input when using the JPN I/O board.

This will resolve https://github.com/jpd002/Play-/issues/1473

### DIFF
--- a/Source/ui_shared/arcadedrivers/NamcoSys246Driver.cpp
+++ b/Source/ui_shared/arcadedrivers/NamcoSys246Driver.cpp
@@ -131,7 +131,8 @@ void CNamcoSys246Driver::PrepareEnvironment(CPS2VM* virtualMachine, const ARCADE
 				break;
 			}
 
-			if (!def.boardId.empty()) {
+			if(!def.boardId.empty())
+			{
 				namcoArcadeModule->SetBoardId(def.boardId);
 			}
 		}


jpd002:
Very nice finding!

I think it would be nice to make this driven by the arcade definition file. Like, could we add an element in the JSON that would be something like `boardName` that could be used to override the default we have inside the emulator?

For the button swapping, I think we could probably do it in arcadedef as well. You can check the `buttons` section in the `timecrs4` or `cobrata` files.

Thanks!

bagelswitch:
> Very nice finding!
> 
> I think it would be nice to make this driven by the arcade definition file. Like, could we add an element in the JSON that would be something like `boardName` that could be used to override the default we have inside the emulator?
> 
> For the button swapping, I think we could probably do it in arcadedef as well. You can check the `buttons` section in the `timecrs4` or `cobrata` files.
> 
> Thanks!

Yep, makes sense. I'll implement both and update the branch.

bagelswitch:
> Very nice finding!
> 
> I think it would be nice to make this driven by the arcade definition file. Like, could we add an element in the JSON that would be something like `boardName` that could be used to override the default we have inside the emulator?
> 
> For the button swapping, I think we could probably do it in arcadedef as well. You can check the `buttons` section in the `timecrs4` or `cobrata` files.
> 
> Thanks!

Adjusted as requested.

Hydreigon223:
This needs to be merged. In addition, this makes it possible to add Taito's specific JVS ID to run Battle Gear 3; 

TAITO CORP. I/O PCB-24/24/8/2SE ver1.2forBG3

I also may add this ID to my game defs at some point;

namco ltd.;ASCA-3;Ver2.02;JPN,Multipurpose + Rotary Encoder

This specific ID comes from one JVS I/O board in my Motocross Go cab. In addition, this will also resolve #1254

jpd002:
Looking great, thanks!

PR #1518, opened 2025-12-22
reduce GPIO output threshold for recoil events to accomodate both I/O board types

bagelswitch:
Unfortunately, https://github.com/jpd002/Play-/pull/1516 broke recoil for TC3, as the recoil output signal has a lower value for whatever reason (0x50 rather than 0xC0) when using the MIU type I/O board. This sets the threshold lower for writing a recoil event via the network output and fixes recoil in TC3.


jpd002:
Thanks! Would it make sense to make this configurable per game in the future?

bagelswitch:
> Thanks! Would it make sense to make this configurable per game in the future?

I thought about it, but it is only two games, currently, that have working recoil to begin with, and they're both fine w/ 0x50 . . . I am still poking at Cobra intermittently to try to figure out how recoil worked there - if I/someone ever figures that out maybe additional config will make sense?

PR #1525, opened 2026-01-16
CI : update upload-artifact version

Mrlinkwii:
updates upload-artifact version to the lastest version (6) 

mainly a copy of teh pr i did in https://github.com/jpd002/Play-/pull/1437 just to the newer versions 

hopefully this dosent break anything , but it shouldn't

also would Recommend enabling [dependabot](https://github.blog/news-insights/product-news/keep-all-your-packages-up-to-date-with-dependabot/) because most of the action versions are out of date


jpd002:
Nice! Thanks for this.

Also thanks for the Dependabot suggestion, I'll give it a try.

PR #1553, opened 2026-02-03
taiko11a.arcadedef added

pizzzza19:
[arcade] Taiko no tatsujin 11 Asia version added.
Asia version Rom will check system NVRAM region value. It is marked as not working now.


jpd002:
Thanks, I guess it's good until we figure this out.

PR #1565, opened 2026-03-20
System 246 controller mapping improvement for driving and drum

pizzzza19:
For driving games such as Battle Gear 3
L1: Shift down
R1: Shift up
L stick: Steeling
R stick vertical: Accelerate and Brake 

For drum games
L2 R2: KA*
Dpad Left and Cross (A): DON
Codes are not optimized so that Right KA may be missed sometimes.


pizzzza19:
There is known issue that TAIKO "KA" response does not accurate but slower.
You can't input double "KA" at the same time but only detects as single.

PR #1579, opened 2026-05-11
Added idle skip patch for the game SD Gundam Force - Showdown!

hector-tomazella:
Fixes high EE usage on the game SD Gundam Force - Showdown!. Performance should be at least 2x the previous framerate on avg (fully tested by playing the entire game)


jpd002:
Nice! Thanks!

PR #1582, opened 2026-05-15
Added idle skip patch for the game After... - Wasure Enu Kizuna

hector-tomazella:
Fixes high EE usage all the time. Improves performance around 1500%+
Note: patched address changed from 0x001407A0 to 0x00106F18, to avoid patching FlushCache (performance improvement remains the same)


jpd002:
`0x001407A0` is the address of `FlushCache`. I would avoid putting it there since FlushCache in a lot of non idle loop scenarios. I suggess `0x00106F18` as a better place.

jpd002:
Thanks!

PR #1583, opened 2026-05-17
Added idle skip patch for the game Simpsons Road Rage

hector-tomazella:
Fixes high EE usage all the time in the game. Gives around a 100% framerate improvement


jpd002:
Working quite well, thanks! Just a note that it seems to make the logo video not to show correctly, but I think it's a different problem that's just made worse/more obvious by this.

PR #1589, opened 2026-06-04
Added idle skip patch for the game Virtua Fighter 4 - Evolution

hector-tomazella:
Fixes high EE usage all the time in the game, improves performance with at least a 100% framerate gain

### DIFF
--- a/GameConfig.xml
+++ b/GameConfig.xml
@@ -31,6 +31,9 @@
 	<GameConfig Executable="SLUS_203.05;1" Title="Simpsons Road Rage">
 		<IdleLoopBlock Address="0x0027C708" />
 	</GameConfig>
+	<GameConfig Executable="SLUS_206.16;1" Title="Virtua Fighter 4 - Evolution">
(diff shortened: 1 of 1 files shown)


GameConfig.xml
```diff
@@ -31,6 +31,9 @@
 	<GameConfig Executable="SLUS_203.05;1" Title="Simpsons Road Rage">
 		<IdleLoopBlock Address="0x0027C708" />
 	</GameConfig>
+	<GameConfig Executable="SLUS_206.16;1" Title="Virtua Fighter 4 - Evolution">
+		<IdleLoopBlock Address="0x00109F18" />
```
jpd002: `0x00109F18` is inside the system call stub for `ReferSemaStatus` and it might be called from other functions that might not be idle waiting. I'm suggesting to mark the block after `ReferSemaStatus` has been called, which is inside a function that is indeed just looping waiting for something to happen:

```suggestion
		<IdleLoopBlock Address="0x001EDA2C" />
```

jpd002:
Thank you very much! 😃

PR #1586, opened 2026-05-27
Battle gear 3 t1 count skipping

pizzzza19:
Old branch removed.
cleaned up previous code and created as new branch.

This will skip T1_Count for Battle Gear 3.

PR #1591, opened 2026-06-04
Update addresses in batlgr3j.arcadedef

pizzzza19:
previous address is wrong and typo.
Updated for 2.10j revision.

Conversation started in #1586 PR, for reference.


jpd002:
Nice! Thanks for this!

PR #1593, opened 2026-06-07
Added idle skip patch for the game Combat Queen

hector-tomazella:
Fixes high EE usage all the time, improves game performance in general by at least 100%, but there's still some improvements to be made since this game is 90% pre rendered videos and those aren't running as fast as they should

### DIFF
--- a/GameConfig.xml
+++ b/GameConfig.xml
@@ -34,6 +34,9 @@
 	<GameConfig Executable="SLUS_206.16;1" Title="Virtua Fighter 4 - Evolution">
 		<IdleLoopBlock Address="0x001EDA2C" />
 	</GameConfig>
+	<GameConfig Executable="SLPM_651.69;1" Title="CombatQueen">
+		<IdleLoopBlock Address="0x1FC0102C" />
+	</GameConfig>
 	<GameConfig Executable="SLUS_203.07;1" Title="Burnout">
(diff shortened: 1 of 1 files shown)


GameConfig.xml
```diff
@@ -34,6 +34,9 @@
 	<GameConfig Executable="SLUS_206.16;1" Title="Virtua Fighter 4 - Evolution">
 		<IdleLoopBlock Address="0x001EDA2C" />
 	</GameConfig>
+	<GameConfig Executable="SLPM_651.69;1" Title="CombatQueen">
+		<IdleLoopBlock Address="0x1FC0102C" />
```
jpd002: `0x1FC0102C` is inside the BIOS's internal DMAC handler, it's probably too broad to be used as a idle block since every DMA transfer will call it.

Code at `0x00104BF8` seems to be more akin to an idle loop:
```
00104BF8    0C04121E    JAL            $00104878
00104BFC    00000000    NOP            
00104C00    1000FFFD    BEQ            R0, R0, $00104BF8
00104C04    00000000    NOP            
```
It's basically just calling the function in a loop (it basically just rotates the threads).

```suggestion
		<IdleLoopBlock Address="0x00104BF8" />
```

jpd002:
Thanks for this!

PR #1600, opened 2026-07-03
Arcade vertical lines removed for pixel gap adjustment

pizzzza19:
Only selected titles will be patched automatically for removing vertical lines
+ soul calibur 3
+ taiko
+ tekken

This list can be modified anytime when other titles needs to be added later.


pizzzza19:
Vertical lines removed but Tekken 4 still screen flashing in adv demo movie.
After insert coins / service, screen gets normal.

Vertical line may be shown during in-game video setting changed from openGL to Vulkan. But this is not nesesarry action and won't be a problems.

PR #1611, opened 2026-08-21
Change envelope log table init

kode54:
Not sure either method is correct. This requires further investigation.

Maybe the existing implementation is already correct? Hmm.


kode54:
I am a dumbass, I couldn't read this last night when I adapted it. The existing method is identical, just neater coded.

